PR #47, opened 2015-01-01
Archives actions

mquandalle:
This PR adds two new actions for archived cards: “send back to board” and “delete”.

A lot of changes in the git diff view are related to empty spaces at the end of lines, which my text editor is configured to automatically remove on file save. Since I'm unlikely to be the only one with this configuration, I would recommend you sticking with this convention and removing trailing spaces in all files.

Still to do:
* [ ] Menu > “Archived items” is not implemented
* [ ] Delete confirmation pop-up

### DIFF
--- a/client/views/activities/templates.html
+++ b/client/views/activities/templates.html
@@ -29,6 +29,10 @@
                         archived <a href="{{ card.absoluteUrl }}" class="action-card">{{ card.title }}</a>
                     {{ /if }}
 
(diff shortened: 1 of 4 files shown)


mquandalle:
Ok, I managed to remove all changes related to trailing spaces, so it's easier to review the PR, but I would still argue in favor of a *no-trailing-spaces* convention.

yasaricli:
very good actions for archived cards; thanks.

PR #48, opened 2015-01-01
Allow a user to “star” a board

mquandalle:
I disabled the trailing white spaces cleaning so this PR should be easier to review.

Still to do:

* [ ] Add a new “Starred Boards” list on the boards home page


yasaricli:
Now serving at http://metrello.meteor.com.

PR #59, opened 2015-01-04
Edit labels

yasaricli:


client/views/cards/events.js
```diff
@@ -207,6 +210,39 @@ Template.cardLabelsPop.events({
...
+            $set = {};
+
+        // set label index 
+        $set[getLabel.key('name')] = name;
+
+        // color current not currentcolor
+        $set[getLabel.key('color')] = selectLabel.color;
+        
+        // update
+        Boards.update(this.boardId, { $set: $set });
```
mquandalle: Hum actually to edit the label name I believe you could rely totally on mongodb without the need to create a `getLabelIndex`. The query is as follow:

```js
Boards.update({
  _id: boardId,
  'labels._id': labelId
}, {
  $set: {
    'labels$.name': newName
  }
});
```

Can you try it?

client/views/cards/helpers.js
```diff
@@ -5,3 +5,14 @@ Template.cardLabelsPop.helpers({
       return _.contains(Cards.findOne(cardId).labelIds, this._id);
     }
 });
+
+Template.editLabelPop.helpers({
+    labels: function() {
+        var defaultLabels = ['green', 'yellow', 'orange', 'red', 'purple', 'blue'],
```
mquandalle: Should we define these default elsewhere; because they seems identical to what we have in `collections/boards.js` so we could share this variable (DRY).

client/views/cards/helpers.js
```diff
@@ -5,3 +5,14 @@ Template.cardLabelsPop.helpers({
...
+            labels = [];
+        _.forEach(defaultLabels, function(color) {
+            labels.push({ color: color, name: '' });
```
mquandalle: Why there is no `_id` here?

client/views/cards/events.js
mquandalle: Also clicking “Save” should return to the previous popup view.

yasaricli: 403, Not permitted. Untrusted code may only documents by ID Meteor.error.

client/views/cards/helpers.js
yasaricli: global defaultLabels variable ?

client/views/cards/events.js
mquandalle: Ok, so let's keep it like this.

That does not makes sense to me, I asked the question on the mailing list, https://groups.google.com/d/msg/meteor-talk/qUa3clv4g5E/R6TziALdUbwJ

PR #69, opened 2015-01-09
Redirect logged in users from login/signup views to boards view

alireza-ahmadi:
When logged in users try to open Login/Signup views they should redirected to Boards view because they already signed in. I personally even like to get redirected to boards view when open Home view but I have not added redirection from home view because I think that Libreboard team may have a reason for not implementing this feature.

Sorry for duplicate pull request, I have accidentally removed the feature branch so I have created another pull request.


mquandalle:
Thank you Alireza, I merged your PR in 2dce97ee79f

>  I personally even like to get redirected to boards view when open Home view but I have not added redirection from home view because I think that Libreboard team may have a reason for not implementing this feature.

Actually I would like that too, what do you think of it @yasaricli?

yasaricli:
Thank you Alireza.

Issue #62, opened 2015-01-06
Popups should be hidden on a click away


Slava:
Click on any button that opens a popup or a submit form such as `#AddCardForm`. A popup/form opens. Click on any part of the body, that is not a button. The popup/form is expected to be closed.

A simple dumb solution that I came up with is listening on the `blur` event, but it is bad, because clicking on buttons is also triggering a blur event.

mquandalle:
Yes, the popup system needs to be improved. I plan to make a general object that we can use more easily in the code (there is a lot of duplicate code regarding click management for instance). And there is also #54.

mquandalle:
About the implementation we can simply add a event listener on the `body` that will close the popup. Then we add a click event on the popup and stop its propagation so that the event does not bubble up to the `body` tag.

Issue #54, opened 2015-01-03
The popup sometimes overflow the screen limit


mquandalle:
We should fix the formula that position the popup window, because it currently doesn't take into account the screen limits.

PR #72, opened 2015-01-12
Popup reform

mquandalle:
* New `Popup` public API that can be used directly in Blaze event maps
* Automatically calculate the popup offset using the event target size
  and the window dimensions -- before this commit the offset was
  hardcoded in the DOM with the "offset" attribute
  #54
* Support localised (translated) popup title
* Fix a bug for the labels popup -- going back was opening the popup
  in a wrong position
* Click anywhere to close a popup
  #62
* Support sub-popups, this is completly transparent in the API, if we
  open a popup B from a popup A that will automatically generate a
  "back" button from B to A
* Change the context of data provided by iron-router

### DIFF
--- a/client/lib/popup.js
+++ b/client/lib/popup.js
@@ -1,0 +1,150 @@
+Popup = {
+    /// This function returns a callback that can be used in an event map:
+    ///
+    ///   Template.tplName.events({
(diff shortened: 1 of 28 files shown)


yasaricli:
:+1: nice.

PR #78, opened 2015-01-16
fix #4

### LINKED ISSUE #4: accounts

Goluis:
let me know if cards are still lost when moved


mquandalle:
Thank you very much @Goluis, I didn't expect you to come fixing directly this project. So this is the new API, right? It fixes https://github.com/Goluis/cottz-publish/issues/4?

mquandalle:
I still experience data lost in an environment with latency. You can try it at http://libreboard.com (yes, deploy in production to test is crazy!)

Goluis:
I've tried several times before uploading but I would like you to tell me how it looks

Goluis:
the latency problem also experienced with publish-composite in production?
http://goluis.meteor.com/ (with publish-composite)

if that's the case I do not find a way to fix except changing the way you subscribe to cards

mquandalle:
It's not exactly the same problem. With publish-composite when you try to move a card sometimes the operation is reverted and the card is back to its original position. In the same situation with cottz:publish the card disappear.

I haven't taken a close look at your respective source codes, but it seems that you are both relying on cursor observation and experiencing the same race condition. Would it be possible to use the meteor merge box to solve this problem? Is your implementation pushed on github?

Goluis:
my theory is that by doing many updates observers do not catch the messages in the correct order (do not know how to control it)
the best solution is to have the subscription of the cards in another publication and subscribe again when updates are completed

Goluis:
to be more clear, even with a basic meteor observer without any package
````js
Meteor.publish('cards', function (listId) {
  return Cards.find({
    listId: listId
  });
});
````
the problem still happen

mquandalle:
So you think it is a bug with Meteor? I'll try to re-write https://github.com/mquandalle/meteor-publish-composite-latency-bug using publish-composite by using two publications instead.

mquandalle:
Oh, but I guess to fix the bug we can simply rely on the `boardId` instead of the `listId`.

mquandalle:
I don't think it's a Meteor bug, the following reproduction works https://github.com/mquandalle/meteor-publish-composite-latency-bug/tree/raw-meteor http://publish-race-condition-2.meteor.com/

I will probably use two distinct publications, but it would be great to have a better fix for this issue.

PR #195, opened 2015-06-14
Update German localization

ePirat:
Added missing translations and fixed some typos and wording.


ePirat:
Sorry that I did not use Transifex, but it was faster for me do to it directly in the json instead of first signing up to some website and translating it there.

mquandalle:
Ok, I'll merge this. But anyway the whole set of translation string will be re-organized before v0.9 because of the UI changes — and also we should take advantage of advanced i18n features such as plural strings.

ePirat:
I agree, just ping me whenever there are new translations needed and I will look into it, making them properly. (BTW why is Github still showing the always failing travis-ci checks here?)

mquandalle:
Merged, thanks :)
It's really easier for me to manage these things on Transifex. Currently the project is a bit frozen until I update the new keys, but may I add you to the German team?

As for Travis CI failure, I think it's just that the test has been executed when you submitted the PR, and it was not updated again when a changed Travis CI configuration (ie, when I disabled it)

ePirat:
Sure, feel free to add me to the German team.

Issue #177, opened 2015-06-09
Buttons misaligned in Safari


ePirat:
Some buttons, namely the tags and (sometimes) Username circles are misaligned in Safari, as seen in the following screenshots:

![bildschirmfoto 2015-06-09 um 16 08 52](https://cloud.githubusercontent.com/assets/323210/8059775/06bf2b98-0ec2-11e5-8346-1e2c95fc6db2.png)
![bildschirmfoto 2015-06-09 um 16 09 26](https://cloud.githubusercontent.com/assets/323210/8059776/089b7c8c-0ec2-11e5-9d3a-3be160a8052b.png)

mquandalle:
Does commit bc9fcc7 fix the problem?

PR #197, opened 2015-06-14
Modified label styling

ePirat:
This is a partial fix for #177, misaligned labels containing text.


ePirat:
While this seemed to work fine in the testcase, it breaks on other places… :disappointed:

PR #198, opened 2015-06-15
change dockerfile to do a full build from git

james-portman:
I could not find a full docker build file for libreboard,
some may find this useful

You would first run mongodb on the host, then build and run this docker image for libreboard code,
this would allow the data to persist reboots and make it easy to try new libreboard code bases or update to latest

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,16 @@
-FROM meteorhacks/meteord:onbuild
-MAINTAINER Maxime Quandalle <maxime@quandalle.com>
+FROM centos:centos7
 
-# Run as you wish!
-# docker run -d --name libreboard-db mongo
-# docker run -d --link "libreboard-db:db" -e "MONGO_URL=mongodb://db" \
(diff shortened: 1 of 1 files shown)


mquandalle:
The current DockerFile works for me, I'm not sure how the change you propose improve things?

james-portman:
The current docker file works, but it just pulls from an image you built
previously,
It's not clear how you built the image or what version/branch/tag it is
built from.

mquandalle:
I'm rejecting this PR because I believe the `meteord` base image is perfectly adapted to this application. If you want to see how the image was built you can read [the corresponding dockerfiles](https://github.com/meteorhacks/meteord). This image also gets update with the correct versions of some of our dependencies (eg, in Wekan v0.9 we need nodejs v0.10.40, and meteord provide this exact version).

Also about your particular Dockerfile, we probably don't want to chip the meteor utility in production images (this is only the build tool), and we don't want to run the server with `meteor` because it is intended for development, not production (for instance the files doesn't get gzipped, no `forever` or equivalent to restart the process in case it dies).

PR #304, opened 2015-09-16
Remove redundant words

AlexanderS:
The `add-label` and `remove-label` strings already contain the "Add" and "Remove" parts.


mquandalle:
Thanks!

PR #305, opened 2015-09-16
Fix multiselect toggle logic

AlexanderS:
If every element already has the label/member, we do not need to add it
but it should be removed and if every element does not have the element,
we should add it.


mquandalle:
Thanks!

For the record, this bug was introduced in 45b662a.

PR #314, opened 2015-09-19
Shortcut: Spacebar assigns self to current card, 'F' toggles filter sidebar

fisle:
Hey,

This implements simple keybind that assigns self to current card when spacebar is pressed.
It's a feature that I like dearly so I implemented it as it was missing.

Also fixed a very minor translate typo in finnish translation.

### DIFF
--- a/client/lib/keyboard.js
+++ b/client/lib/keyboard.js
@@ -36,6 +36,26 @@ Mousetrap.bind(['down', 'up'], (evt, key) => {
   }
 });
 
+// XXX This shortcut should also work when hovering over a card in board view
+Mousetrap.bind('space', (evt) => {
+  if (!Session.get('currentCard')) {
+    return;
+  }
+
+  const currentUserId = Meteor.userId();
+  if (currentUserId === null) {
+    return;
+  }
+
+  if (Meteor.user().isBoardMember()) {
+    const card = Cards.findOne(Session.get('currentCard'));
+    card.toggleMember(currentUserId);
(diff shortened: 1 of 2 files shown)


fisle:
Also added another shortcut: 'F' key toggles filter sidebar.

client/lib/keyboard.js
```diff
@@ -36,6 +44,15 @@ Mousetrap.bind(['down', 'up'], (evt, key) => {
   }
 });
 
+Mousetrap.bind('space', () => {
+  if (!Session.get('currentCard')) {
+    return;
+  }
+  const currentUserId = Meteor.userId();
+  const card = Cards.findOne(Session.get('currentCard'));
+  card.toggleMember(currentUserId);
```
mquandalle: I think you should explicitly handle the case when `currentUserId` is `null`, and verifying that the user is a board member (`Meteor.user().isBoardMember()`) before executing the action

client/lib/keyboard.js
```diff
@@ -36,6 +44,15 @@ Mousetrap.bind(['down', 'up'], (evt, key) => {
   }
 });
 
+Mousetrap.bind('space', () => {
```
mquandalle: BTW in Trello this shortcut works when hovering a card in the board view. Since we currently don’t have a good abstraction to implement that I won't ask you to do so, but could you please add a `// XXX comment` about it?

mquandalle:
Thank you!

PR #316, opened 2015-09-20
Fix card deletion

AlexanderS:
`this` changes during `Popup.afterConfirm` and so we need to save it in a
local variable.

### DIFF
--- a/client/components/cards/cardDetails.js
+++ b/client/components/cards/cardDetails.js
@@ -152,10 +152,10 @@ Template.moveCardPopup.events({
 });
 
 Template.cardMorePopup.events({
-  'click .js-delete': Popup.afterConfirm('cardDelete', () => {
+  'click .js-delete': Popup.afterConfirm('cardDelete', function() {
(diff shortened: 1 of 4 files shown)


client/components/cards/cardDetails.js
```diff
@@ -152,11 +152,15 @@ Template.moveCardPopup.events({
 });
 
 Template.cardMorePopup.events({
-  'click .js-delete': Popup.afterConfirm('cardDelete', () => {
-    Popup.close();
-    Cards.remove(this._id);
-    Utils.goBoardId(this.board()._id);
-  }),
+  'click .js-delete'(event, tpl) {
```
mquandalle: I generally use `evt` here. Could you change it just to be consistent with the rest of the code base?

AlexanderS:
This now fixes all the popup.afterConfirm calls, by using `function() {}` instead of `() {}`.

mquandalle:
Thanks!

PR #326, opened 2015-09-24
sidebar: Filter archived cards/lists for current board

### LINKED ISSUE #318: "Archived items" shows cards/lists by user, not by board
I have a shared board with my team, but when I view "Archived items" it shows all the cards/lists I've archived on my personal board as well. Is this a bug, or I'm missing something? I've been known to overlook the obvious ( #302 :)

Thanks!

AlexanderS:
The archived items should be filtered for the current board or else you
will get a global list of all archived items on all boards.

This fixes #318.

### DIFF
--- a/client/components/sidebar/sidebarArchives.js
+++ b/client/components/sidebar/sidebarArchives.js
@@ -11,11 +11,17 @@ BlazeComponent.extendComponent({
   },
 
   archivedCards() {
-    return Cards.find({ archived: true });
+    return Cards.find({
+      archived: true,
+      boardId: Session.get('currentBoard'),
+    });
   },
 
   archivedLists() {
-    return Lists.find({ archived: true });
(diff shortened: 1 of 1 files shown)


client/components/sidebar/sidebarArchives.js
```diff
@@ -11,11 +11,13 @@ BlazeComponent.extendComponent({
   },
 
   archivedCards() {
-    return Cards.find({ archived: true });
+    const board = Boards.findOne(Session.get('currentBoard'));
```
mquandalle: You don’t need to fetch the entire board document here since only its `_id` is used.

AlexanderS: Fixed.

mquandalle:
Thank you!

Issue #331, opened 2015-09-26
Plugin API


jimmiebtlr:
I'd love a plugin api. I've been doing a bit of work in this area, but would love a discussion on architecture and api.

mquandalle:
Here are my current ideas about the plugin system. Plugins will be distributed as packages that could execute any code on the server or the client. The meteor format (“isopackages”) would probably fit the requirements here but I need to find a way to download them from a market-store like repository, start them and execute them at application runtime, and restrict the execution to certain boards — because most plugins will be enabled at the board level. Whether I end up using the Meteor format or something else, I will definitely rely on some third party package system here as I don’t want to re-implement the abstract logic of a package manager (eg, versions resolutions). So that’s for the “load some third party code in the app”.

About the UI API I would like to take inspiration from the Atom Service API, http://blog.atom.io/2015/03/25/new-services-API.html. Since switching to React is in the plan (see #299) it will probably be better to rely on React render logic to implement the UI extensions mechanisms. It seems that high order components are the recommended way in React, but I’m surprised that no one did implement a UI extend-able by third parties plugins using React (yet).

xavierpriour:
:+1: for starting with the current Meteor package system right now - even something as limited as the import feature would benefit from even a v0.x plugin system.
A first requirement to having plugin packages is to have core packages these plugins can depend on. Something like what the [Telescope](https://github.com/TelescopeJS/Telescope) guys succesfully did, see: http://docs.telescopeapp.org/docs/custom-packages
So I believe that's where we should start. Only "issue" is that some minimum planning will be needed, as this will impact lots of code...

@mquandalle what about the below?

1. define the target packages on paper
2. do a last release on the current architecture when you get back
3. switch to packages
4. happily hack away

mquandalle:
I published a general overview of the desired plugin system: https://discuss.wekan.io/t/rfc-the-plugin-system/132/1. Super curious of what people think of this proposal and I will be happy to discuss it in more details on the forum.

This desired system will be built incrementally, at first we won’t have any distribution mechanism but will just start refactoring existing features into internal proto-plugins. That will help set a working API. I will try to have a first proto-plugin by the end of this week.

xet7:
We are not going to do NPM-based plugin system because of issues mentioned at https://github.com/wekan/wekan/issues/781

PR #332, opened 2015-09-27
Add basic plugin support.

jimmiebtlr:
Move models to packages.
Add plugin package, and insert plugin points in UI.

Would love some discussion on how to improve.
Issue #331 

Adds some simple entry points for a plugin.  Not terribly clean, but setting a template name will result in that template being rendered in the section.

    Wekan.registerPlugin({
      cardAction: "CardTimeToggle",
      listHeader: "ListTimeHeader",
      boardOverview: "TimeBoardOverview"
    });

See an example of a plugin at https://github.com/jimmiebtlr/wekan-time.

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/332)


jimmiebtlr:
It runs, but looks like I have some cleanup for eslint to do.  What's up with the license?

packages/activities/package.js
```diff
@@ -0,0 +1,28 @@
+Package.describe({
+  name: 'wekan:activities',
+  version: '0.0.1',
```
mquandalle: We should keep all the package version number in sync with the main app, ie `0.9.0`.

packages/activities/package.js
```diff
@@ -0,0 +1,28 @@
...
+  // Brief, one-line summary of the package.
+  summary: '',
+  // URL to the Git repository containing the source code for this package.
+  git: '',
+  // By default, Meteor will default to using README.md for documentation.
+  // To avoid submitting documentation, set this field to null.
+  documentation: 'README.md'
```
mquandalle: Let's remove lines 4 to 10.

mquandalle:
Hi @jimmiebtlr,

Thank you very much for this PR. My intuition is that we should be able to customize the set of plugins we use at the board level.

I will probably merge this in a separate branch so I can continue experiment with this architecture. The CLA is pretty common is open-source project, it says that you keep the copyright of your work but you give me a license to give me the right use it.

jimmiebtlr:
I can definitely see the use case for plugins by board.  The specific use case I intended would actually be site wide, but possibly dependent on the user role.

mquandalle:
We are not going to use Meteor packages as the basis of our plugin system, but instead will rely on NPM. I published an overview of the desired plugin system here: https://discuss.wekan.io/t/rfc-the-plugin-system/132/1

As you worked on such a system for Wekan, I would be curious to read your thoughts on my proposal.

PR #340, opened 2015-10-03
1st hack to support linking boards to cards

hsduk:
TODO:
Activities - log link/unlink activities
Filtering - filter boards with/without links
Summary - summarise lists, cards and labels


hsduk:
(please ignore - use https://github.com/wekan/wekan/pull/341)

PR #342, opened 2015-10-03
Initial support for `@user` and `#label` use when creating new cards.

FuzzyWuzzie:
When creating a new [mini]card, typing `@` or `#` brings up an auto-complete box for board members and labels which will get applied to the card upon creation as demonstrated below:

Quick insert demo

The textual tags are removed from the card title before saving to maintain sanity. If a label doesn't have a name, the name of its colour is used (i.e. `red`, `purple`, etc).

This was developed to ease the creation of new cards and allow users to rapidly create cards without having to click numerous times just to apply labels & members.

Live demo available on Heroku: https://rocky-basin-8017.herokuapp.com


zeineath:
this would be so neat! I hope this gets implemented soon.  :+1:

lebrush:
:+1: maybe in 0.10?

client/components/lists/listBody.js
```diff
@@ -36,10 +36,43 @@ BlazeComponent.extendComponent({
...
+
+      // Find all @-mentioned usernames, collect a list of their IDs and strip
+      // their mention out of the title.
+      let foundUserIds = []; // eslint-disable-line prefer-const
+      currentBoard.members.forEach((member) => {
+        const username = Users.findOne(member.userId).username;
+        const nameNdx = title.indexOf(`@${username}`);
+        if(nameNdx !== -1) {
+          foundUserIds.push(member.userId);
+          title = title.substr(0, nameNdx)
```
mquandalle: Maybe there could be a race condition here if the title is modified between the index capture and the substring (unlikely I admit). Any reason to not use

```js
title = title.replace(`@${username}`, '');
```

That seems more clear to me.

client/components/lists/listBody.js
```diff
@@ -36,10 +36,43 @@ BlazeComponent.extendComponent({
...
+      // and applying the appropriate users and labels to the card instead.
+      const currentBoard = Boards.findOne(Session.get('currentBoard'));
+
+      // Find all @-mentioned usernames, collect a list of their IDs and strip
+      // their mention out of the title.
+      let foundUserIds = []; // eslint-disable-line prefer-const
+      currentBoard.members.forEach((member) => {
+        const username = Users.findOne(member.userId).username;
+        const nameNdx = title.indexOf(`@${username}`);
+        if(nameNdx !== -1) {
```
mquandalle: I generally put I space after `if`

mquandalle:
The implementation looks good @FuzzyWuzzie, thank you! I would like to improve the following points before merging this feature:

* Pressing `Escape` when the autocomplete menu is open should close the autocomplete menu, not the entire card
* Pressing `Enter` when the card only contains some meta-data (eg `@username`) and no title should not close the card and lose its content but instead should keep the card open it does when I press `Enter` on an empty card.

Kenton please tell me if you don’t have the time or if you don’t want to handle these modifications, in this case I will work on them :-)

FuzzyWuzzie:
Thanks for reviewing! I made the suggested changes. Any other ideas we can implement with this?

FuzzyWuzzie:
I also added little coloured label badges beside each label in the dropdown:

Coloured label badges

To do so, I added a slightly modified copy of the `.minicard-label` style to the `forms.styl`. I feel like this isn't the best way to do it, but wasn't sure of the actual best way. Any suggestions?

client/components/lists/listBody.js
```diff
@@ -140,4 +175,77 @@ BlazeComponent.extendComponent({
...
+        },
+        index: 1,
+      },
+
+      // Labels
+      {
+        match: /\B#(\w*)$/,
+        search(term, callback) {
+          callback($.map(currentBoard.labels, (label) => {
+            const labelName = (!label.name || label.name === '')
```
mquandalle: I believe that even if a blue label has a name we should still display it in the autocomplete list. If there are both an unnamed blue label, and a few blue labels with names with should display the unnamed label first and named label after that (in no particular order).

mquandalle:
Good idea @FuzzyWuzzie, it looks good. Putting the styles in `forms.styl` isn’t particularly wrong (CSS in fundamentally a mess anyway :p, I'm seriously considering inline style with React components but that’s a bit of topic). After a second thought I think they are additional things to consider:

* I think we should display color names in a lighter grey (like it is done in the filter view — in the sidebar) to visually indicate that these are names we chosen by default because the actual label doesn’t have a name;
* Colors should be translated in the user language, but since it’s currently not done in the filter view, I can merge without it.
* See also my inline comment about color auto-completion on named labels

client/components/lists/listBody.js
```diff
@@ -140,4 +175,77 @@ BlazeComponent.extendComponent({
...
+        },
+        replace(label) {
+          return `#${label} `;
+        },
+        index: 1,
+      },
+    ]);
+
+    // customize hooks for dealing with the dropdowns
+    $textarea.on({
```
mquandalle: Could you add a `// XXX comment` to say that this code is duplicated from `editor.js` and that we should abstract this pattern to keep things DRY?

package.json
```diff
@@ -0,0 +1,11 @@
+{
```
mquandalle: I’m curious why you added this file?

client/components/lists/listBody.js
```diff
@@ -35,11 +35,40 @@ BlazeComponent.extendComponent({
       sortIndex = Utils.calculateIndex(lastCardDom, null).base;
     }
 
+    // Parse for @user and #label mentions, stripping them from the title
+    // and applying the appropriate users and labels to the card instead.
+    const currentBoard = Boards.findOne(Session.get('currentBoard'));
+
+    // Find all @-mentioned usernames, collect a list of their IDs and strip
+    // their mention out of the title.
+    let foundUserIds = []; // eslint-disable-line prefer-const
```
mquandalle: By the way I agree that this ESlint rule isn’t right, I will probably remove it. Actually the following is perfectly valid in ES6:

```es6
const list = [];
list.push('abc');
```

The only thing that is prevented by the `const` usage is re-assigning the value of `list`, which would raise an error. But actual mutation is still possible... that’s confusing to say the least.

Don’t modify this line, I just wanted to share my thoughts on it ;-)

package.json
FuzzyWuzzie: I added that to push it to Heroku for testing, but I didn't mean to add it to the commits, my bad!

client/components/lists/listBody.js
FuzzyWuzzie: For sure!

client/components/lists/listBody.js
FuzzyWuzzie: Huh, TIL. Makes sense in that way, but I agree its a bit confusing.

mquandalle:
I think that the autocomplete action doesn’t behave as expected, and believe the way #351 did implement it (which is also how Trello behaves) is better. The thing is that when I press tab to add `@Kenton` I would prefer to add the `Kenton` member object to the card, which is materialized as an avatar on the bottom right, rather than the string `@Kenton` in the textarea. This approach also solves some edge cases with the current approach (for instance which label should we choose if I insert `#blue` and there are multiple blue labels? Answer: the one that was selected on the autocomplete menu which mean that we need to remove the serialization and parsing step).

@FuzzyWuzzie Do you want me to work on that?

FuzzyWuzzie:
The one catch with that method is: how does the user (or can they even?) remove labels and users from a card (before submitting it). Say I'm adding cards and I accidentally add the `urgent` label when I actually meant the `low priority` label. Is there any way to remove the `urgent` label before creating the card, or does the user have to create the card, break their workflow to open the card, then remove the label in the card's details?

Perhaps capturing the labels and users like many applications use tags (once the user selects a tag, it renders it in a bubble with an `x` appended)?

mquandalle:
Hum you are right. We could have a dedicated Popover to customize the set of members (or labels) by clicking on one of them (as we do in the card pane) which would require the mouse. We can also remove using the keyboard only by “adding it again”, which mean that selecting an item in the autocomplete menu actually toggle it — which is what Trello does. I guess in this case we need to have a visual indication on the autocomplete menu that the item is already added and that its selection will remove it, such as a ✔ postfix.

client/components/lists/listBody.js
mquandalle: Since 8bbc69616f91e9d09302d8a00f845b74e37d535f you can now rename `textcomplete` into `escapeableTextComplete` and remove all the duplicated code (ie the `dropdownMenuIsOpened` declaration, the `EscapeActions` registration and these `$textarea.on` hooks)

mquandalle:
Kenton do you need help to finish this feature?

FuzzyWuzzie:
Yea, I'm not sure when I'll have time to work on it (might be a couple weeks now, unless I can squeeze it in somewhere). If you're able to sort this out, then that would be great!

mquandalle:
Yay! Finally got this merged, and this will be available in the next release of Wekan. Thank you @FuzzyWuzzie!

FuzzyWuzzie:
Awesome! Thanks for helping me start it! :smile:

PR #350, opened 2015-10-07
Generate cards

hsduk:
Gives ability to auto generate cards from an api url
Simply create a new list by entering a valid url that returns json data with a title or name, click Save and the cards will be auto generated for that list. JSON data is dumped into the card description.
Tested with github api urls: 
https://api.github.com/repos/wekan/wekan/issues?state=open
https://api.github.com/users/meteor/repos
https://restcountries.eu/rest/v1/all


mquandalle:
Thank you @hsduk for this PR. I’m not sure if I want to support that in core because that sounds like a specific use case that would make the base product more complex. Instead I would like to create a public API that would allow this kind of features via a plugin system. What do you think?

hsduk:
Hi, ignoring the additional items i mentioned todo, which add much complexity, as it is now, this is just a simple one hit http client.
I stopped working on it further as I realised it would be much more powerful when the wekan api is available (!) and could be used to pull lists and cards from any wekan install (= clone / list templates) as fast as possible.
Probably much more useful for turbo wekan users - e.g. at a hackathon type event where cards need to created/worked on by teams where every second counts. 
As it is it's pretty cool being able to load github data straight into wekan supporting turbo CI.

hsduk:
(I removed the additional todos i had mentioned to keep this PR simpler, for ref they were:
todo: 
support any json data source format, beyond name/title fields
webhooks/refresh source json data at specified intervals, e.g. every hour, updating labels as necessary
live/realtime cards, e.g. for NOC/contact centres 
server side card generation
limit number of cards created
limit data inserted into wekan - user roles)

hsduk:
p.s. Hoping to use Wekan faster than a teenage racing car driver (ref: Damani Marcano) https://www.youtube.com/watch?v=CMhx-dm54Dg :)

hsduk:
will rethink

PR #354, opened 2015-10-12
Add editorconfig support

xavierpriour:
Just added .editorconfig (http://editorconfig.org/) to help with style across different code editors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# EditorConfig is awesome: http://EditorConfig.org
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2


mquandalle:
Hi Xavier,

Our coding style is already handled by ESLint which has a lot more rules that the `.editorconfig` you propose here. I don’t see the point of supporting it given that

* you have to configure (or at least test) ESLint to contribute (otherwise your PR will get rejected by our CI);
* as far as I know, ESLint has far more traction than Editor Config — at least in the JS/ES community
* I don’t want to maintain two separate set of rules. I believe it is better for the project to be opinionated on this kind of non-fundamental issues.

xavierpriour:
Hi Maxime,

The goal is actually different: editorconfig wil just set your compatible IDE to the right style, without needing to change your settings, whereas eslint will tell you a posteriori.
It is voluntarily very limited and basic (spaces, tabs and end of file basically) so as not to duplicate with eslint. It is a cross-IDE setting rather than a set of rules or a linter.

I find both complementary, and a nice touch when arriving on a new project that my IDE automatically complies with whatever tab/space style they have. Would you be open to trying it?

mquandalle:
Ok that makes sense. Thanks for the PR and argumentation, let’s try editorconfig!

xavierpriour:
what IDE do you use btw?

mquandalle:
I use Sublime Text on Ubuntu — which I wouldn’t qualify as an IDE by the way.

xavierpriour:
You'll need the plugin then: https://github.com/sindresorhus/editorconfig-sublime
Or switch to Webstorm :)

mquandalle:
I tried Webstorm at least 2 times but got back to Sublime :-)

xavierpriour:
To each his own then - "les gouts et les couleurs..."

Issue #113, opened 2015-01-23
Support for import from Trello JSON export


singpolyma:

alireza-ahmadi:
According to [Libreboard's Roadmap](http://libreboard.com/boards/MeSsFJaSqeuo9M6bs/libreboard-roadmap), Importing data from Trello is already planned for the next release.

fnkr:
+1

mattab:
Importing data from Trello would be super awesome :-)

FYI I've developed a little PHP script that exports all data from Trello.com boards (JSON). https://github.com/mattab/trello-backup - I look forward to the day where we can migrate our trello.com boards to this free/libre project :+1:

xavierpriour:
I'll be happy to take a shot at it, if nobody's working on it yet.
@mquandalle is it the right place to discuss UI and implementation? or should this happen on the roadmap board?

mquandalle:
Hi @xavierpriour, this is the right place to discuss the feature UI/UX and implementation. As far as a know (I being the lead dev) no one has started working on it. I would be happy to help you to get this merged in core if you decide to work on it :-)

xavierpriour:
I've just had a deeper look, and here's my analysis:

## Exported data
Trello allows to export a full board (at board level: Show menu > More > Print and Export) or a single card (on a card: Share and more > Export JSON).

## Import mechanism
Importing self-contained data (card text and title, board labels) should not be an issue, but we face interesting questions when dealing with "shared" data - data that might also be referenced by elements of the wekan instance, like members, or even labels when importing a single card.

I see 3 possibilities:

* create data: that's what Trello does for labels when you move a card to another board: its labels (if any) get created in the target board. In that case, we have to be cautious of collisions (if we already have a label with the same name/color, what should happen?).
* map data: we find a corresponding, existing data. This can be sophisticated (like, try to map Trello card members to Wekan members based on mail address) or dumb (like, remove all members upon import).
* merge: a combination of both: try to map, and if you fail, create on the fly. Of course, this is the most complex scenario

Your thoughts?

## UI/UX
As a final target, I'd like to allow the user to pick and choose what to import where. For example, upon opening a "board" file, display data, then let user only select a few lists (or even cards).
But as a first goal, I'll aim to import whole files (card or board). This limits shared data problems
So that would give us two UI to discuss: importing a card, and importing a full board.
 
* For the card, I plan on adding an item to the "List Actions" menu, just above "Archive this list", as this is a pretty infrequent action
[...]

Issue #152, opened 2015-04-05
todo.txt import


Matthias84:
Hi,

I'm thinking about a migrating from a www.todotxt.com compatible file to libreboard. As the concepts seem to be very compatible (aspects, priorities, ...) it would be nice if you could support this format for import / export / sync.

xet7:
https://github.com/wekan/wekan/wiki/Sync

PR #358, opened 2015-10-14
Import a single Trello card (and its comments)

xavierpriour:
This is step 1 of #113 
We import a single card, with its title, description, comments, and labels.
Limitations (to be fixed in future PRs):

* card members are removed
* Trello history is lost
* attachments are not transferred


xavierpriour:
Oops forgot to eslint - will fix!

xavierpriour:
merged with core wekan/devel + fixed an eslint issue in one of the new files - now passing :)

client/components/lists/listHeader.jade
```diff
@@ -25,11 +25,20 @@ template(name="listActionPopup")
...
     li: a.js-close-list {{_ 'archive-list'}}
 
 template(name="listMoveCardsPopup")
   +boardLists
 
+template(name="listImportCardPopup")
+  form
+    label
+      | {{_ 'card-json'}}
+      //+editor(class="js-card-json" autofocus=true)
```
mquandalle: to remove?

client/components/lists/listHeader.js
```diff
@@ -47,6 +49,70 @@ Template.listActionPopup.events({
   },
 });
 
+Template.listImportCardPopup.events({
+  submit(evt) {
+    // 1. get the json data out of the form and parse it
+    evt.preventDefault();
+    const jsonData = $(evt.currentTarget).find('textarea').val();
+    const data = JSON.parse(jsonData);
```
mquandalle: I think it’s worth implementing error handling here before merging the feature. Maybe we could use something like [this](https://github.com/wekan/wekan/blob/758be3356cc3fb05de0a29787059b6b34aac08a2/client/components/users/userAvatar.js#L100-L106)?

client/components/lists/listHeader.js
```diff
@@ -47,6 +49,70 @@ Template.listActionPopup.events({
...
+    const cardToCreate = {
+      title: data.name,
+      description: data.desc,
+      listId: this._id,
+      boardId: this.boardId,
+      userId: Meteor.userId(),
+      sort: sortIndex,
+    };
+    // 3. map labels
+    data.labels.forEach((current) => {
```
mquandalle: `currentLabel`?

client/lib/accessibility.js
```diff
@@ -7,7 +7,7 @@
 // Without an href, links are non-keyboard-focusable and are not presented on
 // blind screen readers. We default to the empty anchor `#` href.
 function enforceHref(attributes) {
-  if (! _.has(attributes, 'href')) {
+  if (!_.has(attributes, 'href')) {
```
mquandalle: Sorry for that, I had disabled ESLint on my side :-(

mquandalle:
Hi Xavier,

Thank you for the PR! The code looks good to me. Apart from my tiny inline comments there is one bigger change I would like to consider with you before merging this PR. I don’t know how much you are familiar with Meteor, but basically there are two paths to mutate the database, one is using `Meteor.methods` that are basically RPC (but with potential optimistic update on the client) and the other is using minimongo operations (like we everywhere in Wekan). The thing is that with minimongo `.insert` we enforce some constants, for instance we ensure that the `createdAt` date is correct ([here](https://github.com/wekan/wekan/blob/758be3356cc3fb05de0a29787059b6b34aac08a2/models/cards.js#L195)) but in the case of the card import it is probably better to trust the data we are importing (or at least ensure a set of different rules that we would normally do). So I propose to do the import in a Meteor Method, because it makes it easy to chose which validation rule we want to enforce.

Does that makes sense?

client/components/lists/listHeader.jade
xavierpriour: sorry!

i18n/en.i18n.json
```diff
@@ -65,6 +65,8 @@
     "card-edit-attachments": "Edit attachments",
     "card-edit-labels": "Edit labels",
     "card-edit-members": "Edit members",
+    "card-json": "Go to a Trello card, select 'Share and more...' then 'Export JSON' and copy the resulting text",
```
mquandalle: [..] copy **and paste** the [..]?

client/components/lists/listHeader.jade
mquandalle: Don’t be sorry :-)

xavierpriour:
Perfect sense, and addresses some of my concerns mentioned in #113 
What file should I put my method definition in, though?

mquandalle:
I guess `models/cards.js` — but we’ll probably want to split this file later on.

xavierpriour:
Well, I'm going to start work on importing boards next - so mind if I create a models/trello.js right now and put the method in it?

mquandalle:
Ok, but name if `import.js` because at some point we’ll be able to import data exported from Wekan as well :-)

xavierpriour:
Sure!

mquandalle:
(Or even #152 :-))

client/components/lists/listHeader.js
xavierpriour: I'm having trouble accessing the `error` field (this.error is undefined in my method) - where is it defined?

mquandalle: [here](https://github.com/wekan/wekan/blob/758be3356cc3fb05de0a29787059b6b34aac08a2/client/components/users/userAvatar.js#L56-L58)

xavierpriour:
When generating activities for the card import itself, should it generate an 'importCard' only, or an 'importCard' + a 'createCard' (with same date)?

mquandalle:
I think only “importCard” will be less confusing.

models/import.js
```diff
@@ -0,0 +1,127 @@
+Meteor.methods({
+  /**
+   *
+   */
+  importTrelloCard(trelloCard, listId, sortIndex) {
+    // 1. check parameters are ok from a syntax point of view
```
mquandalle: :+1:

models/import.js
```diff
@@ -0,0 +1,127 @@
...
+      })],
+      members: [Object],
+    }));
+    check(listId, String);
+    check(sortIndex, Number);
+
+    // 2. check parameters are ok from a business point of view (exist & authorized)
+    const list = Lists.findOne(listId);
+    if(!list) {
+      throw 'exception-list-doesNotExist';
```
mquandalle: I think the convention with Meteor is to throw a `Meteor.Error`

xavierpriour: Yes, I'm still on it - I pushed a little too early :)

mquandalle: That’s fine, I close my eyes :-)

xavierpriour:
@mquandalle it should be good now

client/components/lists/listHeader.js
```diff
@@ -47,6 +49,45 @@ Template.listActionPopup.events({
   },
 });
 
+
+BlazeComponent.extendComponent({
+  events() {
+    return [{
+      'submit': (evt) => {
```
mquandalle: Not important, but our eslint convention is `submit(evt) {`

xavierpriour: Done!

mquandalle:
Tested and it works well! Thank you Xavier, I’m ready to merge. Maybe could you just `git rebase -i  21dcd16` and squashes some of the commits (for instance the “fix eslint style”) so we get a cleaner git history?

xavierpriour:
ok, did some squashing.

xavierpriour:
ok, something seems wrong, rebase brought your commits in my PR... how do I get them out?

mquandalle:
Yep something went wrong, I'll fix and merge.

mquandalle:
Merged in b670a1ab36fcf514751e4b356487a96088dbbd24, thanks!

PR #361, opened 2015-10-16
Make card titles accept html

rlsilber:
This change makes card titles accept html.  In particular, it means that a link cut and pasted to a card will still be clickable without requiring navigating to the details section and using the description field.

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/361)

### DIFF
--- a/client/components/cards/minicard.jade
+++ b/client/components/cards/minicard.jade
@@ -7,7 +7,9 @@ template(name="minicard")
       .minicard-labels
         each labels
           .minicard-label(class="card-label-{{color}}" title="{{name}}")
-    .minicard-title= title
+    .minicard-title
+      +viewer
+         = title
     if members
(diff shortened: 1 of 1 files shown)


mquandalle:
I don’t think this is a desirable feature for now. Closing this feature, as it could be very easily re-implemented once we have the plugin system.

rlsilber:
Thanks for letting me know.
Rachel

From: Maxime Quandalle
Reply-To: wekan/wekan
Date: Monday, February 1, 2016 at 11:52 AM
To: wekan/wekan
Cc: "Silber, Rachel"
Subject: Re: [wekan] Make card titles accept html (#361)


I don’t think this is a desirable feature for now. Closing this feature, as it could be very easily re-implemented once we have the plugin system.

PR #362, opened 2015-10-19
Import a full Trello board

xavierpriour:
This is step 2 of #113 

We import a full Trello board (with its title, color, labels, lists, cards, and comments), including all archived items (lists and cards).

Limitations (may be fixed in future PRs, depending on actual needs):

* card members are removed
* Trello activity history ('card created on xxx') is lost
* attachments are not transferred

PS: also did the following "technical stuff"

* moved all import code (including card import) into its own UI component
* in activities.jade, sorted event display code in alphabetical order to be easier to know where to add new events


mquandalle:
At a first glance the code (and its commit history) looks very good, I’m super happy to review that and get it merged in core, thank you :-)

mquandalle:
On a side note do you think something like https://github.com/gaearon/normalizr would be useful (for instance to import all the members at once)?

models/import.js
```diff
@@ -1,5 +1,285 @@
...
+      type: String,
+    })]);
+    // XXX perform deeper checks based on type
+  }
+
+  checkBoard(trelloBoard) {
+    check(trelloBoard, Match.ObjectIncluding({
+      closed: Boolean,
+      labels: [Match.ObjectIncluding({
+        // XXX check versus list
```
mquandalle: What does “check versus list” means?

models/import.js
```diff
@@ -1,5 +1,285 @@
...
+      archived: trelloBoard.closed,
+      color: this.getColor(trelloBoard.prefs.background),
+      createdAt,
+      labels: [],
+      members: [{
+        userId: Meteor.userId(),
+        isAdmin: true,
+        isActive: true,
+      }],
+      // current mapping is easy as trello and wekan use same keys: 'private' and 'public'
```
mquandalle: Hum I think Trello has one extra visibility option at the *organization* level.

models/import.js
xavierpriour: It means as an extra security, we could check that these data (color, permission level, background color) are part of a specific list - because if they are not, we won't be able to import them.
But I also think it's a "nice to have" thing.

models/import.js
xavierpriour: I have no access to a trello orga - do you know where we can get an export sample so I can see what the value is? And is there a Wekan equivalent?

mquandalle: They recently renamed the “organization” into “tersonal team”. This feature is in the free plan, you can create an organization with only one user in it (you) with the `+` icon on the header bar, and then select “Team visible” in the visibility setting.

I just tried and the `permissionLevel` is `"org"`.

xavierpriour:
Hey, I did not know about normalizr, thanks for pointing it out.

(...looking into it...)

Trello json is actually already normalized: lists, labels, members, cards are all listed at root level then only referenced by ID. I only have to build an id index, which is pretty straightforward, so I think normalizr would be overkill. But it's a good one for future imports that would not be presented as well...

models/import.js
xavierpriour: Nice. Does wekan have the same thing? Or should I map it to private?

mquandalle:
Ok, makes sense.

xavierpriour:
Updates the comments to be clearer, handles "team" permission (mapping to Wekan "private"), and a big refactor to share most code with card import.

mquandalle:
I have a bug when importing old boards (in my case 2013). From my investigations the cause seems to be that Trello didn’t have `createList` back at the time — which raise an error because the `createdAt` date of the list is then `undefined`.

mquandalle:
I fixed this issue by providing a default date.

This is now merged in `devel`. Thank you very much once again!

PR #370, opened 2015-10-27
Add ESLint-plugin-Meteor

dferber90:
Enable [ESLint-plugin-Meteor](https://github.com/dferber90/eslint-plugin-meteor).

*This PR is an example for how to properly set up ESLint-plugin-Meteor in a project.*
*That's why I included so many annotations in the source code.*
*Make sure to check out the [Files changed](/wekan/wekan/pull/370/files) tab for details.*

#### ESLint-plugin-Meteor
This PR enables the rules of ESLint-plugin-Meteor. They
- validate usage of the Meteor API
- prevent overwrites of collections
- define Meteor globals based on the context of the files are used in (browser, server, package, ..)
- [and much more](https://github.com/dferber90/eslint-plugin-meteor/tree/master/docs/rules)

#### Local versions
By adding a `package.json` npm's tools can be used for this project.

New developers now don't need to install ESLint and ESLint-plugin-Meteor globally. Instead, they can simply do `npm install` and get going. They will need to have node and npm installed for this to work, which are necessary for ESLint anyways.

Using a local version of ESLint and [ESLint-plugin-Meteor](https://github.com/dferber90/eslint-plugin-meteor) instead of relying on global versions further ensures consistency across the development team and the CI.

#### Pinned versions
The versions of ESLint and ESLint-plugin-Meteor are pinned down exactly in `package.json`. This means even if there are new versions of these plugins the specified ones will be used. Enable the free-for-open-source [greenkeeper.io](http://greenkeeper.io/) for this repository to automatically get pull-requests whenever a new version is released.

#### babel-eslint
This PR also adds the [babel-eslint](https://github.com/babel/babel-eslint) parser which makes ESLint understand all valid "babel code".

#### CI
The CI now no longer installs *eslint* and *eslint-plugin-meteor* globally. Instead it uses local versions.

```bash
# setup
$ npm install

# actual testing
$ npm test
```

`npm test` will only do `npm run lint` for now. `$ npm run lint` can also be executed during development to check which files have errors and resolve them before sending a PR and having the CI figure it out.

*Happy linting* :)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,14 @@
 ecmaFeatures:
   experimentalObjectRestSpread: true
 
+plugins:
+  - meteor
+
+parser: babel-eslint
+
 rules:
+  strict: 0
+  no-undef: 2
   accessor-pairs: 2
   comma-dangle: [2, 'always-multiline']
   consistent-return: 2
@@ -43,36 +50,35 @@ rules:
   prefer-spread: 2
   prefer-template: 2
 
-globals:
-  # Meteor globals
-  Meteor: false
-  DDP: false
-  Mongo: false
-  Session: false
-  Accounts: false
-  Template: false
-  Blaze: false
-  UI: false
-  Match: false
-  check: false
-  Tracker: false
-  Deps: false
-  ReactiveVar: false
-  EJSON: false
-  HTTP: false
-  Email: false
-  Assets: false
-  Handlebars: false
-  Package: false
-  App: false
-  Npm: false
-  Tinytest: false
-  Random: false
-  HTML: false
+  # eslint-plugin-meteor
+  meteor/globals: 2
+  meteor/no-zero-timeout: 2
+  meteor/no-session: 0
+  meteor/pubsub: 2
+  meteor/core: 2
+  meteor/methods: 2
+  meteor/check: 2
+  meteor/connections: 2
+  meteor/collections: 2
+  meteor/session: [2, 'no-equal']
 
+settings:
+  meteor:
+
+    # Our collections
+    collections:
+      - AccountsTemplates
+      - Activities
+      - Attachments
+      - Boards
+      - CardComments
+      - Cards
+      - Lists
+      - UnsavedEditCollection
+      - Users
+
+globals:
   # Exported by packages we use
-  '$': false
-  _: false
   autosize: false
   Avatar: true
   Avatars: true
@@ -96,17 +102,6 @@ globals:
   SubsManager: false
   T9n: false
   TAPi18n: false
-
-  # Our collections
-  AccountsTemplates: true
-  Activities: true
-  Attachments: true
-  Boards: true
-  CardComments: true
-  Cards: true
-  Lists: true
-  UnsavedEditCollection: true
-  Users: true
 
   # Our objects
   CSSEvents: true
--- a/.eslintrc
+++ b/.eslintrc
@@ -51,16 +51,20 @@ rules:
   prefer-template: 2
 
   # eslint-plugin-meteor
+  ## Meteor API
   meteor/globals: 2
-  meteor/no-zero-timeout: 2
-  meteor/no-session: 0
-  meteor/pubsub: 2
   meteor/core: 2
+  meteor/pubsub: 2
   meteor/methods: 2
   meteor/check: 2
   meteor/connections: 2
   meteor/collections: 2
   meteor/session: [2, 'no-equal']
+
+  ## Best practices
+  meteor/no-session: 0
+  meteor/no-zero-timeout: 2
+  meteor/no-blaze-lifecycle-assignment: 2
 
 settings:
   meteor:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: node_js
 node_js:
   - "0.10.40"
 install:
-  - "npm install -g eslint"
-  - "npm install -g eslint-plugin-meteor"
+  - "npm install"
 script:
-  - "eslint ./"
+  - "npm test"
--- a/models/attachments.js
+++ b/models/attachments.js
@@ -1,4 +1,4 @@
-Attachments = new FS.Collection('attachments', {
+Attachments = new FS.Collection('attachments', { // eslint-disable-line meteor/collections
   stores: [
 
     // XXX Add a new store for cover thumbnails so we don't load big images in
--- a/models/users.js
+++ b/models/users.js
@@ -1,4 +1,4 @@
-Users = Meteor.users;
+Users = Meteor.users; // eslint-disable-line meteor/collections
 
 // Search a user in the complete server database by its name or username. This
 // is used for instance to add a new user to a board.
@@ -7,6 +7,24 @@ Users.initEasySearch(searchInFields, {
   use: 'mongo-db',
   returnFields: [...searchInFields, 'profile.avatarUrl'],
 });
+
+if (Meteor.isClient) {
+  Users.helpers({
+    isBoardMember() {
+      const board = Boards.findOne(Session.get('currentBoard'));
+      return board &&
+        _.contains(_.pluck(board.members, 'userId'), this._id) &&
+        _.where(board.members, {userId: this._id})[0].isActive;
+    },
+
+    isBoardAdmin() {
+      const board = Boards.findOne(Session.get('currentBoard'));
+      return board &&
+        this.isBoardMember(board) &&
+        _.where(board.members, {userId: this._id})[0].isAdmin;
+    },
+  });
+}
 
 Users.helpers({
   boards() {
@@ -21,18 +39,6 @@ Users.helpers({
   hasStarred(boardId) {
     const {starredBoards = []} = this.profile;
     return _.contains(starredBoards, boardId);
-  },
-
-  isBoardMember() {
-    const board = Boards.findOne(Session.get('currentBoard'));
-    return board && _.contains(_.pluck(board.members, 'userId'), this._id) &&
-                         _.where(board.members, {userId: this._id})[0].isActive;
-  },
-
-  isBoardAdmin() {
-    const board = Boards.findOne(Session.get('currentBoard'));
-    return board && this.isBoardMember(board) &&
-                          _.where(board.members, {userId: this._id})[0].isAdmin;
   },
 
   getAvatarUrl() {
--- a/package.json
+++ b/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "wekan",
+  "version": "1.0.0",
+  "description": "The open-source Trello-like kanban",
+  "private": true,
+  "scripts": {
+    "lint": "eslint .",
+    "test": "npm run lint"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/wekan/wekan.git"
+  },
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/wekan/wekan/issues"
+  },
+  "homepage": "https://github.com/wekan/wekan#readme",
+  "devDependencies": {
+    "babel-eslint": "4.1.3",
+    "eslint": "1.7.3",
+    "eslint-plugin-meteor": "1.5.0"
+  }
+}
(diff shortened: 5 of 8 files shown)


.eslintrc
```diff
@@ -1,7 +1,14 @@
 ecmaFeatures:
   experimentalObjectRestSpread: true
 
+plugins:
+  - meteor
```
dferber90: This enables ESLint-plugin-Meteor

.eslintrc
```diff
@@ -1,7 +1,14 @@
...
+
+parser: babel-eslint
```
dferber90: Enable babel-eslint so all valid "babel code" can be linted.

.eslintrc
```diff
@@ -43,36 +50,35 @@ rules:
...
-  Email: false
-  Assets: false
-  Handlebars: false
-  Package: false
-  App: false
-  Npm: false
-  Tinytest: false
-  Random: false
-  HTML: false
+  # eslint-plugin-meteor
```
dferber90: All rules of ESLint-plugin-Meteor are turned off by default. The rules below enable them.

.eslintrc
```diff
@@ -43,36 +50,35 @@ rules:
...
+  meteor/globals: 2
+  meteor/no-zero-timeout: 2
+  meteor/no-session: 0
```
dferber90: There was a [proposal](https://meteor.hackpad.com/Proposal-Deprecate-Session-in-favor-of-ReactiveDict-0wbRKtE4GZ9) to deprecate usage of `Session`. In case any application followed this proposal, using `Session` can be explicitly forbidden with this rule.

The rule is disabled by default, so this line is redundant. All other rules are set to error out.

.eslintrc
```diff
@@ -43,36 +50,35 @@ rules:
...
+  meteor/pubsub: 2
+  meteor/core: 2
+  meteor/methods: 2
+  meteor/check: 2
+  meteor/connections: 2
+  meteor/collections: 2
+  meteor/session: [2, 'no-equal']
```
dferber90: Did you ever write `Session.equal('foo', bar)` and wonder why it didn't work? Well, I certainly did.

The option `no-equal` to this rule makes the rule prevent that and tells you to use `Session.equals` instead.

.eslintrc
```diff
@@ -43,36 +50,35 @@ rules:
...
+
+settings:
+  meteor:
 
+    # Our collections
+    collections:
```
dferber90: By telling ESLint-plugin-Meteor about the names of the collections, they will automatically be registered as globals.

Furthermore using `collection.find()` etc will then be linted. And assigning something other than `new Mongo.Collection()` to a collection will error out, because most likely you are doing something wrong then; although there are valid cases like in this project.

models/attachments.js
```diff
@@ -1,4 +1,4 @@
-Attachments = new FS.Collection('attachments', {
+Attachments = new FS.Collection('attachments', { // eslint-disable-line meteor/collections
```
dferber90: Example of a valid case where something other than `new Meteor.Collection()` is assigned to a collection.

models/users.js
```diff
@@ -8,6 +8,24 @@ Users.initEasySearch(searchInFields, {
   returnFields: [...searchInFields, 'profile.avatarUrl'],
 });
 
+if (Meteor.isClient) {
```
dferber90: These helpers access `Session` which is only defined on the client. Therefor they should be defined on the client only. Using them on the server would throw an error because `Session` is not defined there.

package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "wekan",
+  "version": "1.0.0",
+  "description": "The open-source Trello-like kanban",
+  "private": true,
```
dferber90: Marking a package as `private` will suppress multiple warnings of npm. It will also make sure you never accidentally `$ npm publish` this repository as an npm package.

package.json
```diff
@@ -0,0 +1,24 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/wekan/wekan.git"
+  },
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/wekan/wekan/issues"
+  },
+  "homepage": "https://github.com/wekan/wekan#readme",
+  "devDependencies": {
```
dferber90: All dependencies are pinned down to their exact versions to ensure consistency across development environments and the CI.

sandstorm.js
```diff
@@ -109,7 +109,7 @@ if (isSandstorm && Meteor.isClient) {
   // sandstorm client to return relative paths instead of absolutes.
   const _absoluteUrl = Meteor.absoluteUrl;
   const _defaultOptions = Meteor.absoluteUrl.defaultOptions;
-  Meteor.absoluteUrl = (path, options) => {
+  Meteor.absoluteUrl = (path, options) => { // eslint-disable-line meteor/core
```
dferber90: Usually ESLint-plugin-Meteor warns when any part of the Core-API is overwritten. Since monkey-patching is used here, and we seem to know what we are doing, the rule is disabled :-)

.eslintrc
```diff
@@ -1,7 +1,14 @@
 ecmaFeatures:
   experimentalObjectRestSpread: true
 
+plugins:
+  - meteor
+
+parser: babel-eslint
+
 rules:
+  strict: 0
```
dferber90: Explicitly disabling the `strict`-rule is part of `babel-eslint`.

.eslintrc
```diff
@@ -1,7 +1,14 @@
...
+  no-undef: 2
```
dferber90: Enabling `no-undef`, so ESLint warns when undefined variables are used. This works in conjunction with `meteor/globals`, which defines Meteor's global variables (like `Meteor` and `Session`) depending on the environment of the file (`/client`, `/server`, or no special folder).

So `Session` will be defined in a file in `/client`, but not in a file in `/server`. In the latter case `no-undef` will warn when accessing  `Session`.

.eslintrc
```diff
@@ -43,36 +50,39 @@ rules:
...
+  meteor/core: 2
+  meteor/pubsub: 2
+  meteor/methods: 2
+  meteor/check: 2
+  meteor/connections: 2
+  meteor/collections: 2
+  meteor/session: [2, 'no-equal']
+
+  ## Best practices
+  meteor/no-session: 0
```
dferber90: There was a proposal to deprecate usage of `Session`. In case any application followed this proposal, using `Session` can be explicitly forbidden with this rule.

The rule is disabled by default, so this line is redundant. All other rules are set to error out.

mquandalle:
Hey @dferber90, thank you for this canonical Eslint-meteor integration that I’m happy to merge as is.

I guess it would be great if we could integrate the Eslint linter (including the eslint-meteor extension) with the meteor build system, so one could simple `meteor lint`. The main benefit would be meteor will warn them about linting errors while developing — which would avoid sending pull requests with wrong formating and then having to commit again with “Fix eslint” as most contributors have done.

.travis.yml
```diff
@@ -3,7 +3,6 @@ language: node_js
 node_js:
   - "0.10.40"
 install:
-  - "npm install -g eslint"
-  - "npm install -g eslint-plugin-meteor"
+  - "npm install"
 script:
-  - "eslint ./"
+  - "npm test"
```
mquandalle: I think this doesn’t work as intented because even if some tests fails the unix return code will be `0` instead of a greater number as it does with `eslint ./` or even `npm run lint`. Could you verify @dferber90?

mquandalle: Commit dd10154 fixed it for me

dferber90: > I think this doesn’t work as intented because even if some tests fails the unix return code will be 0 instead of a greater number

As `npm test` is intended exactly for this, it should work. I just verified locally and it worked as it should (without the silent option).

> `eslint ./`

This should no longer be used because it will use the globally installed version of ESLint.

> I guess it would be great if we could integrate the Eslint linter (including the eslint-meteor extension) with the meteor build system, so one could simple meteor lint

I have thought about that and it is definitely possible. I have taken it into account while developing eslint-plugin-meteor. But `meteor lint` is also very slow, because the app has to be built before getting any results. People can run `$ npm run lint` or `$ npm test` before sending PRs. This will be way faster, because it doesn't need to compile the whole app first. When the linter is integrated into the editor properly, it's so much better because you get errors immediately during development. Having a plugin for `meteor lint` is less convenient because the errors are in the console and not where you develop.

Still, somebody could write a meteor build plugin which uses the linter and the ESLint-plugin-Meteor behind the scenes. It would mainly be great for beginners or big projects like yours. I think it's much work for not-that-much benefit. It may be better to educate people on how to use npm properly instead of building tools which obscure everything. :)

mquandalle: Hum, I am pretty sure when I tested yesterday `npm test; echo $?` always gave me `0` even if there was some errors, but as I tested it again today I can’t reproduce it, so I suppose I was doing something wrong... Anyway I prefer the output with the `--silent` option because I don’t see the point of having 20 lines of `npm ERR!` to explain that is probably a problem with Wekan and not with NPM.

About Meteor integration, IIRC one argument if favor of it is that it print lint warning in development server started by `meteor run`, so that’s would definitely improve the contributor experience that won’t have to write a lot of LOC before realizing they using different code styling conventions.

Also ESLint recently introduced a `--fix` option to automatically fix trivial formating issues. How could we use that tool if the command line `eslint` refer to the global version that misses some plugins instead of using the local one?

dferber90:
> Also ESLint recently introduced a --fix option to automatically fix trivial formating issues. How could we use that tool if the command line eslint refer to the global version that misses some plugins instead of using the local one?

Fixing only works for a small subset of the rules. Autofixing also isn't implemented in ESLint-plugin-Meteor, because that is too much work. You can add an npm script like `npm run lint:autofix` which just passes the options to eslint like you would with a globally installed version of ESLint.

> About Meteor integration, IIRC one argument if favor of it is that it print lint warning in development server started by meteor run, so that’s would definitely improve the contributor experience that won’t have to write a lot of LOC before realizing they using different code styling conventions.

I agree. As I said, if somebody builds it, that'd maybe introduce the linter to more people. The way it should work is that it bundles `eslint` and `eslint-plugin-meteor`,  reads in all the configuration files, merges them and then executes `eslint`. So basically replicating the `eslint` command-line tool functionality.
That person building it just has to make the benefits of having the linter right in the editor extra clear. The reason I've built this plugin how it is right now and not as a Meteor plugin is so that the editor can use it as well. This won't work trivially with Meteor lint plugins (or if it does, it will be way too slow, because Meteor has to rebuilt).

mquandalle:
Yes, I was not suggesting to replace a Eslint plugin by a meteor one, but I was just saying that I would like to *add* something like https://github.com/dburles/meteor-eslint/ that would work with the current configuration.

dferber90:
Got it :) Was just explaining some thoughts around that.

I had a proof-of-concept of what you're suggesting already so it should be possible :+1:

mquandalle:
Maybe there is a Meteor package that could run `npm install` automatically when Meteor build the application (I think `webpack:webpack` does that for instance, but along a lot of different things). So the user won’t forget to run this command (and upgrades would be handled automatically by the build system). This would also help some of the meter developers that don’t play with raw nodejs and never installed `npm` before by using the `npm` version installed by Meteor.

mquandalle:
@dferber90 could you confirm that I should revert these changes given that you abandonned meteor-eslint?

dferber90:
@mquandalle Yes, you may revert them. I'm sorry for causing the noise. ESLint-plugin-Meteor broke because of some changes in ESLint. ESLint-plugin-Meteor would also need some sort of rewrite to work with the 1.3 modules. I couldn't get any response from MDG back then to prepare the plugin for the changes which was discouraging. Since I'm no longer using Meteor, I don't want to spend the time fixing it.

mquandalle:
Hey don’t be sorry for “the noise”, Meteor-specific static analysis using ESLint was a great initiative even if the ever moving JS ecosystem and the lack of MDG broke your work and discouraged you. I just wanted to confirm that I should remove it in Wekan as well to prevent contributors experiencing known bugs.

PR #389, opened 2015-11-05
Simple timer using markdown

hsduk:
image

Click Card Menu -> Tools to toggle

Simple timer plugin for customising cards. Uses markdown to build the template.
Click to stop/start
Double click to reset and add a comment
Edit plugin.tpl.md file to customise

+ Added toolbar and card copy/clone template - does not copy attachments, labels, activities yet
+ Just select a board and then a list to copy the current card to that list. If lists is empty, simply visit the board (so list data is retrieved) and try again


hsduk:
Removed iframe, Added Tools section using https://atmospherejs.com/simple/markdown-templating
Todo: dynamic/reactive templating where the md is stored in mongo board/list/card collection and not on the fs
Todo: edit the template md on the fly from within the card like the desc field so parameters can be passed to the template
Todo: show a different tools template depending on card label / isboardadmin

hsduk:
Updated to include card copy/clone fx

hsduk:
awaiting new plugin system definition for now

huiwang:
is this timer working ? i tried the demo in sandstorm but i didn't find it.

PR #397, opened 2015-11-13
Added several features

floatinghotpot:
@mquandalle I know you prefer PR per feature... 

just let you know that I've implemented quite a few features according to our team requirement, but I think it might be common needs. 
* preview attached image & allow upload image from clipboard or drag&drop.
* add manHour, dueDate, startDate, finishDate to card.
* allow define status (todo/doing/done) in list settings, when card dropped into it, the startDate/finishDate will be auto updated accordingly.
* import cards from list of another board.
* import board template from another board, including members, labels, list titles.
* import Redmine exported CSV file, allow data mapping.
* when create card, if paste excel CSV/TSV in Redmine CSV format, batch cards will be created.

If you are interested in some (or all) of them, please let me know. 

You don't have to merge this large PR, as I can create separate PR for each of the features.


mquandalle:
Thank you for that, I’m super happy that people like you can hack Wekan for their needs and get some changes merged back. I’ll have to do a careful review of each feature to see how they fit in my plans for Wekan (especially I’d like to design a plugin system, so I need to think about what should be in core and what to expose to plugins to make the kind of features developers and users want possible). Expect to read a more specific review from me soon.

floatinghotpot:
@mquandalle Thanks for the positive feedback. 

Actually I've added more than that, even including integration with QQ (a widely used IM by 600 million people in China) for notification when card is assigned, moved or commented. But as QQ is not common in other countries, I did not put it in PR.

I really appreciate your amazing work, especially the excellent architecture, make it quite easy to add features.

Looking forward to your review comments.

jadeqwang:
:+1: :+1:

Issue #299, opened 2015-09-14
Switching to React


mquandalle:
Wekan currently use Blaze as its view layer, a library written by Meteor Development Group that aims to be easy to use for developers. I've been considering replacing Blaze by React for a few months now, and except the migration cost I can see a lot of advantages in doing so:
- Have a clean component API (I currently use blaze-component, which makes things less worse than “vanilla”-blaze and will facilitate the switch to react);
- Embrace the React eco-system which is very prolific in great tools, library, and components. I'm especially interested in [react-dnd](https://github.com/gaearon/react-dnd) ([great demo](http://cdn.rawgit.com/chenglou/react-motion/master/demos/demo8/index.html)) because I already spent too much time fighting with jquery-ui, sortable, and others;
- Functional programming, I'm especially interested in composition possibility for potential plugins made by the community;
- Attract React developers by being a great example of an open-source React application;
- _In the future_: React native, for native mobile applications (see #480);
- _In the future_: GraphQL and Relay integration for data fetching, as an improvement to our current DDP subscription model.

Despite the (upcoming) official React support by MDG, Meteor build system still misses a lot of requirements to facilitate this transition (imports/exports, easy npm/browserify usage, redux-like hot reloading, etc.). I'll probably need to investigate community solutions such as [meteor-react-pack](https://github.com/jedwards1211/meteor-webpack-react) or [meteor-webpack-react-kickstart](https://github.com/thereactivestack/meteor-webpack-react-kickstart).

We will probably need to run both Blaze and React components at the same time during the transition, which seems to work well with a few helpers functions. Also our templates are currently written in Jade, but the React community tend to prefer JSX, which mean that we will probably want to rewrite our Jade templates in JSX.

nata-g:
@mquandalle ...The reasoning seems solid and with the kind of muscle Facebook has, embracing react ecosystem is the way to go, as they seem to be controlling the current popular ideas in web/mobile development and also increases the project's exposure to other React developers (as you mentioned). However, based on my understanding,  most of the packages in the meteor universe are closely tied to Blaze and I was wondering how is the impact going to be when using React?.

mquandalle:
@nata-g Actually we don’t use a lot of packages that are tightly coupled with Blaze, and for those who are (eg `templates:tabs`) there exists better alternatives in the React ecosystem.

[...]

PR #401, opened 2015-11-14
Import Trello: include members and attachments

xavierpriour:
Adds the ability to map exported Trello members to existing Wekan users before performing the import.
It also downloads the existing attachments and sets proper card covers.

PS: @mquandalle I have a residual UI bug, I could not manage to get a default focus() on the user easy-search field to select mapping - could you have a look? client/components/import/import.js:251


mquandalle:
Hey Xavier, thank you very much for yet another great contribution :-) I’ll review it this week-end.

.meteor/release
```diff
@@ -1 +1 @@
-METEOR@1.2.1
+METEOR@1.2.2-cdn-url
```
mquandalle: Hum, could you explain that? Should I merge it?

client/components/import/import.js
```diff
@@ -11,48 +11,105 @@ const ImportPopup = BlazeComponent.extendComponent({
...
-            } else {
-              Filter.addException(response);
-              this.onFinish(response);
-            }
-          }
-        );
-      },
-    }];
+  jsonText() {
+    return Session.get('import.text');
```
mquandalle: Since Session is a global store I totally like your namespacing convention, I will use it in the future — until eventually we have better build-in solution. :+1:

client/components/import/import.js
```diff
@@ -11,48 +11,105 @@ const ImportPopup = BlazeComponent.extendComponent({
...
+    try {
+      dataObject = JSON.parse(dataJson);
+      this.setError('');
+    } catch (e) {
+      this.setError('error-json-malformed');
+      return;
+    }
+    // if there are members listed in the import and we have no mapping for them...
+    if(dataObject.members.length > 0 && !this.membersMapping()) {
+      // we will work on the list itself (an ordered array of POJO)
```
mquandalle: [POJO](https://fr.wikipedia.org/wiki/Plain_Old_Java_Object)? I’m not familiar with this acronym, I would rather use “JavaScript Object” (or “Dictionary” as in Python, but the “JavaScript Object” is clearer).

client/components/import/import.js
```diff
@@ -11,48 +11,105 @@ const ImportPopup = BlazeComponent.extendComponent({
...
+    Popup.open('mapMembers')(evt);
+  },
+
+  _storeText() {
+    const dataJson = this.$('.js-import-json').val();
+    Session.set('import.text', dataJson);
+    return dataJson;
+  },
+
+  onSubmit(evt){
```
mquandalle: This submit function looks a bit complex to me, could you move the mapping logic in a separate method?

client/components/import/import.js
```diff
@@ -88,3 +145,110 @@ ImportPopup.extendComponent({
...
+    this.mapSelectedMember(this.currentData());
+    Popup.back();
+  },
+  events() {
+    return [{
+      'click .js-select-import': this.onSelectUser,
+    }];
+  },
+  onRendered() {
+    // todo XXX why do I not get the focus??
```
mquandalle: I will look at it, but I had some similar issues in the past. It’s probably an issue with reactivity (ie the `input` is redrawn after the `focus()` call, or something like that).

models/import.js
```diff
@@ -90,6 +95,24 @@ class TrelloCreator {
...
     };
+    // now add other members
+    if(trelloBoard.memberships) {
+      trelloBoard.memberships.forEach((trelloMembership) => {
+        const trelloId = trelloMembership.idMember;
+        // do we have a mapping?
+        if(this.members[trelloId]) {
+          const wekanId = this.members[trelloId];
+          // do we already have it in our list?
+          if(!boardToCreate.members.find((wekanMember) => {return (wekanMember.userId === wekanId);})) {
```
mquandalle: Style note: I would write it either:

```es6
if (!boardToCreate.members.find((wekanMember) => {
  return (wekanMember.userId === wekanId);
})) {
```

or

```es6
if (!boardToCreate.members.find((wekanMember) => wekanMember.userId === wekanId)) {
```

With a preference for the second one.

.meteor/release
xavierpriour: Sorry, I don't know where this came from. Probably a `meteor update` gone too far. I'm removing it.

client/components/import/import.js
xavierpriour: Thanks. I'm pretty disappointed MDG did not provide a cleaner solution when they built components.

client/components/import/import.js
xavierpriour: Sorry, old habit. I'm replacing with 'object'.

client/components/import/import.js
xavierpriour: done

models/import.js
xavierpriour: Went for the later.

xavierpriour:
@mquandalle all fixed.
Sorry about the muddy git history but when I rebase, I get errors when I push to github

xavierpriour:
ping @mquandalle :wave:

mquandalle:
Hey Xavier,

Sorry for the long delay, I’m currently traveling and I’m not as productive as I would like (as you can see in my GitHub profile calendar). This is now merged, thank you again :-)

mquandalle:
Postponed the focus issue since there is a high chance that we move away from Blaze to React (see #299) sooner than latter which would change the rendering engine runtime, let's keep the `XXX` comment as you did for now.

PR #403, opened 2015-11-18
set sandstorm grain URL when router changes

mnutt:
Sandstorm supports a grain sending `postMessage` on the parent window with `setPath` to update the outer URL. This allows the outer URL to change as you browse around within the grain, and if you reload, you'll end up on the right page.

This pull request causes every route entered to update the browser's URL. In an ideal world we wouldn't see the `/sandstorm` part, but that would involve quite a bit more route hacking than has been done so far to integrate with Sandstorm.

### DIFF
--- a/client/config/router.js
+++ b/client/config/router.js
@@ -6,7 +6,7 @@ FlowRouter.triggers.exit([({path}) => {
 FlowRouter.triggers.enter([({path}) => {
   if (window.parent) {
     window.parent.postMessage({
-      setPath: path
+      setPath: path,
     }, '*');
   }
 }]);
--- a/client/config/router.js
+++ b/client/config/router.js
@@ -3,6 +3,14 @@ FlowRouter.triggers.exit([({path}) => {
   previousPath = path;
 }]);
 
+FlowRouter.triggers.enter([({path}) => {
+  if (window.parent) {
+    window.parent.postMessage({
+      setPath: path
+    }, '*');
+  }
+}]);
+
 FlowRouter.route('/', {
   name: 'home',
   triggersEnter: [AccountsTemplates.ensureSignedIn],


paulproteus:
I'm excited about this.

kentonv:
If we let the `/sandstorm` part leak now, does that mean we can never remove it without breaking a lot of existing links? It seems worthwhile to fix that now rather than get stuck with a bad URL schema.

mquandalle:
Hey @mnutt, I didn’t know about this Sandstorm feature. I will look at the exact rooting scheme we need for Sandstorm.

paulproteus:
BTW I documented this today: https://docs.sandstorm.io/en/latest/developing/path/

I guess we're at the phase where we bikeshed the default name of the board at https://github.com/wekan/wekan/blob/devel/sandstorm.js#L10 . My vote is "Make no change to it".

If others have opinions I guess speak up or decide not to, so @mquandalle can merge this and we can give @gwillen the deep linking he wants.

mquandalle:
Hi, this is now implemented in devel, thank you for the help!

I tried to change the routing scheme for the Sandstorm version, especially by removing the board name from it since the grain contain a single board, but that introduces some ambiguity with other routes (for instance we need to be sure that the `/shortcuts` route doesn’t match a `/:cardId` route, which it does not currently because the card route is `/b/:boardId/:cardId`. I still would like to clean the path a little bit but I first need to set some redirection mechanism for old URLs.

gwillen:
Thank you! This is awesome! :+1:

gwillen:
Question: Does it make sense for us (in production usage of wekan) to build and run devel to get this feature? Or would it be smarter to wait for a new release? (I understand that our sandstorm server should auto-update when there's a new release available in the Sandstorm app store?)

Thanks again!

mquandalle:
It is better to wait for the new release (or at least the release canditate). I don’t have any definite plan since I’m traveling and working at the same time, but devel has enough new features to publish a new release so I’ll work on that very soon.

PR #338, opened 2015-10-01
Various accessibility fixes

ndarilek:

PR #408, opened 2015-11-24
Without href attributes, links are invisible to screen readers

ndarilek:


mquandalle:
Hi Nolan,

Thank you for contributing! Actually this issue was already resolved in 3f7df34 but the fix isn’t released yet.

ndarilek:
Is this fix not in the devel branch then? Pretty sure I submitted a 
similar PR a while back and was also told it was fixed, but pulling the 
devel branch as of a couple days ago I still saw lots of missing hrefs. 
Now I'm told they're fixed again.

Which branch contains this fix along with the current state of the code? 
Unfortunately, as of a few days ago when I made this change, the code in 
devel is unusable to me as a screen reader user. This fix makes it 
usable, but barely. If this and other patches have already landed, I'd 
like to assess the current state of things so I can make the 
modifications necessary so I can start using Wekan. Master appears even 
further behind.

Thanks.

mquandalle:
Yes 3f7df34 is in `devel`, so I guess there is an issue remaining. Sorry for closing this issue without a proper investigation. I guess the PR you are referring to is #338. By the way, `devel` is indeed the current state of things, and `master` is the last released version (currently `v0.9`).

ndarilek:
#338 is indeed the PR, but since it never got merged, I decided to attempt something of smaller scope and just add missing hrefs. But there is still lots of work to be done for Wekan to be usable by me as a screen reader user.

Any idea when or if this PR will merge? I've noticed there is already one conflict. I'd really like to move on to other necessary a11y fixes because I'd like to use Wekan in my organization. It would be more ideal if you could incorporate an accessibility checker into your CI processes.

Thanks.

mquandalle:
I’m all for adding `href="#"` to links if that improve the situation on screen readers, I just don’t think that adding them manually to every link in every template is the correct solution because we will likely forget to add some (whether now or in the future) and in case we want to do something else in the future in make evolution harder.

Commit 3f7df34 was supposed to fix that problem by automating the insertion of a `href="#"` to links that doesn’t have a `href` already, and I verified that it worked by opening the Developers Tools and looking for this `href` attributes in the DOM. I don’t know how to reproduce the exact problem that we are facing, what screen reader are you using?

ndarilek:
Ah, gotcha. Apologies, I didn't realize you were automating the process, 
and I'm reading these replies in Thunderbird which doesn't link to the 
commits so there wasn't an easy way to check.

I'll move on to other a11y fixes, then. When I found lots of 
difficult-to-reach links, I went to the template source and found lots 
of missing hrefs, so assumed that was the issue. The problem may be that 
many of these links use icons and lack aria-labels. I'll submit another 
PR that adds those to see if that fixes things.

Thanks.

mquandalle:
Well 3f7df340e1dea4a48bd17c189ccf61e53c47330a is also supposed to automatically add `aria-label` on both `<a>` links and `<i>` icons...

ndarilek:
I don't see lots of aria-labels. In particular:

  * I see an icon with class `fa fa-angle-left` with no label.
  * Likewise for the home icon in the top list
  * Also some icons classed "fa fa-chevron-down".
  * Same for "fa fa-lock".

That's after a couple minutes of looking. There are probably many more, 
but I don't have the time to look for them, especially since I've 
already fixed them all. Imagine all of that functionality presented to 
me as a screen reader user simply as "link", "link link link link". :) 
That significantly diminishes Wekan's usability and is why I'm trying to 
get these PRs in. Automation is nice but it isn't working here.

StommePoes:
"I’m all for adding  href="#"  to links if that improve the situation on screen readers"

It's not only screen readers, in many browsers no href means they can't even be focussed upon.
This locks out all sorts of users. However also clicking a link with href="#" moves focus to the top of the page, which can also suck. If they are really links that don't go anywhere, one trick/hack is to use a target string that doesn't exist on the page (like #void) which holds focus on the link.

If you just need a clickable though that does stuff on a page, a button may be a better element. They let you script in onclicks (and key listeners for enter/space) and don't need an href to be focusable. That might make more sense than automating href="#" to a bunch of links doing button duty. And CSS nowadays can totally unbuttonify buttons if that was a concern (used to be an issue with old IEs, but everyone works pretty well nowadays).

MichielBijl:
This adding of `href`'s to anchors seems unintuitive. Especially since this looks like a proper use case for `button`'s. Buttons are designed to do nothing unless you tell them to do something ;) In the code there is an anchor with the class `js-open-inlined-form`. From that name I guess it opens some sort of form. It performs an action on the current page; that is what buttons are for. Anchors—or anchors with a `href` to be precise—are used to link to pages or parts of pages (internal links for example).

I would be very interested in the reasoning behind the use of anchors.

* * *

As for the icons, Font Awesome is inaccessible on its own; authors have to add bits of information to make them accessible. As it happens there is a project that is trying to figure out what those bits are: [Font Awesome Accessibility Tests](https://github.com/MichielBijl/font-awesome). We are very close to a definite solutions. The closest we've gotten is [`<span aria-label="love" role="img" class="fa fa-heart">`](https://cdn.rawgit.com/MichielBijl/font-awesome/master/tests/index.html#test-4). There is still research to be done, but this is where we are now.

To avoid having to deal with inaccessible icon fonts altogether; I suggest the use of SVG icons. They offer the same benefits as icon fonts, plus you can tailor them to your needs—I'm guessing Wekan includes all FA icons. For more information on inaccessible icon fonts I would like to [my icon font resource list](https://github.com/MichielBijl/font-awesome#resources) and the [Death to Icon Fonts talk](https://www.youtube.com/watch?v=9xXBYcWgCHA) by @ninjanails in particular.

mquandalle:
> Automation is nice but it isn't working here.

So 3f7df34 is doing two things:

* Add a `href="#"` attribute to links that doesn’t have it already
* Add a `aria-label` to icons and links that have a title set

I think these two features are working as expected. The problem is that we need to set some title/aria-label from every relevant element (but this PR only add `href="#"` to links). Did you say that you have a PR ready to do just that @ndarilek?

mquandalle:
For icons `aria-labels` we could use https://www.npmjs.com/package/react-accessible-fa, considering that I plan to switch to React soon after v0.10, cf. #299.

MichielBijl:
>  The problem is that we need to set some title/aria-label from every relevant element

The problem is that Wekan uses anchors where it should use `button`'s.

ndarilek:
I don't have a PR ready for the icon labels.

I guess I'm a bit wary of putting more work into this because, to my 
mind, these are very basic fixes and it's a bit odd to be discussing 
them for as long as we have been. Links without hrefs are broken markup, 
and it seems odd that you're automating fixing broken markup rather than 
putting in infrastructure to check for it to begin with, or encouraging 
that people not use it. If I submitted some other type of broken HTML, I 
imagine I'd be asked to fix it. Similarly, your automation isn't adding 
the aria-label to the `<i/>` tags containing the icon class. It seems 
far more productive to put in place methodologies for designing 
accessibly and encourage those going forward, than to develop an 
automated solution that will break when you transition to React anyway, 
thus undoing all accessibility fixes.

I just don't have a lot more time to put into labeling icons and making 
elements keyboard-accessible.

mquandalle:
I’m closing this PR as there the proposed modifications are already implemented. I’ll report the rest of my accessibility work in #408.

PR #410, opened 2015-11-27
bugfix: 'template.$()' does not accept dom element, but jQuery can

floatinghotpot:
Bug behavior: sometimes dragging lists fails, and console report error sortable() on null. 
According to Meteor doc, ```template.$(selector)``` accepts selector (only?), while ```template.find()``` return DOM element, so ```self.$(self.listsDom)``` often returns null.
jQuery $() accepts selector and also DOM element. 
Tested, drag no more fail.


mquandalle:
Thanks!

Issue #51, opened 2015-01-02
Welcome board not working


mquandalle:
When a new user is created, we want to create a sample board. This feature is broken due to the `userId` parameter of the `after.insert` hook being `null` ; and we consequently create a `BoardMembers` document with a null `userId`.

mquandalle:
[Here](https://github.com/yasaricli/metrello/commit/df1642494c0ac427bdb133fc8b71298701f8fd1b#diff-13b3b47c8335836d2a2e0da24cc7256eL73) is a hacky fix.

mquandalle:
Fixed in 5d43a5f61d777542eddcfdbca1765048268981fa

pauricthelodger:
I just set up an account and I'm afraid that there was no welcome board. Had this commit been released yet?

mquandalle:
Yes, but a commit (8d2f687464660b35103b14856a8e33db2273b55d) broke this feature again.

Anyway I plan to refactor this welcome board with the following goals in mind:

1. Have a board that explains the project
2. Have it available in the user language
3. Displayed in the home page (if the user is not connected)
4. Copied into the user board list when a user create a new account

pauricthelodger:
Good plan! May I also suggest adding an about link in the header member menu? Even a link to this repo as it was rather empty for myself as a new user. Not that I minded, I just searched here. Would you prefer I open a separate issue for that, if you think it's something to think about?

mquandalle:
The home definitely need a lot of work (which in a way is not too bad, because it slow down adoption while we're implementing new features like crazy!). Yes, you can open a new issue about it :-)

PR #411, opened 2015-11-27
add: set user preferred lang and allow select lang for user form

floatinghotpot:
Currently, before login, the user form always display English. This patch set user form language according to browser language, and allow user to select from the language list that WeKan supports.

### DIFF
--- a/client/components/main/layouts.jade
+++ b/client/components/main/layouts.jade
@@ -13,6 +13,13 @@ template(name="userFormsLayout")
     h1.at-form-landing-logo
       img(src="{{pathFor '/wekan-logo.png'}}" alt="Wekan")
     +Template.dynamic(template=content)
+    div.at-form-lang
+      select.select-lang.js-userform-set-language
+        each languages
+          if isCurrentLanguage
+            option(value="{{tag}}" selected="selected") {{name}}
+          else
+            option(value="{{tag}}") {{name}}
 
 template(name="defaultLayout")
   +header
--- a/client/components/main/layouts.js
+++ b/client/components/main/layouts.js
@@ -2,8 +2,41 @@ Meteor.subscribe('boards');
 
 BlazeLayout.setRoot('body');
 
+const i18nTagToT9n = (i18nTag) => {
+  // t9n/i18n tags are same now, see: https://github.com/softwarerero/meteor-accounts-t9n/pull/129
+  // but we keep this conversion function here, to be aware that that they are different system.
+  return i18nTag;
+};
+
 Template.userFormsLayout.onRendered(() => {
+  const i18nTag = navigator.language;
+  if (i18nTag) {
+    T9n.setLanguage(i18nTagToT9n(i18nTag));
+  }
   EscapeActions.executeAll();
+});
+
+Template.userFormsLayout.helpers({
+  languages() {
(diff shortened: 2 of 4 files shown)


floatinghotpot:
The i18n to T9n conversion will not be necessary any more in future. 

My PR to T9n has been merged, to add codes compatible with IETF tags. See: https://github.com/softwarerero/meteor-accounts-t9n/pull/129

But not sure when it will be published to atmosphere. See: https://atmospherejs.com/softwarerero/accounts-t9n

client/components/main/layouts.js
```diff
@@ -2,10 +2,52 @@ Meteor.subscribe('boards');
...
+  'es-ES': 'es_ES',
+  'no-NB': 'no_NB',
+  'pt-BR': 'pt',
+  'pt-PT': 'pt_PT',
+  'zh-CN': 'zh_cn',
+  'zh-HK': 'zh_hk',
+  'zh-TW': 'zh_tw',
+};
+
+const i18nTagToT9n = (i18nTag) => {
```
mquandalle: Since https://github.com/wekan/wekan/pull/421 was released, we can now remove this.

client/components/main/layouts.jade
```diff
@@ -13,6 +13,13 @@ template(name="userFormsLayout")
     h1.at-form-landing-logo
       img(src="{{pathFor '/wekan-logo.png'}}" alt="Wekan")
     +Template.dynamic(template=content)
+    div.at-form-lang
+      select.select-lang.js-userform-set-language
```
mquandalle: I would rather place this `<select>` in the registration form, no?

floatinghotpot: I am okay with your proposal, please go on.

floatinghotpot: BTW, with your proposal, will this section appear when user form in any state (sign-in, registration, pwd-forgot...)?

mquandalle: No just the registration form.

floatinghotpot: If put it into registration form, maybe it should also apply to enroll form when user set initial password ... 

I am not very sure whether it's needed or not, as the default language of user form is already user navigator language. 

I just thought of providing a way for user to change the language of the user form in any state, in case it's not preferred.

mquandalle: Actually the reason why I would like the user to explicitly check the language before registration is because I plan to revive the “Welcome Board” feature. Basically it’s a board filled with some data (like Wekan explanations) so that that the first page that the user see after registration is not a completely empty one and he can start playing with some data.

This board will need to be created (or rather copied from a template) just after the user has registered, and it will not be possible to change the language of the content (card titles, descriptions, comments) after its instantiation.

And so yes, if we go on that path we should also add this field in the enroll by email form.

floatinghotpot: I now understand what you mean. 

A welcome board might be helpful to new user, maybe a simple start board with 4 lists (PLAN / TODO / DOING / DONE) is already enough. If too heavy or too many i18n sample data, maybe it will become a burden.

mquandalle: Actually I didn’t try your implementation locally. I think the way you made it is elegant, so let’s keep it like that. We may reconsider based on the rationale above when implementing #51.

Issue #406, opened 2015-11-19
Discuss: architecture for data import (plugins?)


xavierpriour:
This is a thread to discuss how to implement multiple data import (#113, #152, #395) in the most easy and extensible way, and as a potential starting point for a full plugin architecture (#331, #332).

Here's how I envision it:
- specific system imports should not be part of the core of Wekan
- on the other hand, Wekan core should provide mechanisms for new imports to declare themselves and appear in the appropriate UI - I'm thinking a singleton "plugin registry" of some sort.
- each system import should be a different package ('wekan-import-trello'), that way, adding a new system would not depend on a new release for Wekan
- adding the package would automagically add the option to the board

I put together a quick'n dirty proof of concept on my forked branch, if anybody wants to have a look: https://github.com/xavierpriour/wekan/blob/feature/import-plugin/models/import.js
lines 135+ would be in the import-trello package, while 1-136 would stay in core.

Any feedback?

mquandalle:
Hi Xavier, I actually answered in #331, feel free to continue the discussion there.

Issue #395, opened 2015-11-10
Add Feature: Import/Export CSV/Excel/PDF


jadeqwang:
It would be awesome to be able to import CSV's into a Wekan board, as many other apps have the option of exporting task data into a spreadsheet format (including forms and Gmail-plugin-CRMs). (I'm trying to migrate off of Streak (which is buggy and hard to use) to tracking my tasks on Wekan.

I'm going to suggest a format that's optimized for a CRM-type task, but am open to other suggestions (esp. if this sounds crazy) for Mapping CSV to Kanban board. 

MVP:
- Every row becomes one card (except the header row). 
- The first column is the name of the card. 
- The 2nd column has the "list" that the card should be in (e.g., the "stage" of the customer). 
- The titles of the lists are populated from the limited set of options in the 2nd column.
- Data from all other columns are put into the Description of the card, separated by new lines.

Super fancy:
- Show the user a list of header rows and let the user choose which column is the name of the card, which column is the list, and which columns to ignore completely

floatinghotpot:
@jadeqwang I guess you will be interested in my PR. https://github.com/wekan/wekan/pull/397

floatinghotpot:
This feature has been implemented in PR #413

xet7:
Moved to here from #2926 

From @steep-nb

Exporting Lists, Swimlanes, Boards or Cards to a pdf or Excel sheet would be perfect!
We need a whole List in a sheet which we can show someone, who doesn't use wekan.

From @xet7 

Related code is here:
1. Old export to CSV PR https://github.com/wekan/wekan/pull/413
2. Forked Wekan repo that seem to have Export to Excel, but it has some Chinese characters in code https://github.com/gameendman/wekan and https://snapcraft.io/wekan-ag . Could @gameendman contribute those changes to official Wekan, with Chinese converted to English? I have not checked what all that Excel related code does.
3. From trapa, attached PHP code that converts Wekan Export JSON to CSV summary reports.

[WekanExportJSONtoCSVwithPHP.zip](https://github.com/wekan/wekan/files/4276711/WekanExportJSONtoCSVwithPHP.zip)

It would be nice if some Wekan contributor would spend time to clean up and combine code from these and send pull request to integrate these directly to Wekan.

brymut:
More than happy to have a go at this over the next week and submit a PR

xet7:
@brymut 

Yes you are welcome to do so. I will first wait for you to try add PR.

[...]

PR #413, opened 2015-11-28
add: export/import Excel TSV/CSV data

floatinghotpot:
1. Fix: currently user can add, remove, changePermission for self, which result that if admin changes self to normal user, no one can change the board anymore.
2. extend card data model with manHour, dueDate, startDate, finishDate, which is very basic for project time managing needs.
3. export cards to Excel TSV.
4. import cards from Excel CSV/TSV, implementing based on common import interface.

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/413)


floatinghotpot:
Screenshot:

menu

The content can be coped/pasted between Excel, example:
screen1

mquandalle:
As @xavierpriour discussed in #406 I believe that this kind of import feature should be distributed as Wekan plugin rather than being a core feature (maybe CSV import could still be a core feature, but people have also been requesting less standard imports like in #152).

However since this plugin infrastructure currently doesn’t exist, I will be happy to merge this PR. Then when we get plugins, we might refactor some Wekan core features to put them in plugins. 

Thanks for this!

client/components/import/import.js
```diff
@@ -135,6 +143,40 @@ ImportPopup.extendComponent({
...
+  },
+
+}).register('listImportCardsTsvPopup');
+
+ImportPopup.extendComponent({
+  getAdditionalData() {
+    const listId = this.currentData()._id;
+    const selector = `#js-list-${this.currentData()._id} .js-minicard:first`;
+    const firstCardDom = $(selector).get(0);
+    const sortIndex = Utils.calculateIndex(null, firstCardDom).base;
```
mquandalle: Could you just add a `XXX comment` saying that we should find a way to avoid this crazy DOM reading gymnastic at some point? (This is currently the easiest way to retreive the index so keep the code untouched, I would just like a reminder when we refactor client side state stores).

models/boards.js
```diff
@@ -170,6 +174,7 @@ Boards.mutations({
   },
 
   addMember(memberId) {
+    if (memberId === Meteor.userId()) return null;
```
mquandalle: The security checks aren’t actually implemented in these `mutations` methods but using the allow/deny rules (because it is possible to overpass mutations checks by simply executing the minimongo query on the browser console). I know that this is a bit confusing and I’m considering an evolution of the whole mutation model (see also https://forums.meteor.com/t/new-package-mdg-method-meteor-methods-with-better-scoping-argument-checking-and-good-defaults/13833)

models/importCsvData.js
```diff
@@ -0,0 +1,265 @@
+// we can also copy & paste excel data for batch task
```
mquandalle: Could you write an abstract comment about how this importer work generally speaking?

models/importCsvData.js
```diff
@@ -0,0 +1,265 @@
...
+      if(card.members && card.members.length > maxMembers) maxMembers = card.members.length;
+      if(card.labelIds && card.labelIds.length > maxLabels) maxLabels = card.labelIds.length;
+    });
+    const hmembers = _.range(maxMembers).map(() => 'member');
+    const hlabels = _.range(maxLabels).map(() => 'label');
+    const headers = ['title',
+                     'description',
+                     'status',
+                     'owner',
+                     ].concat(
```
mquandalle: Why do you `concat`?

floatinghotpot: As number of members and labels may be different for cards, we need use max of them as columns, while set those with less members or labels with empty cells.

floatinghotpot:
@mquandalle I've improved the PR with more comments.

models/importCsvData.js
mquandalle: I think using ES6 splash would be easier to understand:

```es6
const headers = ['title', 'description', 'status', 'owner', ...hMembers,
                 ...hLabels, 'manHour', 'dueDate', 'startDate', 'finishDate', 
                 'createdAt', 'updatedAt'];

File: models/importCsvData.js [hunk 5]
```diff
@@ -0,0 +1,291 @@
...
+                     'updatedAt']);
+    const dataGrid = [];
+    dataGrid.push( headers.join(separator) );
+
+    cards.forEach((card) => {
+      // if not many members, fill the rest with empty cells
+      const usernames = (card.members) ? card.members.map((memberId) => mapuser[memberId]) : [];
+      if (usernames.length < maxMembers) {
+        for (let i=usernames.length; i<maxMembers; i++) usernames.push('');
+      }
```
mquandalle:
```es6
_.times(maxMembers - usernames.length, () => usernames.push(''));
```

floatinghotpot:
@mquandalle I've updated the code according to your comments, and also merged the 3 commits into 1.

mquandalle:
I actually plan to remove the “move all cards in this list” and “archive all cards in this list” action from the list menu if favor of the multi-selection feature. I think it would make sense to do the same for the CSV export feature. Basically we would have a link in the sidebar below “Archive selection”, like “Export to CSV”. It would add a single click to achieve the effect that is currently implemented (ie, exporting a single list by clicking “Select all cards in this list” and then “Export to CSV”) but provide much more flexibility to customize the exact set of cards that the users want to export. Thoughts?

floatinghotpot:
It does make more sense. Will see how to do it in this way.

floatinghotpot:
@mquandalle I've refactored the PR based on RC4, now export selection.

mquandalle:
Again, sorry for not having merge that already @floatinghotpot. I would also like to merge this feature as a build in plugin, but I need to have a minimal version of the plugin system first. I will work on that as soon as Meteor release a working version of the ES6 `import`/`export` and will help you in the adaptation process.

I think this is actually having actual feature not yet in core but already implemented will actually help designing the requirements and APIs of the module system.

floatinghotpot:
No problem. I implemented the features to meet current needs from our project team, so in my fork, I cannot wait. But to main branch, I can re-write the PR when your plugin system is ready.

brunombo:
hello team

Please when you hope to merge this very usefull feature to master branch ?

Thanks again for your help

Bruno

HLFH:
@brunombo Go there. https://github.com/wefork/wekan/issues and submit an issue related to this PR. Wekan is no longer active. See https://github.com/wefork/wekan/wiki/FAQ.

jLouzado:
I'm just trying to understand the code in this PR @floatinghotpot, it seems to me that the `manHour`, `startDate`, `dueDate` etc are fields that are being used within your internal team's deployment of Wekan right? As in that's why you're supporting it in your import/export feature; not that you were trying to implement import/export _and_ a due-date feature at the same time. Please correct me if I'm wrong, thank you.

fmonthel:
This feature is awesome to use and print data (export)

xet7:
This pull request has been made a long time ago, it would need some work to add all the new tables, columns etc from MongoDB database, so that this would Import/Export would  correctly. Contributions welcome.

xet7:
Also see wiki page: [Excel and VBA](https://github.com/wekan/wekan/wiki/Excel-and-VBA)

anikasingh201611:
Hi I am interested in this feature. I don't understand from the above conversations if it was implemented. 
Can someone please help me and tell me if it is available ?

xet7:
@anikasingh201611 

This is not available, because nobody has cleaned up this or added latest database fields to export/import or submitted fixed PR. Issue for this is #395

twupp:
Seems that CSV-import is available but I cannot find and docs about the imported fields and the needed columns in CSV-files. Can anyone help with documentation or claryfication?

rdgarlo:
Can you tell me how to find the "import" option? I cannot find it. In answer to your question, I recommend exporting an existing board to get the structure of the CSV headers, and then populating your new CSV file according to that same structure.

xet7:
@rdgarlo 

Thanks! I figured it out and added to wiki: https://github.com/wekan/wekan/wiki/CSV

Issue #417, opened 2015-12-02
Import Trello: exception when importing a board


xavierpriour:
We get an exception when trying to import some Trello boards - for example:
https://trello.com/b/hjBDflxp/meteor-roadmap

This is caused by a missing board creation date.

PR #418, opened 2015-12-02
Import Trello: set creation date for older Trello elements

xavierpriour:
Older Trello elements do not report creation activity. As a result we cannot compute their actual creation date. Currently this crashes the import (issue #417).

This PR changes the import to use the current timestamp in that case - actually, we now log all import actions with the same "current" timestamp, this helps to retrieve what actions belong to what import. 

We also set a few more elements (if the importer properly mapped Trello members to Wekan users):

- comments are now assigned to their author (it used to be the importer)
- members that were admin of the original board are now admin of the imported board
- cards userId is now set to the original user that created them


models/import.js
```diff
@@ -292,31 +342,35 @@ class TrelloCreator {
...
-        userId: Meteor.userId(),
+        // We attribute the import to current user,
+        // not the creator of the original object
+        userId: this._user(),
       });
     });
   }
 
+  getAdmin(trelloMemberType) {
+    if (trelloMemberType === 'admin') {
```
mquandalle:
```es6
return trelloMemberType === 'admin';
```

models/import.js
```diff
@@ -25,6 +34,39 @@ class TrelloCreator {
...
+  /**
+   * If dateString is provided,
+   * return the Date it represents.
+   * If not, will return the date when it was first called.
+   * This is useful for us, as we want all import operations to
+   * have the exact same date for easier later retrieval.
+   *
+   * @param {String} dateString a properly formatted Date
+   */
+  _now(dateString) {
```
mquandalle: :+1:

xavierpriour:
done @mquandalle

mquandalle:
Thank you very much for your reactivity Xavier, I’ll be able demo this feature tomorrow :-)

PR #421, opened 2015-12-03
i18n-ize meteor email templates

floatinghotpot:


config/accounts.js
```diff
@@ -51,4 +51,13 @@ if (Meteor.isServer) {
   if (process.env.MAIL_FROM) {
     Accounts.emailTemplates.from = process.env.MAIL_FROM;
   }
+
+  ['resetPassword-subject', 'resetPassword-text', 'verifyEmail-subject', 'verifyEmail-text', 'enrollAccount-subject', 'enrollAccount-text'].forEach((str) => {
+    const words = str.split('-');
+    Accounts.emailTemplates[words[0]][words[1]] = (user, url) => {
+      const lang = (user.profile && user.profile.language) ? user.profile.language : 'en';
+      const name = (user.profile && user.profile.name) ? user.profile.name : user.username;
```
mquandalle: Could hide the implementation details of the retrieval by creating the `getLanguage()` and `getName()` method on the user object (ie to the helper list defined in `models/users`)?

config/accounts.js
```diff
@@ -51,4 +51,13 @@ if (Meteor.isServer) {
...
+      return TAPi18n.__(`email-${str}`, { user: name, url }, lang);
```
mquandalle: Doesn’t this require the language code conversion that you implemented in #411?

mquandalle:
:+1: This is an excellent idea, thank you :)

config/accounts.js
floatinghotpot: That code conversion is only for T9n, to set language for meteor account UI resources. Here is for the i18n resources inside WeKan so conversion is not needed.

mquandalle: Perfect!

PR #422, opened 2015-12-03
add: invite new user to board via email

floatinghotpot:
Currently, we can only add user existing in system to our board. 

Here introduce a way to invite new user to join the board by sending invitation email. The new user will be created and added to the board, but no password is set. The new user has to click the link in enrollment email to set initial password.

This feature will enhance social interaction.


mquandalle:
Hi Raymond,

This is a feature I planned to implement for a long time, so I’m super happy with this PR :-) I haven’t pulled the code locally yet, but just from reading the source I think this feature misses a notion of “verified user” which would be required to show a special notice on invited members details popover like “This member has been invited but haven’t joined yet”. What do you think?

models/users.js
```diff
@@ -107,6 +133,75 @@ Meteor.methods({
...
+        }
+        Accounts.sendEnrollmentEmail(newUserId);
+        user = Users.findOne(newUserId);
+      }
+
+      board.addMember(user._id);
+      user.addInvite(boardId);
+
+      try {
+        const rootUrl = process.env.ROOT_URL || '';
```
mquandalle: I think it better to read it from `Meteor.absoluteUrl.defaultOptions.rootUrl` than from the `process.env`

floatinghotpot:
@mquandalle Thanks for confirmation. This feature was in my mind for weeks, I could not sleep well if it's not implemented. :-)

Currently implementation is some kind of compromise. In my mind, the perfect process is: inviter add the user to pre-members list, when invitee click "accept", then moved from pre-members to members. And, member should have the right to say reject invitation, or quit from a board. But unfortunately, only admin have the privilege update board. Any idea to do it?

You may notice that, the board is added to invitee's list of invitedBoards. I can polish the code to show the message as you comment (and also a red dot on the avatar?), and also display the newly invited boards separately at bottom of invitee's "all boards" page, so invitee can decide to accept or deny.

models/users.js
floatinghotpot: Ok. Will update in next move, after everything is clear in discussion.

mquandalle:
Hum I didn’t get that you had implemented this pre-member list from my too short look at the code. It sounds good to me, I’ll pull locally and provide a complete review.

floatinghotpot:
The preMembers is not in current compromised code. 

I tried to extend the board with new property preMembers{ type: [String], optional: true }, but removed it when I found out that only admin can update board, invitee will failed to move self from preMembers to members. 

Is possible to allow non-admin user to: (1) move self from pre-members to members. (2) remove self from members?

mquandalle:
Ok I see the problem. Meteor always has this non-clear story about how we should update the data, I created my own layer (the `Collection.mutations` thing) on top of minimongo calls, but I plan to refactor that once the new Meteor guide gives its recommendations. For now I would say that we should use simples Meteor method for the two use cases that you outlined. You can define them in the `model/board.js` and be sure to enforce security checks because allow and deny rules aren’t enforced for Meteor methods.

floatinghotpot:
@mquandalle the feature "invite user via email, invited user can accept or decline" is fully implemented, please review.

client/components/sidebar/sidebar.js
```diff
@@ -194,25 +222,77 @@ function draggableMembersLabelsWidgets() {
...
-});
 
-Template.addMemberPopup.events({
-  'click .js-select-member'() {
-    const userId = this._id;
-    const currentBoard = Boards.findOne(Session.get('currentBoard'));
-    currentBoard.addMember(userId);
-    Popup.close();
+  isValidEmail(email) {
+    const re = /^([\w-]+(?:\.[\w-]+)*)@((?:[\w-]+\.)*\w[\w-]{0,66})\.([a-z]{2,6}(?:\.[a-z]{2})?)$/i;
```
mquandalle: I would use `SimpleSchema.RegEx.Email` (see https://github.com/aldeed/meteor-simple-schema#regex)

client/components/boards/boardHeader.jade
```diff
@@ -117,6 +117,9 @@ template(name="boardChangeTitlePopup")
     label
       | {{_ 'title'}}
       input.js-board-name(type="text" value=title autofocus)
+    label
+      | {{_ 'description'}}
+      textarea.js-board-desc= description
```
mquandalle: Hey that’s another feature! I won’t ask you to put it in a separate commit because I understand it would be painful, but could you just mention it in the commit in case (just in case we have some issue with it latter and we want to explore the history).

mquandalle:
Tested locally, and the invitation process was totally smooth, this is very good :) Please just consider my inline comments and I’ll merge as is.

floatinghotpot:
@mquandalle According to your line comments, I've split the commit into 2 commits by feature, and SimpleSchema.RegEx.Email is used. 

As to red dot on avatar, I prefer it for its simplicity. it depends on how to understand the presence or availability. now I change css tag "invited" to "pending", is it now making more sense?

mquandalle:
Thank you very much for your fast responses! This is now merged (and I plan to release v0.10-RC tomorrow with it).

PR #426, opened 2015-12-08
bugfix: only care active members, also optimize some code

floatinghotpot:
My previous implementation of removeMember() is changing userId to "x". The original purpose is to hide the boards from the user once the membership removed. Now I find it brings severe side-effects: when user removed from board, related cards and activities will cause null.username error. This error has to be corrected, and change it back to keep userId unchanged.

Now make changes to the publication, should be the right approach. Also change wherever should see active members only. And some code optimization.

### DIFF
--- a/client/components/cards/cardDetails.jade
+++ b/client/components/cards/cardDetails.jade
@@ -94,7 +94,7 @@ template(name="moveCardPopup")
 
 template(name="cardMembersPopup")
   ul.pop-over-list.js-card-member-list
-    each board.members
+    each board.activeMembers
       li.item(class="{{#if isCardMember}}active{{/if}}")
         a.name.js-select-member(href="#")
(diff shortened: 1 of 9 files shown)


mquandalle:
Hum I didn’t realize I merged the 'x' index because I was doubtful about it. I’m glad you improved the implementation in this PR. Any reason you re-used the `isActive` flag instead of creating a `isInvited`? How do we distinguish between invited members (that are not yet board members) and former members (that are no longer board members)?

floatinghotpot:
The invited member isActive:true, he is same as other normal members, so that he can access the board. 

We recognize the invited members by checking whether the board Id is in his invitedBoards list or not.

mquandalle:
Ok thank you for the clarification.

PR #430, opened 2015-12-08
Full screen board import

mquandalle:
Work in progress, see https://github.com/wekan/wekan/issues/113#issuecomment-162945488 for rationale.

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/430)


mquandalle:
@xavierpriour, I plan to change the import flow to make it linear:

step 1: chose the import format
step 2: upload the data
step 3: member mapping
then loading bar, or spinner

I’ve just realized that this changes the flow currently implemented when you go back and forth between step 2 and 3, but I think this is a pretty reasonable model.

I imagine the UI should have a horizontal component showing the process progress. Clicking on a previous step should bring the user back to it. On the implementation side, all the import data would be saved under the `import` root component and not in the individual step components, which is a easier model to reason about. I imagine in the future, third-parties importers will be able to customize the steps, and add more if needed.

I’ve also been blindly refactoring the client-side import code, but I’m not sure if I do things correctly :-/

Also, as the actual import “business logic” is located in the `models/` directory, so it is distributed in both the client and the server, which means that the `importTrelloBoard` method runs on both the environments. Is that on purpose? It seems unnecessary heavy to me to run the latency compensation method on the client, especially since we are waiting for server completion with a spinner anyway.

xavierpriour:
@mquandalle, a multi-step wizard definitely makes sense for such a process.

Import was latency-compensated because why not? I could see this useful in a bad network/offline scenario, so not sure it's really worth removing it.

mquandalle:
I need to look closer at the client side method performance for latency compensation, but I was under the impression that it freezes the client for a few seconds if you import large boards. Otherwise, yes, why not.

client/components/import/import.js
```diff
@@ -76,99 +53,70 @@ const ImportPopup = BlazeComponent.extendComponent({
...
+    return this.parentComponent().membersToMap.get();
   },
+
   _refreshMembers(listOfMembers) {
-    Session.set('import.membersToMap', listOfMembers);
+    return this.parentComponent().membersToMap.set(listOfMembers);
   },
+
   /**
    * Will look into the list of members to import for the specified memberId,
```
mquandalle: It seems to me that some complexity related to this component, and this function in particular, are due to the Popup environment restrictions. Aside from the from all the helpers to read and write the state, what is the simple data structure we need to use here?

xavierpriour: The data you need:

1. the list of members from the import file
2. the list of members from Wekan
3. a mapping between the 2

The current implementation used:

1. the members field from the Trello import, which is not perfect if you want to re-use the mechanism for other imports
2. the Users collection - which causes a bug: the auto-mapping looks into the currently loaded list of users, which is only the current user if you haven't opened a board yet
3. for mapping, we added the Wekan User object to the import member directly. Pretty hacky...

mquandalle:
@xavierpriour this should be completed as a basic UI change from popup to full screen, it also fixes #504. Could you take a quick look?

client/components/import/import.js
mquandalle: Point 2 is fixed. I think I should also tackle point 3 as we only need to reference the user id, not the whole document.

mquandalle: Point 3 is also fixed now

savolai:
Hi. I just created a sandstorm grain for Wekan but can't find the import functionality anywhere. Ideas? Thanks. ^_^

xet7:
@savolai 

It's not yet on Sandstorm, see #799

savolai:
Thanks.

PR #435, opened 2015-12-13
Export a board to JSON

### LINKED ISSUE #396: Support for export to JSON
to import into another instance with the import from json function

xavierpriour:
Exports a board to JSON (specific Wekan format).
On a given board, go to Board Menu (the "gear" icon) then select "Export board".
It downloads as a json file under filename "wekan-export-board-<boardId>.json"

Fixes #396.


.meteor/packages
```diff
@@ -73,3 +73,4 @@ perak:markdown
 seriousm:emoji-continued
 templates:tabs
 verron:autosize
+pfafman:filesaver
```
mquandalle: As discussed in https://github.com/wekan/wekan/issues/396#issuecomment-164301574 I would favor a server-side export generation over a client one.

client/components/boards/boardHeader.js
```diff
@@ -1,3 +1,4 @@
+/* global saveAs */
```
mquandalle: Soon we will have ES6 modules, yay!

models/export.js
```diff
@@ -0,0 +1,58 @@
...
+  },
+});
+
+class Exporter {
+  constructor(boardId) {
+    this._boardId = boardId;
+  }
+
+  build() {
+    const byBoard = {boardId: this._boardId};
```
mquandalle: I think the export format miss a `formatVersion` number that we can increment whenever we change the format.

xavierpriour: added a "_format" field, set to "wekan-board-1.0.0" to be as self-descriptive as possible

.meteor/packages
xavierpriour: Done!

client/components/boards/boardHeader.js
xavierpriour: Gone with client-side saving anyway :smile:

xavierpriour:
Rewrote to be server-based, authentication through a login token passed as part of the request URL (the future REST API should pass it as a request header, it's cleaner).
I also fixed the fact that my initial version did not export attachments.

server/lib/utils.js
```diff
@@ -5,3 +5,24 @@ allowIsBoardAdmin = function(userId, board) {
 allowIsBoardMember = function(userId, board) {
   return board && board.hasMember(userId);
 };
+
+// todo XXX not really server-specific,
+// so move it to a common (client+server) lib?
```
mquandalle: It’s reasonable for now :)

models/export.js
mquandalle: :+1:

models/export.js
```diff
@@ -0,0 +1,85 @@
+/* global JsonRoutes */
+if(Meteor.isServer) {
+  JsonRoutes.add('get', '/api/b/:boardId/:userId/:loginToken', function (req, res) {
+    const { userId, loginToken, boardId } = req.params;
```
mquandalle: Maybe add the link to the https://blog.kayla.com.au/server-side-route-authentication-in-meteor/ article to explain to future readers what is happening here

server/lib/utils.js
xavierpriour: By the way, any reason why the methods in this file (allowIsBoardMember and allowIsBoardAdmin) are not defined as Boards.helpers?

models/export.js
xavierpriour: Done!

server/lib/utils.js
mquandalle: IIRC, it’s just historical chronology, we defined these methods before using collection-helpers.

client/components/boards/boardHeader.js
```diff
@@ -15,6 +15,19 @@ Template.boardMenuPopup.events({
   }),
 });
 
+Template.boardMenuPopup.helpers({
+  exportUrl() {
+    const boardId = Session.get('currentBoard');
+    const userId = Meteor.userId();
+    const loginToken = Accounts._storedLoginToken();
+    return Meteor.absoluteUrl(`api/b/${boardId}/${userId}/${loginToken}`);
```
mquandalle: Since you explicitly designed the export as the beginning point of a REST-like API, which is a good idea, I think it is worth considering the following point about the URI scheme:

* I don’t see the point of having to transmit both the `loginToken` and the `userId`, the former should be enough. This is how HTTP authentication usually work (either with a cookie on a web page, or with an explicit GET parameter on APIs) and I feel we should stick with this convention
* If we look at the [Trello API](https://developers.trello.com/advanced-reference/board) only the “mandatory method parameters” are defined in the first part of URI, the rest is defined using GET parameters. This also assume that we have a method name.

With these elements the URL would be:

```js
`api/b/${boardId}/export?tokent=${loginToken}`
```

We might also want to follow Trello and use the explicit `boards` namespace instead of its `b` shortcut:

```js
`api/boards/${boardId}/export?tokent=${loginToken}`
```

I wonder if we should also explicit the return format in the URI using something like:

```js
`api/boards/${boardId}/export.json?tokent=${loginToken}`
```

I guess that would be useful if someone wants to implement some other serialization format at some point (XML lol). But this information could also fit elsewhere like in a GET parameter (`&format=json`) and since Trello doesn’t use it we may as well follow them to facilitate plugin interpolation between Trello and Wekan.

What do you think Xavier?

models/export.js
```diff
@@ -0,0 +1,95 @@
+/* global JsonRoutes */
+if(Meteor.isServer) {
+  // todo XXX once we have a real API in place, move that route there
```
mquandalle: Another `XXX TODO` would be to share the route definition between the client and the server so that we could use something like `ApiRoutes.path('boards/export', boardId)` on the client instead of copy/pasting the route path manually between the client and the server.

client/components/boards/boardHeader.js
xavierpriour: - using only the token, and as a parameter, is great. That way, we can use the same route for api access, just with code looking for token in parameter then in header. I like that :heart: 
- I'm not a fan of using an "export" method name, it seems non-RESTy, we just want a given board: `api/boards/${boardId}?authToken=${loginToken}`
- the format parameter seems overkill for the time being. I suggest we add it later if needed (and keep current JSON as the default). But when we add it, we might use it to specify versioning as well (like `?format=wekan-board-1.0.0.json`)

models/export.js
xavierpriour: ok

client/components/boards/boardHeader.js
mquandalle: Ok with the 3 points :+1:

singpolyma:
If we could get this into the old newui.libreboard.com instance, then I can finally migrate my board off that creaky thing :)

mquandalle:
Well, it’s not yet possible to import the JSON export just yet. It should be fairly easy to implement though, we’ll have it for Wekan 0.11 due on February 4.

Issue #437, opened 2015-12-14
Support Mobile Browser


floatinghotpot:
If WeKan works in mobile device (at least mobile browser as the first step), it will be much more convenient. Maybe we can consider raise the priority to support mobile device. I'd like to open a tracker to discuss how to implement this feature.

Here are the screen size of mobile browser:
iphone, width 320px
iphone6 plus, width 414px
and more:
http://www.paintcodeapp.com/news/ultimate-guide-to-iphone-resolutions

My thought is when detecting mobile device, adjust all the topmost UI to fit the full screen. The related UIs are:
1. pop up (current width is fixed 300px).
2. carde details (current width is fixed 470px).
3. user login form (current width is fixed 325px).
4. sidebar (current width is fixed 248px).
5. lists (current width is fixed 270px).

I will try prototyping the above ideas, and any comments are welcome.

nata-g:
@floatinghotpot Question: Does this sizing dimensions apply only to Iphone devices or android phones as well?

floatinghotpot:
I suppose the approach should apply to all small screen, even a PC browser window narrower than a fixed width, say, 420px.

floatinghotpot:
It's quite smooth for the prototyping, and I even think it's good enough as a new feature. 

@mquandalle Please help to review. PR is created, #442. 

Some screenshots, FYI:

Board List & Board View:
![img_3162](https://cloud.githubusercontent.com/assets/2339512/11846408/763a409e-a453-11e5-8999-d716dc1fc6ff.jpg)  ---  ![img_3163](https://cloud.githubusercontent.com/assets/2339512/11846419/88838580-a453-11e5-9c5e-7b60a5f6e234.jpg)

Card View, Sidebar:
![img_3164](https://cloud.githubusercontent.com/assets/2339512/11846491/db2c5e42-a453-11e5-838d-84b19b7cad7d.jpg) ---  ![img_3167](https://cloud.githubusercontent.com/assets/2339512/11846517/ef07084a-a453-11e5-9052-03ad0b8c18c8.jpg)

Pop-over, & Form:
![img_3165](https://cloud.githubusercontent.com/assets/2339512/11846536/079dbc5a-a454-11e5-84e1-2e3586581d6a.jpg) --- ![img_3166](https://cloud.githubusercontent.com/assets/2339512/11846546/0e6b8ab2-a454-11e5-98a8-f36ead93828d.jpg)

mquandalle:
[...]

PR #442, opened 2015-12-16
add: support mobile web, adapt to small browser window

floatinghotpot:
PR for feature mobile web.

See discussion and screenshots: #437

### DIFF
--- a/client/components/boards/boardsList.jade
+++ b/client/components/boards/boardsList.jade
@@ -27,7 +27,6 @@ template(name="boardList")
 
 template(name="boardListHeaderBar")
   h1 {{_ 'my-boards'}}
-
   .board-header-btns.right
     a.board-header-btn.js-open-archived-board
       i.fa.fa-archive
--- a/client/components/boards/boardsList.styl
+++ b/client/components/boards/boardsList.styl
@@ -1,7 +1,7 @@
 $spaceBetweenTiles = 16px
 
 .board-list
-  margin: $spaceBetweenTiles ($spaceBetweenTiles/-2) 0
+  margin: 0 ($spaceBetweenTiles/2)
 
   li
     float: left
@@ -24,8 +24,8 @@ $spaceBetweenTiles = 16px
     display: block
     font-weight: 700
     min-height: 18px
-    padding: 8px 12px 8px 12px
-    margin: 0 ($spaceBetweenTiles/2) $spaceBetweenTiles
+    padding: 8px
+    margin: ($spaceBetweenTiles/2)
     position: relative
     text-decoration: none
 
@@ -128,3 +128,22 @@ $spaceBetweenTiles = 16px
         font-size: 25px
         color: white
 
+@media screen and (max-width: 800px)
+  .board-list
+    height: 100%
+    overflow: scroll
+
+    li
+      width: 33.3%
+
+    .board-list-item
+      overflow: hidden
+
+    .board-list-item-sub-name
+      position: relative
+      top: -100px
+      left: -100px
+
+@media screen and (max-width: 360px)
+    li
+      width: 50%
--- a/client/components/main/header.styl
+++ b/client/components/main/header.styl
@@ -5,75 +5,8 @@
   transition: background-color 0.4s
   background: #2980B9
 
-  #header-quick-access
-    background-color: rgba(0, 0, 0, 0.2)
-    padding: 0px 10px
-    height: 28px
-    font-size: 12px
-    display: flex
-
-    #header-user-bar,
-    #header-new-board-icon,
-    ul li
-      color: darken(white, 17%)
-
-      .fa
-        color: inherit
-
-      a:hover, a.is-active
-        color: white
-
-    ul
-      transition: opacity 0.2s
-      margin: 4px 0 0 5px
-      overflow: hidden
-
-      li
-        display: block
-        float: left
-        width: auto
-        color: darken(white, 15%)
-        padding: 2px 5px 0
-
-        &.current
-          color: darken(white, 5%)
-
-        &:first-child .fa-home
-          margin-right: 5px
-
-        a.js-create-board
-          margin-left: 5px
-
-    #header-user-bar,
-    #header-new-board-icon
-      flex-shrink: 0
-
-    #header-user-bar
-      margin: 2px 0
-
-      .header-user-bar-avatar
-        float: left
-
-        .member
-          width: 24px
-          height: @width
-          margin: 0
-          margin-top: 1px
-
-      .header-user-bar-name
-        margin: 4px 8px 0 0
-        float: left
-
-        i.fa-chevron-down
-          margin-right: 4px
-
-    #header-new-board-icon
-      flex-grow: 1
-      margin: 6px 5px 0
-      width: 12px
-
   #header-main-bar
-    height: 28px * 1.618034 - 6px
+    height: 40px
     padding: 7px 10px 0
 
     h1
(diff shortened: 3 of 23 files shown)


client/lib/utils.js
```diff
@@ -22,6 +22,11 @@ Utils = {
     return string.charAt(0).toUpperCase() + string.slice(1);
   },
 
+  isMobile() {
+    // return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);
+    return $(window).width() <= 800;
```
mquandalle: I think it is better to rely on the device size than the user agent indeed. So you should remove the comment above.

Also we should use Meteor reactivity here! The window width may change (eg if a browser user change resize the window, or with the multi-app mode of new iOs devices). The way to support that is to register a tracker dependency like the one I created [here](https://github.com/wekan/wekan/blob/677e9937e0c1872bfe4ec5ce87bf2a8ca99febe1/client/lib/popup.js#L5-L6) and used [here](https://github.com/wekan/wekan/blob/677e9937e0c1872bfe4ec5ce87bf2a8ca99febe1/client/lib/popup.js#L163).

Actually since the dependency is actually the same as in the Popup example, you could move its declaration in the Utils and use it like `Utils.windowWidthDep.depend();`.

Feel free to ask any question if something isn’t clear :)

client/lib/popup.js
```diff
@@ -161,6 +161,14 @@ window.Popup = new class {
     const $element = $(element);
     return () => {
       windowResizeDep.depend();
+
+      if(Utils.isMobile()) {
+        $('.pop-over').addClass('full');
```
mquandalle: Instead of imperative `addClass` and `removeClass` I would expose the `isMobile` variable to Blaze templates and use something like 

```jade
.pop-over(class="{{#if isMobile}}full{{/if}}")
```

client/components/main/popup.js
```diff
@@ -27,6 +27,9 @@ Popup.template.events({
 // we need to wait for the container translation to end before removing the
 // actual DOM element. For that purpose we use the undocumented `_uihooks` API.
 Popup.template.onRendered(() => {
+  if(Utils.isMobile()) $('.pop-over').addClass('full');
```
mquandalle: Same here: imperative vs declarative.

client/components/main/layouts.styl
```diff
@@ -355,3 +355,15 @@ a
 @keyframes flexGrowIn
   from
     flex-basis: 0
+
+@media screen and (max-width: 800px)
+  #content
+    margin: 1px 0px 49px 0px
+    height: calc(100% - 96px)
```
mquandalle: I never dared using `calc` because of browser support but that seem reasonable indeed http://caniuse.com/#search=calc. Cool!

client/components/boards/boardsList.styl
```diff
@@ -128,3 +125,18 @@ $spaceBetweenTiles = 16px
         font-size: 25px
         color: white
 
+@media screen and (max-width: 800px)
```
mquandalle: Maybe we could have a mode with 3 items per row, between the 2 on mobile and 4 on desktop?

client/components/main/header.styl
```diff
@@ -155,3 +88,109 @@
       border-left: 1px solid rgba(255, 255, 255, .3)
       height: 24px
       float: left
+
+#header-quick-access
+  color: white
+  transition: background-color 0.4s
+  background: #2573a7
+  padding: 10px 10px
```
mquandalle: Why did you enlarge the quick-access bar even on desktop (with adding 20px of vertical padding)? I would recommend to keep it unchanged on desktop.

Anyway I have some plan to transform this top bar into a left sidebar as does Rocket.chat for instance.

client/components/main/header.styl
```diff
@@ -155,3 +88,109 @@
...
+    z-index: 30
+    position: absolute
+    bottom: 0px
+
+    ul
+      width: calc(100% - 150px)
+      overflow: ellipsis
+
+      li
+        height: 28px
```
mquandalle: Always add a new line at the end of files.

client/components/boards/boardsList.styl
```diff
@@ -128,3 +125,18 @@ $spaceBetweenTiles = 16px
...
+    li
+      width: 50%
+
+    .board-list-item
+      overflow: hidden
+
+    .board-list-item-sub-name
+      position: relative
+      top: -100px
+      left: -100px
```
mquandalle: same, new line.

mquandalle:
This is very cool, thank you so much for working on that @floatinghotpot. Please consider my inline comments.

floatinghotpot:
@mquandalle Code optimized, please review again.

Thanks for the comments. I can always find a better solution after your review. :-)

client/components/boards/boardsList.styl
```diff
@@ -1,7 +1,7 @@
 $spaceBetweenTiles = 16px
 
 .board-list
-  margin: $spaceBetweenTiles ($spaceBetweenTiles/-2) 0
+  margin: 0 ($spaceBetweenTiles/2) 0 ($spaceBetweenTiles/2)
```
mquandalle: Oh btw, I generally use the shorten form if `margin-top == margin-bottom` and `margin-left == margin-right`, ie:

```styl
  margin: 0 ($spaceBetweenTiles/2)
```

client/lib/utils.js
```diff
@@ -22,6 +22,21 @@ Utils = {
     return string.charAt(0).toUpperCase() + string.slice(1);
   },
 
+  isMobile() {
```
mquandalle: Where this function is actually used?

Anyway, even if we need to rely on the user-agent somewhere (which is usually a bad idea) we should use a [issmobile](https://atmospherejs.com/mquandalle/ismobile) package instead of using a custom (and in this case, a bit simple) regular expression to test it.

client/components/boards/boardsList.jade
```diff
@@ -26,8 +26,6 @@ template(name="boardList")
 
 
 template(name="boardListHeaderBar")
-  h1 {{_ 'my-boards'}}
```
mquandalle: I think you unexpectedly removed this title on the home page (or is there a reason?)

client/lib/popup.js
```diff
@@ -183,7 +186,9 @@ window.Popup = new class {
       // was available and returns `false`. There is a (small) risk a false
       // positives.
       const title = TAPi18n.__(translationKey);
-      return title !== translationKey ? title : false;
+      // we need a default header [X], if we show popup as full of small screen
```
mquandalle: I don’t really understand what you mean by this comment?

client/lib/utils.js
```diff
@@ -22,6 +22,21 @@ Utils = {
...
+    return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);
+  },
+
+  windowResizeDep: new Tracker.Dependency(),
+
+  // in fact, what we really care is screen size
+  // large mobile device like iPad or android Pad has a big screen, it should also behave like a desktop
+  // in a small window (even on desktop), WeKan run in compact mode.
```
mquandalle: s/Wekan/Wekan/

client/lib/utils.js
floatinghotpot: This function is not used at all, so I am going to remove it.

mquandalle: :+1:

client/components/boards/boardsList.jade
floatinghotpot: I removed the title intentionally for purpose of simplicity. As there is already a logo as the title, the text looks a little superfluous on the main page, especially on the header of small screen.

client/lib/popup.js
floatinghotpot: If no title, then it's miniprofile mode. When full screen, we cannot see the header and find the [x] button to close it, so we need a default title.

client/components/boards/boardsList.jade
mquandalle: I would say, keep it on desktop and hide it on mobile. This is the visual position where we put a title on all pages.

client/lib/popup.js
mquandalle: ah, ok.

floatinghotpot:
polished again, now merge 3 commits into 1.

client/components/boards/boardsList.styl
```diff
@@ -128,3 +128,26 @@ $spaceBetweenTiles = 16px
...
+
+    .board-list-item-sub-name
+      position: relative
+      top: -100px
+      left: -100px
+
+@media screen and (max-width: 360px)
+  .board-list
+    height: 100%
+    overflow: scroll
```
mquandalle: Is there a need to repeat these 3 lines since there are already implied above by `max-width: 800px`

client/components/boards/boardsList.jade
mquandalle: You didn’t address this comment.

client/lib/popup.js
```diff
@@ -161,6 +161,9 @@ window.Popup = new class {
     const $element = $(element);
     return () => {
       windowResizeDep.depend();
```
mquandalle: You could use the new `Utils.windowResizeDep.changed()` here and remove the `windowResizeDep` declaration.

client/components/boards/boardsList.jade
floatinghotpot: Ok, done as you suggest.

mquandalle: iYes, sorry I didn’t refresh the diff view (when I said that you didn’t address my comment, I removed the message 1min after that when I realized that I was wrong)

mquandalle: Oh, you have actually done the change after my comment, that’s why :-)

mquandalle:
Perfect! Thank you very much for the feature and for implementing my recommendations :-)

floatinghotpot:
Cheers! :+1:

mquandalle:
Hum, I didn’t see that you changed the board header bar layout on desktop. I would prefer to keep it unchanged on desktop (it’s perfect on mobile), cf.

board-headerbar

PR #444, opened 2015-12-17
fix board header issue

floatinghotpot:


mquandalle:
Thank you!

floatinghotpot:
@mquandalle Thanks for your email. But when I reply, the email failed to be delivered with following message:
```
Delivery to the following recipient failed permanently:

    mquandalle@wekan.io

Technical details of permanent failure: 
Google tried to deliver your message, but it was rejected by the server for the recipient domain wekan.io by redirect.ovh.net. [213.186.33.5].

The error that the other server returned was:
553 sorry, that domain isn't allowed to be relayed thru this MTA (#5.7.1)
```

Do you have another email address?

mquandalle:
Oh woah, let’s use maxime[dot]quandalle[at]gmail[dot]com.

Issue #211, opened 2015-06-25
Feature idea: Same cards, multiple column sets


kentonv:
I've been using Libreboard to organize the roadmap for Sandstorm.io, and I thought of what I think would be an incredibly powerful feature.

The basic problem right now is that there are several different ways in which I might want to organize my cards. For example:

- Completion state (the usual wishes / not started / in progress / needs polish / done)
- Who is assigned to do the task
- Category
- How long it will take
- Milestones
- Priority

Right now, some of the above is handled by labels or by assignments, but the UI for these is far less pleasant than dragging cards around. Setting a label requires 5 precision clicks and delay time while UI elements load. Dragging cards between columns is way more fun.

So what I want instead is to be able to define multiple column sets for the same set of cards. Then, instead of setting labels or assignments, I would switch to the column set I am thinking about at the moment and push cards around.

Moreover, when I'm looking at one column set, the cards should show internal labels indicating the card's position is column sets I'm not currently viewing. So, e.g., when I'm looking at "completion state" I can also see each card's "priority" shown inside the card. (It may be nice to be able to choose exactly which labels are shown in order to avoid clutter... not sure. It would also be nice if I could color-code the labels by assigning colors to the columns from which they are derived.)

With this feature implemented, the usual label and assignments features could potentially be removed, simplifying the code and further differentiating from Trello.

What do you think?

ocdtrekkie:
I, for one, think this sounds really cool.

snolahc:
Hi,

This is an excellent idea, to abstract boards and cards, allowing multiple organizing schemas over the same cards.

Libreboard is the best tool to create lists.
In order to organize lists, we need to make lists of lists with priorities.
This tool allows us to be totally free on our organization, but could suggest a way to easily have a "todo list" from several boards where cards are assigned to me.

Managing those "duplicated/meta cards" would affect the underlying boards, cleaning them of the "done" stuff. This would be another board collecting cards with specific attributes/tags/...

PS: i created an issue but closed it after reading yours :)

UltimaTom:
The idea is very interesting/original, but i'm not sure about the use case, plus it may complexify the UI and the usability. 

For now I use Libreboard this way : 
- Completion state :arrow_right: I use the columns
- Who is assigned to do the task :arrow_right: Name initials or avatar + filters
- Category :arrow_right: I use labels
- How long it will take :arrow_right: I don't use that :p 
[...]

PR #458, opened 2015-12-28
Optimize for mobile web, show single list per page with navigate bar

floatinghotpot:
After optimization, it looks more like a mobile app in compact mode.
1. On board, show the list entries.
2. Click list entry, show a single list per page.
3. Click the list name item on the quick access bar below to navigate to list.

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/458)

### DIFF
--- a/client/components/boards/boardBody.jade
+++ b/client/components/boards/boardBody.jade
@@ -21,12 +21,21 @@ template(name="boardBody")
       if showOverlay.get
         .board-overlay
       .lists.js-lists
-        each currentBoard.lists
-          +list(this)
-          if currentCardIsInThisList
-            +cardDetails(currentCard)
-        if currentUser.isBoardMember
-          +addListForm
+        if isMiniScreen
+          if currentList
+            +list(currentList)
+          else
+            each currentBoard.lists
+              +miniList(this)
+            if currentUser.isBoardMember
+              +addListForm
+        else
+          each currentBoard.lists
+            +list(this)
+            if currentCardIsInThisList
+              +cardDetails(currentCard)
+          if currentUser.isBoardMember
+            +addListForm
 
 template(name="addListForm")
   .list.js-list.list-composer.js-list-composer
--- a/client/components/boards/boardBody.styl
+++ b/client/components/boards/boardBody.styl
@@ -43,3 +43,18 @@ position()
       .open-minicard-composer,
       .minicard-wrapper.is-checked
         display: none
+
+@media screen and (max-width: 800px)
+  .board-wrapper
+
+    .board-canvas
+
+      .lists
+        align-items: flex-start
+        display: flex
+        flex-direction: column
+        margin: 0
+        padding: 0 40px 0px 0
+        overflow-x: hidden
+        overflow-y: auto
+        position: cover
--- a/client/components/boards/boardColors.styl
+++ b/client/components/boards/boardColors.styl
@@ -56,6 +56,9 @@ setBoardColor(color)
       background: color
       color: white
 
+    &#header ul li.current, &#header-quick-access ul li.current
+      border-bottom: 4px solid lighten(color, 20%)
+
 .board-color-nephritis
   setBoardColor(#27AE60)
 
--- a/client/components/boards/boardHeader.js
+++ b/client/components/boards/boardHeader.js
@@ -51,10 +51,6 @@ BlazeComponent.extendComponent({
     return user && user.hasStarred(boardId);
   },
 
-  isMiniScreen() {
-    return Utils.isMiniScreen();
-  },
-
   // Only show the star counter if the number of star is greater than 2
   showStarCounter() {
     const currentBoard = Boards.findOne(Session.get('currentBoard'));
--- a/client/components/lists/list.jade
+++ b/client/components/lists/list.jade
@@ -2,3 +2,7 @@ template(name='list')
   .list.js-list(id="js-list-{{_id}}")
     +listHeader
     +listBody
+
+template(name='miniList')
+  a.mini-list.js-select-list.js-list(id="js-list-{{_id}}")
+    +listHeader
--- a/client/components/lists/list.js
+++ b/client/components/lists/list.js
@@ -112,3 +112,10 @@ BlazeComponent.extendComponent({
     });
   },
 }).register('list');
+
+Template.miniList.events({
+  'click .js-select-list'() {
+    const listId = this._id;
+    Session.set('currentList', listId);
+  },
+});
--- a/client/components/lists/list.styl
+++ b/client/components/lists/list.styl
@@ -53,6 +53,9 @@
   &.ui-sortable-handle
     cursor: grab
 
+  .list-header-left-icon
+    display: none
+
   .list-header-name
     display: inline
     font-size: 16px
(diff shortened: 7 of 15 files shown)


floatinghotpot:
Screenshot:

boardlist --- listentries --- singlelistperpage

mquandalle:
I spend a fair amount of time playing with this UI on my mobile and I’m impress that it works that well. So yes, that’s some very good work here, and I want to get it merged.

However there are also some design decisions involved in this new UI that requires some coordination, and probably some modifications in this PR. The biggest one is that I planned to remove the notion of list as a first class abstraction leaving only “boards” (or “projects”) and cards (or “tasks”). I wanted to write a full rationale for that change but a big part of the motivation is explained in #211. This change would have some repercussions on the mobile and desktop, but I obviously can’t blame you for not having anticipating this since I didn’t really talked about it.

Also at this point I think we need to think of the UI globally which may also involves changing the desktop UI. For instance you putted the favorites boards header bar on the bottom of the screen on the mobile version, which makes sense if you consider only mobile but is inconsistent with the desktop version. So maybe one solution to this problem would be to create a left swapping menu on mobile that would display as a left sidebar on desktop and which would contain the user informations and the list of boards in a similar way with what Rocket.Chat is doing. Another example is the “material” floating button to open the sidebar, I think it would make sense to have the same one on the desktop version.

I will try to come with a proposition and some mockups here, and I propose that we agree on some UI design decisions that would work on both desktop and mobile so that we don’t loose too much time changing the implementations details multiple time. Also if we end up changing a lot of UI elements (as you already have done in this PR), I would say that it is time to use React for new components as discussed in #299 (and now that the app build time has considerably decreased, yay!)

floatinghotpot:
Yes, some discussion and alignment is definitely needed on the UI decisions. 

Maybe the consistency can be more on content, instead of exactly same UX. Anyway, the screen/pointer on PC and mobile device is different.

The quick access board navigate bar, changing to a left sidebar is okay. But I think current PC appearance is not wrong and good enough, just like Chrome bookmark bar, to navigate between boards/sites. On mobile web, it can minimize as a bookmark icon in bottom bar (at right of home icon), to trigger a popup list of starred bards.

And, on mobile web, the round button to toggle sidebar on/off, will occupy the right side area of a very limited screen, and looks very strange on a mobile device. Maybe it's better to embed it as an icon in top bar. 

What do you think?

After this PR is merged, I think the heavy work on the layout for mobile web should be almost done. 

To get even better UX on mobile device, I agree it's time to turn to React and React Native.

mquandalle:
So I said that I needed to write done my thoughts about abstracting the notion of a list, here they are: https://discuss.wekan.io/t/abtract-lists-into-a-generalized-task-meta-data/121. I’m curious of what you think about it @floatinghotpot.

mquandalle:
I don’t know if you had time to read the above @floatinghotpot, but I shared it because with this new potential model in mind having a list-selection page and then display only the cards of this list wouldn’t make much sense. Being able to see cards instantly after clicking on a board (or “project”) is actually the desired flow. Sure the navigation need to be improved on mobile, I actually like the “tab system” that you created in this PR, and I think it would makes sense to just arrive on the board view with this tab bar just below the board header.

floatinghotpot:
@mquandalle Sorry for not replying in time. I read your thoughts, and I think it's good to abstract the cards with meta data. Actually, even when define labels, they can also be grouped into customized groups, like version/priority/module, etc.

The list in a Kanban board, to my understanding, usually means stage/progress, it's the core concept of Kanban methodology, value flow. It will be more flexible if you can define arbitrary dimension for a board, but, I will still ask the same question, is it over-engineering? Will it be enough to select any label group as vertical dimension? 

Another question jump into my mind is, now I can watch a list for progress and notification, I am just wondering, how to implement if the list is a meta data?

BTW, the animation gif showing swimming lane and dragging card to another user, it looks cool, but I am not very sure about its value in a real project, if several users all have multiple cards, the vertical space won't be enough. And, in this implementation, how to assign a card to multiple members?

The above questions are just my personal thinking, for your reference. If you do have good answers and decide to go forward, I will fully support you on it.

And, yes, you are correct. I definitely agree that "Sure the navigation need to be improved on mobile". I also feel current steps to enter a board on mobile is not very ideal, it does feel more compatible to see cards with tabs, instead of seeing a list of lists. I will try improving this.

nata-g:
@mquandalle  I was thinking along the same lines as @floatinghotpot, who articulated it better than I could have. Below is my .02 cents:-). 

While having  cards with meta labels  abstracted to be used as proxy to lists is useful, in using as swimlanes and animation and can give a bird-eye view of the point in time awareness, I am not sure  it is possible to capture all of them in an abstract meta-model with the various permutations and combinations that go along in a real life project. It can easily get overly complicated because there can be so  many configurations and my concern is it might  lead to unintended consequences. 

Like, for example, if we use a meta-label tag of  "Done", and map it as  one of the  dimension,  its definition is different for different projects. For some some projects "Done" means done of the whole project, while in other projects "Done" means closure in one stage but it is still a continuation in the whole scheme of things and goes to different stage where it might be the beginning of the same project.  

For Example, if I have a project of writing and publishing a book, and I assign "lists" (or meta-labels) denoting each chapter, every chapter can have different 'done' stages that is defined by the author depending on his workflow. May be his workflow is taking each chapter  instead of the whole book.  Like, one task might be the  'outline of each chapter' and other task might be 'writing each chapter' and after he is 'done' writing he might send to 'editing' to a different person(who is doing his own tasks )  before taking on to the next chapter. Just this workflow  piece can have so many configurations that needs some kind  of  'task completion' 'progress' , 'review', 'notifications' etc. As you can see the flexibility will be lost if we provide anything that is pre-defined. I am assuming other projects can be as complicated, if not more. 

I mean it is a good idea  and I can see how just having 'cards' in 'projects' and using 'labels' as a meta for 'lists' is useful and can make it easy to map in to different stages and creating cool views. It can be made more 'specific' and I guess the biggest complaint against Trello is it is too general, so if we can be able to create the same flexibility using 'meta' notion of lists then it would prove to be really useful and I can see definite advantages in using it. But I really think it is going to be really hard to define all those configurations in an abstract meta model.

mquandalle:
@floatinghotpot and @nata-g I have copy pasted your comments on the list abstraction proposal back in the forum thread: https://discuss.wekan.io/t/abtract-lists-into-a-generalized-task-meta-data/121/6.

smerschjohann:
I'm curious about the status of mobile support for Wekan. I just tried to use it on my android device and various things aren't working in the current state (version 0.54.0)

I've followed this conversation, but as the last comment is over a year back, I'm not sure what the current status is.

xet7:
@smerschjohann

Improving mobile web is at [Roadmap](https://github.com/wekan/wekan/wiki/Roadmap).

xet7:
Merged after conflict resolution at https://github.com/wekan/wekan/pull/1365 . Thanks a lot @floatinghotpot !

lichtamberg:
I have following 3 issues on my phone (Samsung Galaxy Note 8 with latest Chrome):

1) its really hard to click on the row names (board titles), most of the time it doesn't work and no click/touch event happens (independent if I click on the title or the arrow on the right). It also doesn't work with the Galaxy S Pen, so it's not related about my fingers :) 
2) Sometimes it starts the swap-rows feature, even if my click was just a "press" - not a drag and drop (maybe it's better to remove/deactivate the swap card-titles feature in general on phones?) 
3) When I achieve to reach a cards page, when I press the back button - I don't go to the cards index page back, but to the general board selection page.

Shall I open a new issue for this?

xet7:
@lichtamberg 

Your issue moved to #953

PR #461, opened 2015-12-30
Bugfix stylus error on mobileweb

floatinghotpot:
@mquandalle After you change package "mquandalle:stylus" to "stylus", the display and drag&drop of cards on mobile web (on iPhone Chrome and Safari) will be buggy. (It's looks okay on PC chrome narrow window.) 

This PR just revert this change.

When use "mquandalle:stylus": | When use "stylus":
---|---
correct | bug
---|---

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/461)

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -7,9 +7,9 @@ meteor-base
 
 # Build system
 ecmascript
-stylus
 standard-minifiers
 mquandalle:jade
+mquandalle:stylus
 
 # Polyfills
 es5-shim
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -99,6 +99,7 @@ mquandalle:jquery-ui-drag-drop-sort@0.2.0
 mquandalle:moment@1.0.1
 mquandalle:mousetrap-bindglobal@0.0.1
 mquandalle:perfect-scrollbar@0.6.5_2
+mquandalle:stylus@1.1.1
 npm-bcrypt@0.7.8_2
 npm-mongo@1.4.40-faster-rebuild.0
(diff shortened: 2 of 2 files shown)


mquandalle:
As indicated in c69f9930137fc6f530ab92468041f06c2df69c63 commit message I decided to switch the stylus package for build performance reasons (on my computer along with the 2 others changes reloading the client now takes 3s instead of 15). I hope you alos experienced some better reload time while developing! I also indicated that the generated output is slightly different because we miss some of the generated prefixed properties ([diff](https://gist.github.com/mquandalle/e95198626767b56fc63a)).

I mostly use Firefox and Chrome and Android for browsing, and in these browser the reported issue doesn’t appear. It seems that you are using Safari on iOs (my mother has an iPhone, I’ll test the mobile view on it). So basically there are 2 different solutions:

* Find a way to enable autoprefixing issues on the official `stylus` package (https://github.com/meteor/meteor/issues/5219) rather than switching back to `mquandalle:stylus`
* Implement performant caching compiler on `mquandalle:stylus` and witch back to it

I’ll decide shortly on one of these, thanks for raising the issue :-)

floatinghotpot:
Yes, the build is faster than before.

The above screenshot is from iOS Chrome. The issue happens on both Chrome and Safari on iOS.

mquandalle:
Hum, so maybe I should check again on Android/Chrome because I may have confused the branches (while testing your mobile improvements that were relying on `mquandalle:stylus`)

PR #468, opened 2016-01-02
added exact time as tooltip to activity time

choclin:
This is a very small change, but it helps very much to see the exact time of an activity.
The activity time is a good description relative to the current time (like "1 month ago"). But sometimes it is important to know the exact time. Now you only have to point your mouse onto it, and it show the exact time stamp.

I hope this helps others, too.

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/468)

### DIFF
--- a/client/components/activities/activities.jade
+++ b/client/components/activities/activities.jade
@@ -71,7 +71,7 @@ template(name="boardActivities")
           else
             | {{{_ 'activity-removed' memberLink cardLink}}}.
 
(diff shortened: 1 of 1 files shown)


mquandalle:
Yes, that makes sense. When we will migrate to React, we’ll probably want to refactor this duplicated code into a single small component, but it isn’t worth the effort to do that on Blaze. Thank you for this!

PR #478, opened 2016-01-08
ports in docker-compose.yml must be an array

seschwar:
Otherwise we run into the following issue with docker-compose 1.5.2:

    $ docker-compose up
    ERROR: Validation failed in file './docker-compose.yml', reason(s):
    Service 'wekandb' configuration key 'ports' contains an invalid type, it should be an array

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/478)


mquandalle:
Thank you!

PR #528, opened 2016-02-25
Added/updated translations for Spanish

mdeniz:
Fixed some wrongly translated items and added new ones


AlexanderS:
The translations are managed in transifex: https://www.transifex.com/wekan/wekan/dashboard/
You should update the translations there.

PR #531, opened 2016-02-29
Plural of Emoji

TheElf:
Minor thing, Plural of emoji is either emojis or emoji. 

This may be language dependent, so I can remove it from the other language files, but this is definitely true in US English

[Review on Reviewable](https://reviewable.io/reviews/wekan/wekan/531)


mquandalle:
Thank you!

PR #550, opened 2016-03-21
modefied japanese langage file

s-honda:
modefied japanese langage file

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/550)


AlexanderS:
The translations are managed in transifex: https://www.transifex.com/wekan/wekan/dashboard/
You should update the translations there.

s-honda:
thanks

PR #577, opened 2016-04-20
Fix initials avatar generation

AlexanderS:
The ES5 reduce method also needs a initial value. This bug was
introduced in aa974aa54ab6e5b7db7450206d12b44ffb3a0306.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/577)


mquandalle:
Thank you!

PR #578, opened 2016-04-20
Some UI enhancements for the public board view 

AlexanderS:
The public board view was kind of broken:

* The notification settings are only useful if you are logged in. It will use the email address from your profile, that is not there if you do not have a user account.
* If the board is private or if the board does not exists the options should not be visible. The visibility indicator for example will display "undefined". So the header bar should be empty if there is no board available.
* If you are member of a public board and you are currently not signed in, you may want to sign in to take changes if you follow a link to a board/card. So I think it is ok to replace the visibility indicator with a sign-in link for public boards. It does not provide any more value. If you can see the board without being logged in, it will always display "public".

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/578)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,27 @@
+# v0.11
+
+This release adds the following new features:
+
+* Notification system with email notifications of the changes in a board,
+  a list or on a card;
+* Show the exact time when hovering the activity time;
+* Allow to edit more easily longer card titles by resizing the input box;
+* Add shortcuts to move cards to the top or the bottom of a list;
+* A new log-in button on the public board view to sign in, even if the board
+  is published;
+
+and fixes the following bugs:
+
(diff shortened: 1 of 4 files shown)


mquandalle:
Could you please add a line in the `CHANGELOG.md` (under v0.11) with a description of your different changes?

AlexanderS:
I rebased this commits on the current devel, so that they can be merged more easily, added on commit that adds two missing title attributes in the board header (I missed that earlier and noticed it while rebasing) and added all changes since v0.10.1 to the changelog (hopefully I did not missed something).

mquandalle:
Oh woah, I didn’t ask for the full changelog, just the items you implemented, but thank you very much that’s perfect. I will release v0.11 with these changes soon.

PR #579, opened 2016-04-20
Do not publish the whole user doc of board members

AlexanderS:
The user document contains hashed passwords and hashed resume tokens.
We should only publish the required bits.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/579)


nidico:
This sounds pretty severe and worth a new release ASAP, no?

PR #581, opened 2016-04-20
Center text in the initails avatars

AlexanderS:
I had the impression, that the text inside the avatars is not really centered. Instead of adjusting the x-coordinate it seems easier to change the text-anchor to the middle and center the text.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/581)


mquandalle:
I didn’t know this `text-anchor` property, that’s a good idea, thank you :-)

Issue #548, opened 2016-03-20
Cover images should retain proportions


hgupta9:
When I add a cover image to a task it is displayed with a fixed 2:1 ratio.

Can the image proportions be retained instead?

![image](https://cloud.githubusercontent.com/assets/6797866/13905599/80de9f54-eeea-11e5-9958-9bbd6e6b8427.png)

PR #593, opened 2016-04-25
UI: Fix scaling of cover images (Fixes: #548)

### LINKED ISSUE #196: Card cover image breaks if attachment contains spaces in filename
If a attachment with spaces in the filename is uploaded and set as cover image, it will not display properly, due to wrong encoding of the spaces.

AlexanderS:
This basically reverts f039923ac134e4a3cc70a1a7d47c21460676b1c0 and fixes #196 in
a different way (adding quotes). So that we can use the css background
properties to scale the cover images.

We could decide, if we want to add a `background-color: black` or something like that to `.minicard-cover`.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/593)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and fixes the following bugs:
 * Fix the generation and alignment of the initials avatars;
 * Only display the buttons in the board header, if the data is avialable
   and the user is able to use it;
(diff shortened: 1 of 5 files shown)


AlexanderS:
This now also fixes the view of the attachments inside the card details.

PR #605, opened 2016-05-19
Fix move to top

AlexanderS:
If the `minOrder` is 0, the previous code does not work. This code is now
doing the obvious stuff to change the order. 

I could not imagine a case where this is not working.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/605)


mquandalle:
I encountered this bug, but didn’t realize it came from here, thank you for the fix! Merged in 81a35be8562cb6c28848afc2cb24ae9ec145ce47.

Issue #473, opened 2016-01-06
Bug: infrequent exception during invitation a new user to board via mail - TypeError: Cannot read property 'title' of undefined


rgoltz:
Hi,

I'm using wekan 0.10.1 (mongodb: v2.4.14 / node: v0.10.40). While I'm trying to add a new user -by typing a mail-address- to an existing board, I'm getting frequently:

```
Exception while invoking method 'inviteUserToBoard' TypeError: Cannot read property 'title' of undefined
    at Object.<anonymous> (models/activities.js:69:27)
    at packages/matb33_collection-hooks/insert.js:35:1
    at Array.forEach (packages/es5-shim/.npm/package/node_modules/es5-shim/es5-shim.js:417:1)
    at Function._.each._.forEach (packages/underscore/underscore.js:105:1)
    at after (packages/matb33_collection-hooks/insert.js:34:1)
    at Object.CollectionHooks.defineAdvice.self (packages/matb33_collection-hooks/insert.js:49:1)
    at Object.collection.(anonymous function) [as insert] (packages/matb33_collection-hooks/collection-hooks.js:117:1)
    at [object Object].Mongo.Collection.(anonymous function) (packages/mongo/collection.js:590:1)
    at [object Object].Mongo.Collection.(anonymous function) [as insert] (packages/aldeed_collection2-core/lib/collection2.js:101:1)
    at Object.Lists.after.update.Activities.insert.userId (models/lists.js:90:16)
    at packages/matb33_collection-hooks/insert.js:35:1
    at Array.forEach (packages/es5-shim/.npm/package/node_modules/es5-shim/es5-shim.js:417:1)
    at Function._.each._.forEach (packages/underscore/underscore.js:105:1)
    at after (packages/matb33_collection-hooks/insert.js:34:1)
    at Object.CollectionHooks.defineAdvice.self (packages/matb33_collection-hooks/insert.js:49:1)
    at Object.collection.(anonymous function) [as insert] (packages/matb33_collection-hooks/collection-hooks.js:117:1)
```

On UI side I can see "500" within a yellow box.

From my testing I would say, it's a timing issue: If I'm entering the address and click on "Invite via Email" really fast, I can see the Exception. If I wait a small amount of time (e.g. 2 sec) and then clicking on "Invite via Email" the mail is sent successfully.

I'm using a gmail-account on this installation along following parameter: `MAIL_URL='smtp://address@gmail.com:pwd@smtp.gmail.com:465/'` (fyi)

[...]

PR #616, opened 2016-05-31
Replace before.insert with autoValue from SimpleSchema

AlexanderS:
The before-insert collection hook has the serious problem, that it is called different if the insert is initiated on the server or on the client. From the client the hook is called before schema validation and can insert default values from required fields. When inserting from the server the hook is called after the validation of simple schema. So the validation fails, because the insert does not contain all required fields. This is a known bug from collection-hooks but there is not workaround available: matb33/meteor-collection-hooks#167

Because of this limitation, it is better to use a autoValue function in the simple schema.

This PR:
* replaces all before-insert hooks with appropriate autoValue functions
* removes a workaround for creating the welcome board after inserting the user
* fixes the creating of a welcome board for invited users
* add the possibility to localize the name of the welcome board (and lists)

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/616)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -74,6 +74,7 @@
     "Avatars": true,
     "BlazeComponent": false,
     "BlazeLayout": false,
+    "CollectionHooks": false,
     "DocHead": false,
     "ESSearchResults": false,
     "FastRender": false,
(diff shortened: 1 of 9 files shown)


blaskovic:
Will this be merged any soon? It should fix issue #473 which occours quite a lot.

mquandalle:
Merged. Please accept my genuine excuses for the long delay.

PR #625, opened 2016-06-13
Global Search Function

GSWemigo:
Is there any global search function , like in trello implemented?

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/625)


GSWemigo:
WHY my request got closed?

AlexanderS:
This should be an issue not a pull request.

GSWemigo:
If there is no search function thats its not an issue

AlexanderS:
The issue is, that there is no search function. **Pull Request** means, that you have an own branch with the required changes to fix a problem and want to get it merged in the upstream (see [Using pull requests](https://help.github.com/articles/using-pull-requests/)), it is not something like a Feature Request. You have opened a Pull Request requesting to merge *devel* into *meteor-1.3*, but *devel* does not fix your problem.

PR #628, opened 2016-06-20
Added login-buttons for packages

kasperrt:
* Added login buttons for packages. Makes it possible to add packages, such as accounts-google, accounts-github and such for easy and fast authentication up against third-party providers. Doesen't "damage" the already existing view for login if the user doesn't add any third-party authentications services.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/628)

### DIFF
--- a/client/components/main/layouts.jade
+++ b/client/components/main/layouts.jade
@@ -15,6 +15,9 @@ template(name="userFormsLayout")
     section.auth-dialog
       +Template.dynamic(template=content)
       div.at-form-lang
+      
+        | {{> loginButtons}}
+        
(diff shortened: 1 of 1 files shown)


AngeloMelonas:
Can someone possibly explain how I would go about adding a working "Sign in with Google" button? I added the accounts-google package and a "Sign in with Google" button appeared, however, when I click it I get an error:

`Error: invalid_client

The OAuth client was not found.`

I do have my ClientID and Secret key from Google.

AngeloMelonas:
A screenshot of what I currently have

PR #639, opened 2016-07-05
Add Cloudron install button

nebulade:
Similar to the other deployment options already in the readme, Cloudron offers a server platform to install and maintain wekan and other apps. Thanks for all the effort already gone into wekan, very good app :-)

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/639)


mquandalle:
Thanks!

PR #645, opened 2016-07-18
Update .eslintrc

mbeaudru:
Filter true repeated twice

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/645)


mquandalle:
The `eslint` file was already modified and no longer includes `Filter` twice.

PR #646, opened 2016-07-19
Fix showing card activity history in IE11

zarvox:
The animationend event doesn't seem to fire the way we expect it to under IE11.  This means that card history will never be visible when looking at card details in IE11, so work around it by also setting a timeout matching the animation time in case the animationend event never fires.

If you know of a better way to deal with this, I'm happy to adjust this patchset. :)

Tested in IE11 on Win7 via the modern.ie VMs: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/646)

### DIFF
--- a/client/components/cards/cardDetails.js
+++ b/client/components/cards/cardDetails.js
@@ -21,6 +21,14 @@ BlazeComponent.extendComponent({
     this.parentComponent().showOverlay.set(true);
     this.parentComponent().mouseHasEnterCardDetails = false;
     this.calculateNextPeak();
+    setTimeout(() => {
+      // Work around animationend event dispatch bug in IE11.
+      // The animationend listener is placed on the root node,
+      // not on the node specified by the selector.  As a result,
+      // IE never sees the completion of the animation, and never
+      // shows the activity list.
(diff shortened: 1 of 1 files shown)


mquandalle:
Thank you, that actually looks reasonable to me. The deffer-ed display has less to do with the actual animation than with Blaze slowdown on large list rendering, so having a timeout to say that the feed should be displayed after 100ms whatever happened isn’t that bad.

mquandalle:
Thanks for the hint to use Microsoft VMs, I was able to find the root cause of the bug, which is now fixed.

dwrensha:
bda1df5 does not seem to fix the issue for me on IE11.

dwrensha:
To be more precise: after https://github.com/wefork/wekan/commit/6090ede8c98d98ba5d3664169cae4f2696ccd557, the comments immediately show up on the versions of Edge, Chrome, and Firefox that I've tested. On IE11, I still need to wave my mouse over the card before the comments show up (but I guess that's better than the comments never showing up at all).

PR #649, opened 2016-07-20
Fix bug where you get a "null" option in @-mention autocomplete

zarvox:
When autocompleting an @-mention of a user, if any users would be filtered out by the currently-entered text, there will be an entry "null" shown in the autocomplete dropdown.  This is caused by null return values not getting filtered out of the possible completions array before being visualized.

This commit filters out falsy autocomplete options before passing them off to the UI.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/649)

### DIFF
--- a/client/components/main/editor.js
+++ b/client/components/main/editor.js
@@ -10,7 +10,7 @@ Template.editor.onRendered(() => {
       search(term, callback) {
         callback(Emoji.values.map((emoji) => {
           return emoji.includes(term) ? emoji : null;
-        }));
+        }).filter(emoji => !!emoji));
       },
       template(value) {
         const imgSrc = Emoji.baseImagePath + value;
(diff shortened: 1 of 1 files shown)


client/components/lists/listBody.js
```diff
@@ -186,7 +186,7 @@ BlazeComponent.extendComponent({
           callback($.map(currentBoard.activeMembers(), (member) => {
             const user = Users.findOne(member.userId);
             return user.username.indexOf(term) === 0 ? user : null;
-          }));
+          }).filter(user => !!user));
```
AlexanderS: This should not be needed here, because jQuery map, will already filter out `null` values.

AlexanderS:
This could also fixed by using the jQuery map, because it will filter the null values.

Another location with the same problem is the emoji search in `components/main/editor.js`.

client/components/lists/listBody.js
zarvox: Today I learned!  Backed out in bed1d55.

zarvox:
Updated to remove unneeded fix in `listBody.js` and to fix the emoji search as well.

PR #669, opened 2016-08-16
Fix remove user actions

### LINKED ISSUE #667: User removed from Board continue receiving email updates of that Board
Hi,
I have deployed Wekan's v0.11.0-rc2 release on Bluemix and have been testing it for a while.
I have an issue with email notifications.
A user was removed from a Board but is still being notified (by email) of the updates on that board.
He can't change his subscription configuration on the cards or activities in the board because he is not a member of the board anymore.
Probably Wekan should implement a unsubscription flow in cascade when a user is removed from a board.

AlexanderS:
This fixes the current actions (adding an activity) when a user leaves a board. The current check if the member is `$pull`'ed from the members list is wrong, because the user is only set to inactive (`isActive` will be false).

This also adds two new features:
* The removed user will be removed from all cards, lists and the board itself from the watchers list (and the member list of the cards). This fixes #667.
* If the board is not public, the board will be remove from the starred list of the removed user. This fixes the starred count of the board. The user could not see the board before but because the `boardId` was contained in the list, the user counted in the star count.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/669)

### DIFF
--- a/models/boards.js
+++ b/models/boards.js
@@ -494,6 +494,18 @@ if (Meteor.isServer) {
 
         const board = Boards._transform(doc);
         board.setWatcher(memberId, false);
+
+        // Remove board from users starred list
+        if (!board.isPublic()) {
+          Users.update(
+            memberId,
+            {
+              $pull: {
+                'profile.starredBoards': boardId,
+              },
+            }
+          );
+        }
       });
     }
   });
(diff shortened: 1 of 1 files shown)


mquandalle:
Thank you!

PR #715, opened 2016-10-19
fix a typo

umbertooo:
missing 'r' added

---

This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/715)


xet7:
I fixed this and another similar in other commit, thanks!

PR #716, opened 2016-10-19
Use navigator language by default instead of english

umbertooo:
There is no profile language set in a newly created user account. Wekan uses english as fallback language in this case. If there is no current user, Wekan takes care of the navigator properties. Instead we should always have a look at the navigator language.

If the navigator provides multiple preferred languages, select the most-preferred one.
https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages

---

This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/716)

### DIFF
--- a/client/lib/i18n.js
+++ b/client/lib/i18n.js
@@ -7,8 +7,14 @@ Tracker.autorun(() => {
   let language;
   if (currentUser) {
     language = currentUser.profile && currentUser.profile.language;
-  } else {
-    language = navigator.language || navigator.userLanguage;
+  }
+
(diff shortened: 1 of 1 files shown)


xet7:
Has conflicts, so won't merge. Closing issue.

xet7:
Possibly merging later.

Zokormazo:
I'm suffering this issue... any problem is I try to fix it properly?

Zokormazo:
I'm suffering this issue, any problem if I try to fix it?

PR #780, opened 2017-02-04
Deprecate Contributing.md

jLouzado:
Moving all this information to the Wiki, there's no need for there to be another file for contributor information. It just makes it harder to keep up to date.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/780)


xet7:
@jLouzado 

I only left link to wiki to Contributing.md file, these info about deprecation etc is not needed.

Thanks for notification about this, and moving Wefork wiki to Wekan wiki !

Issue #859, opened 2017-02-21
Remove someone from board = Activity generated for each card of board


fmonthel:
Hi,

On the last version, when we're removing someone from a board, even he don't have any card affected' and event will be generate for each card of the board to indicate that user x left the card
Should be only on affected cards

fmonthel:
PR raised : https://github.com/wekan/wekan/pull/861

xet7:
Merged.

PR #860, opened 2017-02-21
[BUG] Bug when we're removing user from board that generate activity for all the card of the board

fmonthel:
To resolve this bug : https://github.com/wekan/wekan/issues/859

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/860)


fmonthel:
to rebuild

Issue #784, opened 2017-02-06
Download file(unicode filename) action cause server process crashed with exception.


xet7:
Moved to here from https://github.com/wefork/wekan/issues/87

@yatusiter can you submit this as pull request to devel branch, so it shows you as author of fix in git history, and I can add you as contributor to changelog?

Originally by @yatusiter

When click download from card with unicode(Chinese) filename (upload is OK), the server process crash with following exception : 

> ./wekan/bundle/programs/server/node_modules/fibers/future.js:280
> throw(ex);
> ^
> TypeError: The header content contains invalid characters
> at ServerResponse.OutgoingMessage.setHeader (http.js:733:13)
> at ServerResponse.res.setHeader (/home/sjyb/install/wekan/bundle/programs/server/npm/node_modules/meteor/webapp/node_modules/connect/lib/patch.js:134:22)
> at packages/cfs_http-methods/http.methods.server.api.js:599:1
> at Function._.each._.forEach (packages/underscore/underscore.js:113:1)
> at packages/cfs_http-methods/http.methods.server.api.js:595:1
> 

----
After googled around, it's seems the HTTP header "Content-Disposition" cannot accept unicode encoding characters, and should be binary or URI encoded.

[https://en.wikipedia.org/wiki/List_of_HTTP_header_fields](url)
[https://my.oschina.net/jsan/blog/180333](url)

Fix with this solution:
#bundle/programs/server/packages/cfs_access-point.js add after line 444, FS.HTTP.Handlers.Get function：
```
const originalHandler = FS.HTTP.Handlers.Get;
FS.HTTP.Handlers.Get = function (ref) {
//console.log(ref.filename);
  try {
     var userAgent = (this.requestHeaders['user-agent']||'').toLowerCase();

        if(userAgent.indexOf('msie') >= 0 || userAgent.indexOf('chrome') >= 0) {
            ref.filename =  encodeURIComponent(ref.filename);
        } else if(userAgent.indexOf('firefox') >= 0) {
            ref.filename = new Buffer(ref.filename).toString('binary');
        } else {
            /* safari*/
            ref.filename = new Buffer(ref.filename).toString('binary');
        }   
   } catch (ex){
        ref.filename = 'tempfix';
   } 
   return originalHandler.call(this, ref);
};

```

xet7:
Possible to be related to https://github.com/wekan/wekan/issues/698

yatusiter:
[...]

Issue #698, opened 2016-09-28
Bad file name while attachment downloading


dinya:
I'm attaching files to my cards with cyrillic symbols in file name. But the file name is correct in the card view, but it is wrong while download.

I use Wekan v0.10.1 under Ubuntu 14.04 and nginx (charset utf-8 in site settings)

![wekan_cyrillic_attachment](https://cloud.githubusercontent.com/assets/2182222/18900334/d4677a50-8559-11e6-92d6-211b1b5f5d14.png)

gabrin:
Same trable

xet7:
Could be related to this:
https://github.com/wefork/wekan/issues/87

xet7:
Moved to #784

Dudki01:
My issue is just the same with only difference is  that i cannot apply code fix to cfs_access-point.js because there is no such file. The structure of applications has changed since then, is there any ways to solve this promblem now?
![image](https://github.com/wekan/wekan/assets/18345675/4f37d2d9-49ab-43d1-9d43-dd3ae99918c2)
![image](https://github.com/wekan/wekan/assets/18345675/3042fd4b-32a7-4118-a18f-8d41654a8d5f)

Wekan 6.0.9 with snap. Ubuntu 22.

ajimix:
I'm facing the same problem with spaces in filenames. `test file with spaces.pdf` becomes `test%20file%20with%20spaces.pdf`

Wekan 7.91 with snap. Ubuntu 22

xet7:
This works at WeKan 9.04

Issue #544, opened 2016-03-16
attachment(txt) uploading mozilla firefox 45


vitaliy-kuzmich:
after last firefox update, attachment uploading stops working. There is no reaction when press "Computer" button, also no errors in logs nor in browser console. But it works in Chromium.

xet7:
Uploading in Firefox seems to work now.

Issue #386, opened 2015-11-03
Avatars won’t display if the app is deployed under a sub-path


mquandalle:
Deploying Wekan under a sub-path of a domain name was recently implemented in devel (see #133) and will be shipped in the next version of Wekan.

I strongly suspect this “avatar not displaying” bug comes from Collection-fs not taking into account the `ROOT_URL` environment variable correctly to support prefixed URLs. I will probably open a PR to support it (which should be fairly similar to the support introduced in Flow-Router in https://github.com/kadirahq/flow-router/pull/329)

mquandalle:
I was slightly wrong on the problem description, I reported the bug to CollectionFs with a corrected explanations at https://github.com/CollectionFS/Meteor-CollectionFS/issues/884.

xet7:
Moved to #973

PR #875, opened 2017-02-27
Replace CollectionFS with meteor-file-collection

eemeli:
The library currently used by Wekan for attachments (CollectionFS) is not very actively maintained, having been deprecated for most of 2016. It's also somewhat extensive in its feature set, of which Wekan uses only a fraction but is burdened by bugs resulting from said complexity. So I'm here proposing to replace it with [meteor-file-collection](/vsivsi/meteor-file-collection/), which has mostly the same interfaces as CollectionFS and also uses GridFS for storage.

In addition to being simpler and more secure, the built-in [resumable.js](http://resumablejs.com/) integration will enable simultaneous upload of multiple files, as well as drag-and-drop for attachment upload.

The particular CollectionFS bug that I hit that prompted its replacement was its inability to deal well with local uploads, a part of a custom import job that I needed to get to work.

This is a work-in-progress, so should not be merged until the following are all done:
- [x] Basic attachment upload/download/delete support
- [ ] Avatar upload support
- [ ] Drag-and-drop for attachment upload (#300)
- [ ] Upload progress indicator
- [ ] Better fix for opening non-image attachments (#99)
- [ ] Re-enabling attachment import
- [ ] Migration script from CollectionFS

At the moment, using missing features result in console warnings with the prefix `[re-attach]`.

In addition to the above, other issues that would probably get resolved by this include: #142, #386, #544, #698, #784, CollectionFS/Meteor-CollectionFS#449.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/875)


xet7:
@eemeli

Thanks!

I think this can be merged when it automatically detects CollectionFS attachments and converts them meteor-file-collection. On some platforms like Sandstorm where Wekan code is signed and read-only, it's not possible to run separate migration scripts.

GhassenRjab:
Can other developers join and contribute to this PR? If yes, how?

xet7:
@GhassenRjab 

https://stackoverflow.com/questions/22237609/adding-to-someone-else-s-pull-request-on-a-github-project-that-i-do-not-own

eemeli:
Sorry, looks like I don't have the time I expected to have to work on Wekan stuff. The "Allow edits from maintainers" tag is set on this, so at least @xet7 should have push rights to the source branch in our fork. I'm also ok with someone else starting a new PR based on this, using their own fork.

PR #930, opened 2017-03-21
update sandstorm metadata for appVersion 19 release

dwrensha:
As of this writing, the spk is in queue to be published in the app market. You can test it out here: https://apps.sandstorm.io/app/m86q05rdvj14yvn78ghaxynqz7u2svw6rnttptxx49g1785cdv1h?experimental=true

It might be good if someone who knows about the new admin interface can verify that it's not accidentally accessible from the Sandstorm app. As far as I can tell it's not, but I don't really know where to look.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/930)

### DIFF
--- a/sandstorm-pkgdef.capnp
+++ b/sandstorm-pkgdef.capnp
@@ -22,10 +22,10 @@ const pkgdef :Spk.PackageDefinition = (
     appTitle = (defaultText = "Wekan"),
     # The name of the app as it is displayed to the user.
 
-    appVersion = 17,
+    appVersion = 19,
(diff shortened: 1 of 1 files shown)


dwrensha:
Ah, the notes [here](https://github.com/wekan/wekan/blob/devel/CHANGELOG.md#v0111-rc2-2017-03-05-wekan-prerelease) suggest that the database needs to be manually poked in order to make someone admin. So it looks like there's no danger of accidental admins on Sandstorm.

Issue #785, opened 2017-02-06
Double slash problem on card pop-ups


xet7:
Moved here from https://github.com/wefork/wekan/issues/49

# Summary

The card's pop-up links to URL like:
http://192.168.30.240/wekan//b/Gtw5WWes4kvanupHu/devel/HduAsvhiCJLPJF4Hb

Note the double slash //b, if I remove one slash, it works as expected. Also, navigating to the boards doesn't suffer from this.

This issue affects Wekan running behind webserver (Apache/Nginx/Caddy) proxy. Wekan only works when using locally without proxy, for example like http://localhost:3000 .

@Serubin tried to get that FlowRouter fix https://github.com/kadirahq/flow-router/pull/691 merged, with process like forking multiple other repos that would then be hosted under Wekan GitHub organization, but was not able to make Wekan build process working correctly.

For more details, see previous discussion at https://github.com/wefork/wekan/issues/49

Serubin:
My initial attempt was to update the remote flow_router repo to my own repo named serubin:flow_router. This however required me to update every subsequent repo that used any version of flow_router. I also ran into some odd dependency issues and decided it wasn't worth the effort.

My second attempt (which I've only just begun and am working on fixing up some meteor environment issues) is to add kadira:flow_router as a local package and let meteor use that over the remote versions. meteor hasn't been able to see my local package yet, but I'm still working to fix this.

ckuhn203:
I'd be happy to pitch in anyway I can. I spent most of the day trying to come up with an Nginx rewrite that would fix it and failed so far. 

I'm unfamiliar with meteor and flow router, but I'm wondering if it's possible to get off of the call to absoluteUri() and use a relative path in the view instead.

Serubin:
Currently the most realistic thing to do seems to be adding my fork of flowrouter to a "local package."

The flow router fix requires two packages, one of the main branch and another of the ssr branch. flowrouter separates the two for someone reason and requires both.


So far I've been unable to get meteor to see my local packages though. I've been dealing with some other stuff in life and haven't had time to investigate further into this issue.

Serubin:
The current issue with flow-router is this:
I went through and added some logging to the FlowRouter.url() function and found this output:

```js
  var completePath = this.path.apply(this, arguments);
  // /wekan/b
  var basePath = this._basePath || '/';
  // /wekan
  var pathWithoutBase = completePath.replace(new RegExp('^' + basePath), '');
  // /b
[...]

PR #932, opened 2017-03-23
Fixed: Double slash problem 

Serubin:
This fixes issue #785 by adding a patched local version of kadira:flow-router to wekan. The flow-router fix is explained in issue #785.

I tested this in three instances, each was newly installed in it's own location. The urls that were tested are as follows: `example.com/wekan`, `brd.example.com/wekan`, `brd.example.com`.

It's important to note that when building or starting the server it is required to execute the following command to export the local package directory for meteor: `source app.env`. All worked as expected.

The patched flow-router repo can be transferred to the wekan namespace upon request.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/932)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "packages/kadira:flow-router"]
+	path = packages/kadira:flow-router
+	url = git@github.com:Serubin/flow-router.git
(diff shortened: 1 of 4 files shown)


Serubin:
Hold off on merging this. Just realized there's a pretty big flaw. This won't handle paths if there isn't a prefix path.

Serubin:
Eslint was erroring on flow-router. I've changed the test command to ignore the packages/ directory to avoid this issue.

PR #945, opened 2017-03-28
Dockerfile fix for local packages

Serubin:
This in theory should be the docker fix for the double slash bug. 

It would be useful if someone could test this to make sure it works - I don't have a docker machine setup to do so.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/945)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,6 +77,7 @@ RUN \
     gosu wekan:wekan sh ./install_meteor.sh && \
     \
     # Build app
+    source /home/wekan/app/app.env && \
     cd /home/wekan/app && \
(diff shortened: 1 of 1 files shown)


Serubin:
This should be a fix for the issue mentioned in #945

stephenmoloney:
To be honest, this doesn't work, 

The docker build process exits with error 
`source: not found`

stephenmoloney:
@xet7 
This commit definitely breaks the docker build.
I was working on other ways to investigate if this fix works or not.

xet7:
@stephenmoloney 

Yes Docker build was already broken :D 

When it gets fixed correctly then I'll make new release.

PR #957, opened 2017-03-30
meteor update all packages

brylie:
Related issue #788 

Upgrade Meteor to 1.4.x branch

Upgrade Atmosphere packages

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/957)

Issue #928, opened 2017-03-21
Wekan Docker image build failing on Docker Hub


Barthalion:
v0.16 wasn't tagged on Docker Hub, so it's impossible to pull it this way.

xet7:
I'm not able to add tags, because I don't yet have access to Docker Hub of Wekan. You can always get latest version using tag "latest" as in docker install instructions, it's automatically built after every commit to repo. I'll try to contact @mquandalle sometime. More tasks to @mquandalle is at Wekan roadmap board that is linked on readme.

rubberduck203:
Is there any chance we can open this back up. It's important to me to be able to pin to a version and upgrade only after I've thoroughly tested it locally. 0.17 just released and now if I need to restart my app/server I'll get the latest version instead of 0.16.

rubberduck203:
So, this can worked around if you already have a copy of the 0.16 image locally. 
Tag your copy of the image yourself with this command.

```bash
docker tag 82e663c127dc mquandalle/wekan:0.16 
```

I'm going to also create a tag for the newest version. 
I'll see if I can set up a mirror.

rubberduck203:
Mirror up here.
https://hub.docker.com/r/rubberduck/wekan/

xet7:
@rubberduck203 

I have already sent support request to Docker Hub about adding tags, and also have direct contact to @mquandalle that has full access. Please be patient, it's taken care of. No need to open issues. :)

rubberduck203:
Thanks @xet, but it's not so much a patience thing as a "You just scared me." thing. It's the kind of behavior that means I can't recommend this fantastic project to my clients as an ultra low cost option. Reproducible deployments matter. I apologize if I panicked, but I have plans to migrate to a new server and this, well, *panicked* me. =)

Anyway, the mirror is up for anyone who wants to pin to 0.16.

xet7:
@rubberduck203 

I have sent message to @mquandalle that he adds tags to Docker Hub.

I don't know why some panic here, I try to be as friendly and welcoming as possible :)

rubberduck203:
And I didn't mean to be unfriendly. 
The panic is because without the tags, it is impossible to pin to a known stable version. I simply can't afford to risk automatically upgrading (without a way to go back) just because I bounced a container.

xet7:
@rubberduck203 

I'm currently creating new automated build with all tags to:
https://hub.docker.com/r/wekanteam/wekan/

[...]

PR #959, opened 2017-03-30
Dockerfile fix for local packages (Take 2)

Serubin:
**This hasn't been tested, I don't have a way to test this currently** 

Can someone please confirm that this works as expected BEFORE being merged in?

This should fix the local flow-router dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/959)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,7 @@ ENV SRC_PATH ${SRC_PATH:-./}
 
 # Copy the app to the image
 COPY ${SRC_PATH} /home/wekan/app
+ENV PACKAGE_DIRS="${SRC_PATH}/packages"
 
 RUN \
     # Add non-root user wekan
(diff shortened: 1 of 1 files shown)


stephenmoloney:
As previously discussed, you cannot run the command `source` like that inside a dockerfile.

This will not work.

https://travis-ci.org/wekan/wekan#L4565

Serubin:
@stephenmoloney Can you please offer a suggestion on an alternative way to add an environment variable?

stephenmoloney:
@Serubin 
I did [in this post](https://github.com/wekan/wekan/issues/928#issuecomment-289859392)

I even tried those changes for the `ENV`

It doesn't work. 

I would focus on https://github.com/wekan/wekan/pull/957 for now to get the Dockerfile up and running.

Serubin:
I see what's going on now. Have we looked into options like this? https://github.com/npm/npm/issues/9863

stephenmoloney:
The closest issue is documented [here](https://github.com/wekan/wekan/issues/928#issuecomment-289714017)

We need to upgrade meteor, various dependencies and node js.

That is ongoing now at #957

Serubin:
Understood.

Issue #951, opened 2017-03-29
Registration mail never sent using 0.16


anlx-sw:
Hello,

I currently run into problems implementing wekan using wekan-snap:
https://github.com/kubiko/wekan-snap/
https://uappexplorer.com/app/wekan-ondra.ondra

I first thought it might be a snap issue  - so i did some troubleshooting and opened a ticket in the wekan-snap repo:
https://github.com/kubiko/wekan-snap/issues/3

please read through that issue as it contains problem details and troubleshooting steps so far.

the author advised me to seek help at the wekan developers as the problem seems to be in wekan itself and not the snap implementation.

to reproduce the problem:
run ubuntu 16.04 server
```
sudo apt-get install snapd
sudo snap install wekan-ondra
sudo snap set wekan-ondra mail-url='smtp://10.100.11.50:25/'
sudo snap set wekan-ondra mail-from='wekan@domain.local'
sudo snap set wekan-ondra port='3000'
sudo snap set wekan-ondra root-url='http://wekan.domain.local:3000'
```
then register the first user and you get the error mentioned

i also tried already changing the mail url to:
'smtp://wekan@10.100.11.50:25/'
'smtp://wekan:password@10.100.11.50:25/'

i get the following error:
```
Mar 29 15:40:21 myservername snap[21719]: Exception while invoking method 'ATCreateUserServer' Error: getaddrinfo ESRCH
Mar 29 15:40:21 myservername snap[21719]:     at Object.Future.wait (/snap/wekan-ondra/46/programs/server/node_modules/fibers/fu                                          ture.js:449:15)
Mar 29 15:40:21 myservername snap[21719]:     at smtpSend (packages/email/email.js:89:1)
Mar 29 15:40:21 myservername snap[21719]:     at Object.Email.send (packages/email/email.js:192:1)
Mar 29 15:40:21 myservername snap[21719]:     at AccountsServer.Accounts.sendVerificationEmail (packages/accounts-password/passw                                          ord_server.js:797:9)
Mar 29 15:40:21 myservername snap[21719]:     at [object Object].Meteor.methods.ATCreateUserServer (packages/useraccounts_core/l                                          ib/server_methods.js:119:1)
Mar 29 15:40:21 myservername snap[21719]:     at packages/check/match.js:107:1
Mar 29 15:40:21 myservername snap[21719]:     at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
[...]

Issue #948, opened 2017-03-29
The mail From: field should specify that it is an email address, not just a name


vampirefrog:
In the admin page, as well as the [run-wekan.sh example](https://github.com/wekan/wekan/wiki/Install-and-Update#run-wekan-manually-if-youve-built-for-production) in the install guide, it is not specified what format the MAIL_FROM field needs to be in. Some mail servers generate an error when this From: field is not a correct email that they can find in their accounts list. The format is either: email@example.com, or Firstname Lastname <email@example.com>. Please add a help text or an extra field (one for name, one for email) to clarify this format. It has taken me 3 days of debugging and trials to figure this one out. Thanks

xet7:
@vampirefrog 

Does Admin Panel not work for you for setting email options? See Wekan GitHub project readme.

Issue #937, opened 2017-03-25
Bug: Invitation mails from (sender) is not set.


NoodleBB:
The "from" field is set in the Admin panel. But the invitationmails have no sender name.

lkisme:
So, the "from" value in your setting is a name, instead of an email addr?
if so, make sure your smtp host allows you to do this.

lkisme:
Are you using wekan version greater that 0.13?

Usually the from would be like `Thomas Anderson<thomas@anderson.com>`.

Just FYI, the value of the "from" field will be set into the MAIL_FROM env as before.

NoodleBB:
> Are you using wekan version greater that 0.13?

Yes. It's 0.16.

> Usually the from would be like Thomas Anderson<thomas@anderson.com>.

Yes. In my case it would be Wekan Admin<admin@wekan.myhost.de>

> Just FYI, the value of the "from" field will be set into the MAIL_FROM env as before.

The name was set in the admin panel in the field "From".

PR #968, opened 2017-04-01
Fix email settings loading

Zokormazo:
MAIL_URL was overriden with database info all the time.
Now if MAIL_URL exists is not overwritten and if neither MAIL_URL nor
exists valid admin panel data MAIL_URL is not set.
MAIL_FROM was ignored. Same behaviour, env variable has bigger priority
than database configuration.
On both cases, althrought environment variable is set, updating admin-panel
mail settings will load new info and ignore the environment variable.
Remove some code that is not needed anymore

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/968)

### DIFF
--- a/config/accounts.js
+++ b/config/accounts.js
@@ -16,17 +16,12 @@ AccountsTemplates.addFields([{
   template: 'invitationCode',
 }]);
 
-let sendVerificationEmail = false;
-if (process.env.MAIL_URL) {
-  sendVerificationEmail = true;
-}
-
 AccountsTemplates.configure({
   defaultLayout: 'userFormsLayout',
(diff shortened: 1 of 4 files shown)


Zokormazo:
May fix issues #951 #948 #937  

@anlx-sw @vampirefrog @NoodleBB can you test it?

anlx-sw:
my issue was fixed already before your commit. i used the admin panel  to set the mail configuration and it worked afterwards.

Zokormazo:
oh yeah, admin panel was working, the problem was with environment variables or testing mode without MAIL_FROM

PR #970, opened 2017-04-02
Add isAdmin to user Schema

### LINKED ISSUE #924: Add Feature: Set admin on first run
When you freshly install Wekan from the bare metal, there's no default user and no admin account. So i would suggest to create a first user / admin account on the first run, by setting a env variable to something like `FIRST_USER =potus44:y3sw3c4n:bo@whitehouse.gov`.

Zokormazo:
The code for making first user an Admin was there, but missing isAdmin field on userSchema.

Fixes #924

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/970)

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -409,7 +409,7 @@ if (Meteor.isServer) {
   });
   Accounts.onCreateUser((options, user) => {
     const userCount = Users.find().count();
-    if (userCount === 0){
+    if (!isSandstorm && userCount === 0 ){
       user.isAdmin = true;
       return user;
     }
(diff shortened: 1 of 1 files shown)


xet7:
@Zokormazo 

You need to first check that Wekan is not running at Sandstorm, and only then add first user as Admin, otherwise I cannot accept this pull request. See:
https://github.com/wekan/wekan/blob/devel/sandstorm.js

Zokormazo:
Ok

xet7:
So it needs to check environment variables first, is it running on Sandstorm.

Zokormazo:
Done

PR #925, opened 2017-03-19
Comment permissions

rhelsing:
see issue #870 .. probably needs a bit of refactoring, but it accomplishes the intended goal. Let me know what to do next. @xet7

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/925)

PR #976, opened 2017-04-04
fix updateUserPermissions() on Sandstorm

dwrensha:
Fixes breakage from https://github.com/wekan/wekan/pull/925.

The new isCommentOnly property needs to be set when Sandstorm updates user permissions, because otherwise a schema validation error occurs and users never get added as members of boards.

In this pull request we always set `isCommentOnly` to false. Eventually it might make sense to allow "comment-only" permissions to be used in Sandstorm. Doing so would probably involve adding a `canComment` permission (in the sandstorm-pkgdef.capnp) that defaults to `true`.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/976)


ocdtrekkie:
I would definitely be in favor of seeing the comment-only permission make it to Sandstorm at some point.

Issue #988, opened 2017-04-07
Wekan docker image segfaults with newer kernels


thypon:
```
$ docker version
Client:
 Version:      17.04.0-ce
 API version:  1.28
 Go version:   go1.8
 Git commit:   v17.04.0-ce
 Built:        Thu Apr  6 16:27:08 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.04.0-ce
 API version:  1.28 (minimum version 1.12)
 Go version:   go1.8
 Git commit:   v17.04.0-ce
 Built:        Thu Apr  6 16:27:08 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

```
$ uname -a
Linux pompel.me 4.10.8_2 #1 SMP PREEMPT Mon Apr 3 15:23:46 UTC 2017 x86_64 GNU/Linux
```
The problem seems to be related to https://github.com/CentOS/sig-cloud-instance-images/issues/62 the new `vsyscall=none` policy

stephenmoloney:
I don't know how to fix this issue.

same issue on the `alpine image` at wekan-launchpad?

thypon:
The problem is with the `debian:wheezy` glibc: switching to alpine or a more recent debian should do the trick.

stephenmoloney:
@thypon If you get a chance, does [this](https://github.com/wekan/wekan/pull/992) fix this issue?

thypon:
It fixes the problem. Thank you :) :+1:

PR #991, opened 2017-04-10
Met1.4

stephenmoloney:
Potential Fix for #988 - needs to be tested.

#### Note:

Versions `1.4.4` and `1.4.4.1` do not seem to have incorporated a fix for the EXDEV 
issue #928 YET.

So for now 1.4.4-rc.6 is the only working version.

Version 1.4.4-rc.9 has a different issue with a meteor dependency -- seems easy to fix. 

So I'd propose waiting for release of meteor 1.5 before changing dockerfil again and when 1.5 is released working through any issues at that stage.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/991)


stephenmoloney:
wrong branch

PR #1006, opened 2017-04-16
Binary package for debian and devuan

soohwa:
I uploaded binary packages into my repository for an easy install of wekan for testing only.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1006)


xet7:
Thanks. I already merged this manually.

PR #1017, opened 2017-04-28
changes to travis.yml.

stephenmoloney:
- Add gcc-4.9 to travis host.
- namespace travis variables as seem to conflict with docker env.
- set versions on travis.
- run npm tests again.
- slight readme mod for docker

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1017)


stephenmoloney:
Updating to `gcc-4.9` on the travis host is actually optional I think as the output from the travis build says that it is already newest version - so that can be removed in next PR from `travis.yml`. It is required in the docker container though. I'll make a PR for it now - might is well get it over with.

Issue #1037, opened 2017-05-11
Roadmap for REST-API features


mayjs:
@huneau Do you have a roadmap/list of features you plan to add to the REST-API? 
I would really like to use this API for multiple projects, but it is still missing some important features to accomplish my goals in these projects.
To be able to start working on them, I would like to help to implement missing endpoints, which is why I was looking for a roadmap.

huneau:
@mayjs I work for an employer, I have no roadmap no lists of features. 
I will maintain bugfix my own code but I don't spend the time to add a feature which is not decided by my product owner.

Up to you to contribute :)

mayjs:
Okay, I will start to create the API once I have the time to do it. Here is a list of endpoints and features I'm planning to add to the existing API:

- [ ] The current authorization is fine to protect a wekan instance, but to restrictive for my uses. Every user should be allowed to use the API, but an authorization system must be used to restrict the access to certain objects.
- [ ] Add `GET /api/user/boards`, this endpoint should return all boards the user is part of. (Instead of `/api/user/boards` we could also extend the functionality of `GET /api/boards`)
- [ ] `POST /api/boards/:boardId/lists/:listId/cards` could be extended to be able to update existing cards by giving an ID in the posted object.
- [ ] `GET /api/user/cards`

mayjs:
I added label management to the todo list. The assignment to existing cards can be handled in the card updating endpoint.

huneau:
> POST /api/boards/:boardId/lists/:listId/cards could be extended to be able to update existing cards by giving an ID in the posted object. (Or to stay restful, add POST /api/boards/:boardId/lists/:listId/cards/:cardId)

Update = PUT

mayjs:
I changed the Update endpoint to `PUT`.
For the filtering endpoint I was actually planning on some type of JSON query format, which is why I chose `POST` for it. I also was thinking about more than just simply searching for a name, which is why I did not want to use `GET`-parameters. But I get your point, we should use a `GET`-request because this endpoint does not add data to the database. On the other hand I am also not fond of passing JSON via `GET`.

I will change the text of the filtering feature to a more general description of the feature, I think a specification of what exactly should be possible with this feature is needed before deciding how to implement it.

mass3y:
How is auth working now? Can't get anything but "Unauthorized" response. Tried basic auth,  URL args, Cookies.....nothing appears to work. I am running a local instance, zero changes to Wekan config.
```
[...]

PR #1041, opened 2017-05-15
REST API features

mayjs:
This pull requests extends the REST API and changes the authorization system for some of the endpoints.
It implements some of the features listed in #1037. All changed or added endpoints are read-only.

The following changes were made:

- Add additional functions to `authentication.js`. It is now possible to check if a user should get access to a given board.
- Added the endpoint `GET /api/user` which allows to get basic information about the API user.
- Added the endpoint `GET /api/user/boards`. This endpoint can be used to get all boards this user is part of.
- Changed the reading access methods created by @huneau to be usable by any logged in user. I used a new authorization function which grants access to every board for admins, but restricts access for normal users. This change currently only affects boards, lists and cards, which are my main concern to be able to access.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1041)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ tmp/
 node_modules/
 .vscode/
 .build/*
+wekan.http
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ tmp/
 node_modules/
 .vscode/
 .build/*
-wekan.http
--- a/models/boards.js
+++ b/models/boards.js
@@ -562,13 +562,13 @@ if (Meteor.isServer) {
 
     const data = Boards.find({
         archived: false,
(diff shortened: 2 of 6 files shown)


huneau:
Hi @mayjs,

I not sure about that `GET /api/user/boards` this is not RESTFUL.
You use the request.userId to get the user but in REST you should use `GET/api/users/:userId/boards`
Otherwise it's good for me.

mayjs:
@huneau thanks for your feedback. You are right, it would be better to use `GET /api/users/:userId/boards`. I will fix this issue and add it to this PR.

huneau:
you have missed a `s` to `/user`

mayjs:
Oh yeah you are right. It's fixed.

PR #1059, opened 2017-06-12
Reenable the export feature

### LINKED ISSUE #1055: Board export not working
At least since version 0.22 the export menu item for a board does not seem to be doing anything. According to the webinspector there is no href nor any even attached to the dom element, but I might just not fully understand how it is patched up. Just tried v0.23 and it has still the same behavior. There is nothing in the webinspector logs nor on the server side from what I can tell.

Any ideas how to debug this further? Thanks.

nebulade:
Fixes #1055

Still does not fit nicely into the new REST API but at least allows the export to work again for now.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1059)

### DIFF
--- a/client/components/boards/boardHeader.js
+++ b/client/components/boards/boardHeader.js
@@ -15,17 +15,17 @@ Template.boardMenuPopup.events({
   }),
 });
 
-// Template.boardMenuPopup.helpers({
-//  exportUrl() {
-//    const boardId = Session.get('currentBoard');
-//    const loginToken = Accounts._storedLoginToken();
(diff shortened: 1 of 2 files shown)


xet7:
@nebulade 

On what menu in Wekan is Export option? I did not find it yet.

nebulade:
It is the export menu item above the board archive item, when you click the sandwich bar icon on the top right

huneau:
Hi, this is exactly the code I have commented. I was asking to my boss for time to fix that but I think the solution is enough.

ocdtrekkie:
@huneau I would note for the future that disabling a feature as important as export really shouldn't happen in a release version, IMHO.

PR #1091, opened 2017-06-27
Adding snapcraft build support from source

kubiko:
Adding snapcraft build support. This can be used together with launchpad build system for continuous integration.
Intention is to push snaps build from devel branch to edge channel, which can be installed by calling:
$ snap install —edge wekan
device will keep automatically updating to revisions as they are released to edge channel

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1091)


kubiko:
@xet7 Snap build job is now created, you can see it here:
[https://code.launchpad.net/~ondrak/+snap/wekan-devel]
it's building from devel branch, fee free to change that if you prefer it to build from different branch.
Same as for release apply here, auto builds on change, only resulting snap is pushed to edge channel instead.
Aso check please version name if you want it to be something different than: 0.26-SNAPSHOT

PR #1114, opened 2017-07-05
Fix Docker Build when USE_EDGE=true

### LINKED ISSUE #1110: Docker Deployment - double-slash with ROOT_URL
When doing a Docker deployment, I am experiencing the oft-reported Double Slash issue when using the ROOT_URL environment variable.  I see that there are several closed issues for this, but they seem to be from several versions ago, and I am wondering if there is something different about the Docker setup that is causing it to fail.

I tried both the wekanteam/wekan:latest tag as well as wekanteam/wekan:v0.27.  I also tried building a fresh clone from devel, same behavior.

jtickle:
This kind-of fixes #1110, but only if you build your Dockerfile with USE_EDGE=true.  The wekanteam/wekan:latest on Docker Hub still has the double-slash bug.

Note that the changes are: fixed a path from /home/wekan/packages to /home/wekan/.meteor/packages; and commented out the 'api.versionsFrom(...)' code in package.js for both kadira-flow-router and meteor-useraccounts-core.  I know nothing about Meteor so there may be a better way to fix this, but the change got it to where my Dockerfile would build, and now there is no double-slash issue.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1114)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,9 +96,11 @@ RUN \
       gosu wekan:wekan sh ./install_meteor.sh; \
     else \
       gosu wekan:wekan git clone --recursive --depth 1 -b release/METEOR@${METEOR_EDGE} git://github.com/meteor/meteor.git /home/wekan/.meteor && \
-      cd /home/wekan/packages && \
+      cd /home/wekan/.meteor/packages && \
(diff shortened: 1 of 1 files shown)


xet7:
@jtickle 

Please test latest wekanteam docker image after it's built, does this really fix it for you.

Issue #1045, opened 2017-05-22
Activity user messed up when creating a card using the REST-API


mayjs:
I noticed that the REST-API endpoint for adding cards to a list seems to generate activities with invalid user ids.
To reproduce this issue, it is sufficient to create a card using the API-Endpoint `POST /api/boards/:boardId/lists/:listId/cards` with appropriate JSON in the requests body.
While the membership of the user is generated correctly, the user for the activity seems to be set to an invalid or empty ID. This results in a buggy display:
![wekan_add_card_bug](https://cloud.githubusercontent.com/assets/5578376/26328390/14f02754-3f43-11e7-8297-2502637db9df.png)
I suspect the `Cards.after.insert` hook used to create the activities is missing the userId for some reason.

mayjs:
@huneau Are you still active on this project?
@xet7 any idea how we can set the current user in the REST-API in order to get this fixed? I am kind of stuck on this right now.

huneau:
Hi, 
@mayjs I'm looking notifications and I respond when I can. 
Otherwise, I've got a coworker that work on this project (don't know his pseudo).

Anyway, I transfer this issues to my coworker, because he will probably impact.

xet7:
@mayjs 

I have not yet looked at REST-API code with great detail yet so coworker of @huneau most likely is more up-to-speed on this.

zarnifoulette:
Hi, 

I fix the issue. I did a pull request.

PR #1116, opened 2017-07-07
Checking the authorId when adding card

zarnifoulette:
Hi, 

That is the fixe for the issue  #1045.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1116)

### DIFF
--- a/models/cardComments.js
+++ b/models/cardComments.js
@@ -56,6 +56,16 @@ CardComments.helpers({
 
 CardComments.hookOptions.after.update = { fetchPrevious: false };
 
+function commentCreation(userId, doc){
+  Activities.insert({
+    userId,
+    activityType: 'addComment',
+    boardId: doc.boardId,
+    cardId: doc.cardId,
+    commentId: doc._id,
+  });
+}
+
 if (Meteor.isServer) {
   // Comments are often fetched within a card, so we create an index to make these
   // queries more efficient.
@@ -64,13 +74,7 @@ if (Meteor.isServer) {
   });
 
   CardComments.after.insert((userId, doc) => {
-    Activities.insert({
-      userId,
-      activityType: 'addComment',
-      boardId: doc.boardId,
-      cardId: doc.cardId,
-      commentId: doc._id,
-    });
+    commentCreation(userId, doc);
   });
 
   CardComments.after.remove((userId, doc) => {
@@ -114,12 +118,16 @@ if (Meteor.isServer) {
     Authentication.checkUserId( req.userId);
     const paramBoardId = req.params.boardId;
     const paramCardId = req.params.cardId;
-    const id = CardComments.insert({
+    const id = CardComments.direct.insert({
       userId: req.body.authorId,
       text: req.body.comment,
       cardId: paramCardId,
       boardId: paramBoardId,
     });
+
+    const cardComment = CardComments.findOne({_id: id, cardId:paramCardId, boardId: paramBoardId });
+    commentCreation(req.body.authorId, cardComment);
+
     JsonRoutes.sendResult(res, {
       code: 200,
       data: {
--- a/models/cards.js
+++ b/models/cards.js
@@ -18,9 +18,9 @@ Cards.attachSchema(new SimpleSchema({
   listId: {
     type: String,
   },
-  // The system could work without this `boardId` information (we could deduce
-  // the board identifier from the card), but it would make the system more
-  // difficult to manage and less efficient.
+    // The system could work without this `boardId` information (we could deduce
+    // the board identifier from the card), but it would make the system more
+    // difficult to manage and less efficient.
   boardId: {
     type: String,
   },
@@ -64,8 +64,8 @@ Cards.attachSchema(new SimpleSchema({
     type: Date,
     optional: true,
   },
-  // XXX Should probably be called `authorId`. Is it even needed since we have
-  // the `members` field?
+    // XXX Should probably be called `authorId`. Is it even needed since we have
+    // the `members` field?
   userId: {
     type: String,
     autoValue() { // eslint-disable-line consistent-return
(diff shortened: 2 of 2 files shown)


mayjs:
This PR does not really fix the issue. While it does make sure to add the author as a member, the activity for the creation of the card is still wrong. I think this happens because of this code:
```
Cards.after.update((userId, doc, fieldNames) => {
    if (_.contains(fieldNames, 'archived')) {
      if (doc.archived) {
        Activities.insert({
          userId,
          activityType: 'archivedCard',
          boardId: doc.boardId,
          listId: doc.listId,
          cardId: doc._id,
        });
      } else {
        Activities.insert({
          userId,
          activityType: 'restoredCard',
          boardId: doc.boardId,
          listId: doc.listId,
          cardId: doc._id,
        });
      }
    }
  });
```

The `userId` parameter seems to be invalid since we do not really have a logged in user when using the REST-API, so the activity ends up getting a wrong userId (I think it is 0). The resulting activity is shown in the screenshot in #1045.

zarnifoulette:
I will investigate on it, into the next days.

mayjs:
If you don't find a solution for this problem, we could also try what happens if we use `restivus`. I think this issue would not occur when using this library, haven't tried it though.

zarnifoulette:
Tell me if you have better result with restivus, that interest me a lot . I observed the same issue when you want to put a comment in a card. The author doesn't appear.

mayjs:
So, I have looked into the issue, it does not seem like `restivus` works better on this.
And there is actually a good reason for it: the hooks are not a default part of meteor, but a library called [meteor-collection-hooks](https://github.com/matb33/meteor-collection-hooks). Luckily the readme of that library contains information on what to do if this issue occurs. The recommended way is to set `CollectionHooks.defaultUserId` to the correct userId. I'm not sure how this would interact with simultaneous requests though (If there is multithreading in meteor). Another way could be to use [Direct access](https://github.com/matb33/meteor-collection-hooks#direct-access-circumventing-hooks) to prevent the hooks from running. We could then move all hooks into separate functions and just call them manually from the REST-API.

Also, `restivus` does a pretty good job of defining the API, one part I like about it is that the integrated authentication works pretty good and there is an easy way to provide different API versions. But I don't think the benefits are enough to make the switch from our existing API. If you want to have a look at what I did, take a look at https://github.com/mayjs/wekan/commit/98dfbbf368bedf46bf7577b5c749507c4271633e and https://github.com/mayjs/wekan/commit/71f274dfa39931c6fa56169310342e8b90f98c57 for a usage example.

zarnifoulette:
Ok I don't think that is a good idea to switch for restivus too. But I don't understand why we have to call the hooks if we use direct access ?

mayjs:
Isn't the author of the card set in the object that is inserted into the collection? I think [this line of code](https://github.com/wekan/wekan/blob/devel/models/cards.js#L411) is responsible for that. The hook I quoted [above](https://github.com/wekan/wekan/pull/1116#issuecomment-313646111) is for inserting the activity for the card creation, which fails with the current code (and also the code of your PR).

If we use the direct access method, we can insert objects into collections without running the hooks. This would lead to non created activities if, for example, a card is created with the API. If we call the hook manually, this problem would be gone.

zarnifoulette:
Ok after testing the problem is solve if we do that for exemple : 

  JsonRoutes.add('POST', '/api/boards/:boardId/lists/:listId/cards', function (req, res, next) {
    Authentication.checkUserId(req.userId);
    const paramBoardId = req.params.boardId;
    const paramListId = req.params.listId;
    const check = Users.findOne({_id: req.body.authorId});
    if (typeof  check !== 'undefined') {
      const id = Cards.direct.insert({
        title: req.body.title,
        boardId: paramBoardId,
        listId: paramListId,
        description: req.body.description,
        userId: req.body.authorId,
        sort: 0,
        members: [req.body.authorId],
      });
      JsonRoutes.sendResult(res, {
        code: 200,
        data: {
          _id: id,
        },
      });
      Activities.insert({
        userId: req.body.authorId,
        activityType: 'createCard',
        boardId: paramBoardId,
        listId: paramListId,
        cardId: id,
      });


    } else {
      JsonRoutes.sendResult(res, {
        code: 401,
      });
    }
  });


This mess up the code but that is the only solution. I will do some external functions. And I will push all of that.

mayjs:
Thats what I meant. If you extract the Activity insertion into a function and replace the hook function (the `Cards.after.update`) part with a reference to that function, the code is a bit cleaner and we do not have the same code at two locations.

zarnifoulette:
I have a problem, do you know how can I update the _doc_ without using the hook ? Because we need to have the _doc_ update. For exemple (one function to do manually the job of the hook): 

 `function cardMove(userId, doc, fieldNames) {
  const oldListId = this.previous.listId;
  if (_.contains(fieldNames, 'listId') && doc.listId !== oldListId) {
    Activities.insert({
      userId,
      oldListId,
      activityType: 'moveCard',
      listId: doc.listId,
      boardId: doc.boardId,
      cardId: doc._id,
    });
  }
}`

If we create the card using the API and if we want to update it to move it in an other column the _  const oldListId = this.previous.listId;_ will not be initialize and we will have a problem.
For the moment I create a doc manually like this :  _const doc = {boardid: paramBoardId, listId: paramListId, _id: id};_. But I don't know, and I don't found in the documentation of the hook how can I modify the doc without using a hook.

mayjs:
According to the documentation, `doc` should be the object that was added to the collection. For the POST endpoint of cards you should be able to do something like `const card = Cards.findOne({_id: id});` to get the card that was previously created using `Cards.insert`, since the `insert` function returns the new id.

zarnifoulette:
My real problem is this line `  const oldListId = this.previous.listId;` I don't understand how can I set `previous `without calling a hook.

mayjs:
Looks good to me, haven't tested it though. I assume you tested all endpoints to work properly? Thanks for your work, I think I'll be able to progress on #1037 once this was merged.

PR #1133, opened 2017-07-15
Export and import attachments as base64 encoded files

GhassenRjab:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1133)


GhassenRjab:
Forgot to run the linter

Issue #799, opened 2017-02-08
Add feature: Import/Export Trello/Wekan board to/from Sandstorm Wekan.


xet7:
_From @xet7 on December 16, 2016 11:14_

Original request from @gaconkzk at Wekan Gitter.

_Copied from original issue: wefork/wekan#51_

xet7:
_From @tml on January 18, 2017 16:14_

With the Trello purchase, I suspect this might be something that a number of people become interested in rather quickly.

xet7:
@tml 

In standalone Wefork it's already possible to import Trello export JSON files, including attachment images. For example using VirtualBox appliance at https://wekan.xet7.org

This feature is not enabled for Sandstorm Wefork.

xet7:
_From @ocdtrekkie on January 18, 2017 16:20_

If the import from Trello requires actually logging into Trello, rather than importing like an export file, that'd probably be why, since it requires a network opening to do, and Sandstorm sandboxes the application. Future updates to Sandstorm should make this easier to do securely.

xet7:
Checklists are not imported yet, because that feature #11 is not yet merged to Wefork. Some other features can be missing also.

xet7:
_From @tml on January 18, 2017 16:30_

@ocdtrekkie I have an export file from Trello, and a standalone Wekan knows how to deal with those; there's just no opportunity to provide that information to the Wekan grain created by Sandstorm.

xet7:
_From @ocdtrekkie on January 18, 2017 16:32_

Should definitely tag @dwrensha on this one then.

tml:
This seems to be a duplicate of #800?

xet7:
@tml

Thanks, I closed #800 now. Duplicate came when I used https://github-issue-mover.appspot.com/ to move issues from Wefork to Wekan

xet7:
@FiloSottile said at #549 
I'd love this, too. My understanding from #113 is that #430 should make this possible.

I have not checked those yet.

brevilo:
If it means anything: +1 👍

ajrossnz:
Really looking forward to this. As wekan on its own doesn't have LDAP, we can't use that, so waiting on this feature before we go for implementation here.

xet7:
@ajrossnz @brevilo 

Please do not wait. In Open Source, although current code is available for free, the way to use it in your company is to have a programmer in your company implementing missing features and submitting them as pull requests.

[...]

PR #1134, opened 2017-07-15
Export and import attachments as base64 encoded files

GhassenRjab:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1134)

### DIFF
--- a/client/components/import/import.js
+++ b/client/components/import/import.js
@@ -3,7 +3,7 @@ import wekanMembersMapper from './wekanMembersMapper';
 
 BlazeComponent.extendComponent({
   title() {
-    return `import-board-title-${Session.get('importSource')}!`;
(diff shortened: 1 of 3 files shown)


xet7:
@GhassenRjab 

I tested with this Wekan board, that has attachment image, checklist and due date:

1-board

Export file seems to have base64 attachment.

Then I went to import that json file back. I don't know is there missing translation, or why title of import page seems strange:

2-paste

When I mapped Wekan board to users, and tried to import it, I got this error, so I was unable to import it:

3-map

Also see my comments about getting this to work on Sandstorm at #799

GhassenRjab:
I pushed two commits fixing both errors. The translated texts and JSON format verification

HLFH:
@xet7 I believe with this great PR, we can release a 0.29 version of wekan :)

Issue #1154, opened 2017-07-31
Add Feature: Web manifest to support 'progressive web app' use


danhawkes:
This [google reference](https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/) explains the advantages.

To do so should be pretty simple: at a minimum it requires serving a few additional static files - a manifest json file, and some higher res icons. 
  
<https://w3c.github.io/manifest/>

I think this'd also be relevant to offline support: #466

xet7:
@danhawkes 

Can you try to add pull request for those files to devel branch?

danhawkes:
Yeah, will try and have a look this evening.

Issue #466, opened 2016-01-02
Support offline mode


mquandalle:
It would be great to be able to use Wekan while offline with the confidence that modified data will be synced when the client gets back online. There are multiple component involved for this experience to became possible:
- **Keeping the application code into the browser cache.** Meteor has some support for [appcache](https://github.com/meteor/meteor/blob/353e8e3a4a2cc96a10f360e26c0636166c3b163e/packages/appcache/README.md) but this standard [is now deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache) in favor of the new [service workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers) that would allow to do a lot more things. We should build our offline mode using service workers.
- **Mutating the database state on the client.** Thanks to Meteor that allow us to build “rich-clients” this is already almost supported. We just need to keep the collections and the methods call into the browser cache so that there are not lost when the user close a Wekan tab while offline. [Groundb]() should be perfect for this job.
- **Reconcile concurrent state mutations.** We currently have a “last-write-win” strategy which will become very problematic with offline mode as the delays between the client write and the server reception will increase, which will also increase the number of potential conflicts. Depending on the fields and precise use cases we will want to use OT or CRDT (conflict-free replicated data type). An interesting project to follow on this matter is [Swarm](https://github.com/gritzko/swarm). As far as I know Meteor doesn't currently have any good integration with this kind of solution but I suspect that they would benefit a lot of Meteor applications.

WJakub:
H @mquandalle - I was wondering about any movement on the offline support front?

Considering building this in, pending a final decision. I would be happy to contribute my approach / code if I do.

Main answer I am seeking is: Has any work / planning begun on this please?

Thanks,
J

neooleg:
Hi @xet7, @mquandalle,

[PWA (Progressive Web Apps)](https://developers.google.com/web/progressive-web-apps/) should be faster way (in terms of development) to achieve offline support. Did you consider on PWA?

It also correlates with a set of ideas found:
- https://github.com/wekan/wekan/issues/480 -- React Native app is great, but some requirements from https://github.com/wekan/wekan/issues/480 can be covered with PWA, e.g. this offline support, faster loading the app, some integrations (push notifications) with OS.

And more over -- other PWA feature which are lack in native app will be available, e.g. one click install out of the app store.
[...]

PR #1184, opened 2017-08-20
Add web manifest

danhawkes:
To discuss. This allows Wekan to be launched as a standalone app via Chrome's and FF mobile's 'add to desktop' feature. See #1154.

Some details that could be improved upon:

- Icons have been copied from `/misc` into `/public` so they're served as static files. There may be a better way to do this as part of the build process, but am not familiar enough with meteor to suggest how.
- No localisation (again, a build process thing). May not be particularly important as it's just the description field that would benefit from it.
- I haven't included a service worker, which means you won't get an app install banner as described here: https://developers.google.com/web/fundamentals/engage-and-retain/app-install-banners/. Think this is relevant to #466, which would presumably involve a service worker. For now, a no-op SW could be added to fulfil the requirement, but not sure where to do so (`layouts.js`?). In any case, not sure whether this would be desired.

Please let me know what you think.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1184)


xet7:
I tested this using Chrome on Puppy Linux Finnish version [FinPuppy](http://www.keijos.arkku.net), it did work fine, Desktop Shortcut would work similarly:

```
/opt/google/chrome/chrome --app="http://192.168.10.100/sign-in"
```

On Windows Desktop Shorcut options did work similarly:

```
"C:\Program Files (x86)\Google\Chrome\Application\chrome.exe" --app="http://192.168.10.100/sign-in"
```

Thanks!

PR #1212, opened 2017-09-10
Decreasing the number of untranslated items (cs).

killua-eu:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1212)


xet7:
Only en.i18n.json is updated with pull requests when adding new features.

Other translations are done at Transifex:
https://www.transifex.com/wekan/wekan/

Thanks! :)

Issue #876, opened 2017-02-28
Reordering checklists has visual glitch (Already added Feature: Reorder Checklists and Checklist entries with click-and-drag, and drag from one checklist to another)


cyphar:
In Trello, you can click-and-drag to move checklist entries (just like cards). While this is mostly a graphical nicety, in my team we might write a checklistlist out-of-order and then later re-order it by priority or the timeline.

GuillaumeAmat:
By *checkpoint* you meant *checklist* right? ;)

Nefeth:
Please this would help immensely with writing action plans which has to be done in order. Multiple times I had to play around with copy pasting multiple lines and it is annoying. Don't know much about the difficultly of this feature though. If it is too hard, then just call it ;)

dejIO:
Any update on whether this can be done? Maybe by dragging and dropping list items?

We are in the process of migrating to Wekan in our office and some of our coworkers are having a hard time adapting their workflow, especially with checklists, to this new software.

xet7:
@dejIO

There could be some open or closed pull request or GitHub issue that also has some drag and drop for labels (or some other drag and drop) already implemented, but somebody would need to find it and add info to this issue.

There are some drag-and-drop features in Wekan elsewhere like when dragging cards in a list.

frmwrk123:
I managed to reorder items in a checklist. That is relatively easy for itself but the hierarchy of the DOM makes it practically impossible. The reason is that the card details and with it the checklist are below the container for the lists. But this container is already a sortable and multiple independent sortables inside a hierarchy don't seem to work with jQueryUI Sortable. For testing purposes I removed the code that makes it a sortable (inside boardBody.js) and immediately the checklist sortable worked.

If that problem is solved then it will be quite easy to implement this feature.

xet7:
Moved to here from https://github.com/wekan/wekan/pull/1215 

Noticed by @ocdtrekkie

> Quirks:
>
> - Somehow I managed to overwrite a checklist item by dragging and lost a couple that way while rapidly playing with this.
> - You cannot drag into a checklist that has no existing items. So, if you have two lists, you can be dragging around, drag them all into one list, and before you are able to drag them back into the other list, you have to manually add an item to that list so dragging works again.

cc @frmwrk123

xet7:
Moved to here from https://github.com/wekan/wekan/pull/1215#issuecomment-329415525

From @max-marschall

[...]

PR #1215, opened 2017-09-12
Add feature: Reorder checklist items 

frmwrk123:
This pull request partially implements #876. It makes checklist items sortable both within a checklist and across checklists. Checklists themselves cannot be sorted but model support for this was added as well.

The previous implementation of the model of checklists didn't support sortable checklist items so a bigger overhaul was necessary.

Known visual glitches: 
When dragging a checklist item to it's new place and releasing the mouse button, it doesn't persist immediately but reversed back to it's previous position. Less than a second later it is at the new position. Reason for this is the interior behaviour of Blaze. Blaze updates the template based upon database changes. Therefore we cannot modify the DOM with the jQuery UI Sortable widget and must cancel the movement after making the database change.
The small time difference between canceling the movement and the reactive change to the template is visible. But that should be acceptable.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1215)

### DIFF
--- a/client/components/cards/checklists.jade
+++ b/client/components/cards/checklists.jade
@@ -46,8 +46,8 @@ template(name="editChecklistItemForm")
       a.js-delete-checklist-item {{_ "delete"}}...
 
 template(name="checklistItems")
-  .checklist-items
-    each item in checklist.items
+  .checklist-items.js-checklist-items
+    each item in checklist.getItems
       +inlinedForm(classNames="js-edit-checklist-item" item = item checklist = checklist)
         +editChecklistItemForm(type = 'item' item = item checklist = checklist)
       else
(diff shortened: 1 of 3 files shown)


ocdtrekkie:
Quirks:

- Someone I managed to overwrite a checklist item by dragging and lost a couple that way while rapidly playing with this.
- You cannot drag into a checklist that has no existing items. So, if you have two lists, you can be dragging around, drag them all into one list, and before you are able to drag them back into the other list, you have to manually add an item to that list so dragging works again.

xet7:
Discussion moved to #876

max-marschall:
The following I actually see (came from 0.37 to 0.38 dockerhub-image):

- Extend existing checklists (before upgrade): Not possible
- Move checklist-entries from one to another checklist (existing checklist before upgrade): Items are lost immediately

= does not happen with checklists created **after** the upgrade

xet7:
Everybody:

Please continue discussion at #876 .

Do not add comments to this already merged pull request.

PR #1240, opened 2017-09-22
Add migration script to add sort field to checklists and their items

GhassenRjab:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1240)


ocdtrekkie:
Thank you for tackling this bug, @GhassenRjab!

Issue #1249, opened 2017-09-25
Can't login/create user in docker installation


erezeshkol:
## Issue
**Impacted version**: latest

I installed and ran the server using the instructions (but without setting up a mail server): https://github.com/wekan/wekan/wiki/Install-Wekan-Docker-for-testing

But I can't login or create a user, so it's completely unusable for me. When I try to create a user, an error is displayed: "From admin is required". Here are the docker-compose logs:
```
wekan_1    | Exception while invoking method 'ATCreateUserServer' Error: From admin is required
wekan_1    |     at getErrorObject (packages/aldeed_collection2-core.js:480:15)
wekan_1    |     at [object Object].doValidate (packages/aldeed_collection2-core.js:462:13)
wekan_1    |     at [object Object].Mongo.Collection.(anonymous function) [as insert] (packages/aldeed_collection2-core.js:214:25)
wekan_1    |     at AccountsServer.Ap.insertUserDoc (packages/accounts-base/accounts_server.js:1328:25)
wekan_1    |     at createUser (packages/accounts-password/password_server.js:1050:25)
wekan_1    |     at AccountsServer.Accounts.createUser (packages/accounts-password/password_server.js:1119:10)
wekan_1    |     at [object Object].Meteor.methods.ATCreateUserServer (packages/useraccounts_core.js:1294:27)
wekan_1    |     at packages/check.js:130:16
wekan_1    |     at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
wekan_1    |     at Object.exports.Match._failIfArgumentsAreNotAllChecked (packages/check.js:129:41)
wekan_1    |     at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1734:18)
wekan_1    |     at packages/ddp-server/livedata_server.js:719:19
wekan_1    |     at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
wekan_1    |     at packages/ddp-server/livedata_server.js:717:40
wekan_1    |     at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
wekan_1    |     at packages/ddp-server/livedata_server.js:715:46
wekan_1    | Sanitized and reported to the client as: From admin is required [400]
wekan_1    |
```

temperatio:
Same here.

xet7:
@erezeshkol @temperatio 

[...]

PR #1251, opened 2017-09-25
Fix: Add Feature: REST API: Create user despite disabling registration https://github.com/wekan/wekan/issues/1249

soohwa:
I just did set optional to false instead of true ...
Sorry ... :-D

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1251)

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -537,7 +537,7 @@ if (Meteor.isServer) {
     if (doc.createdThroughApi) {
       // The admin user should be able to create a user despite disabling registration because
       // it is two different things (registration and creation).
-      // So, when a new user is created via the api (only admin user can do that) one must avoid 
+      // So, when a new user is created via the api (only admin user can do that) one must avoid
       // the disableRegistration check.
       // Issue : https://github.com/wekan/wekan/issues/1232
       // PR    : https://github.com/wekan/wekan/pull/1251
--- a/models/users.js
+++ b/models/users.js
@@ -594,7 +594,7 @@ if (Meteor.isServer) {
     const id = Accounts.createUser({
       username: req.body.username,
       email: req.body.email,
-      password: 'default',
+      password: req.body.password,
       from: 'admin',
     });
 
--- a/models/users.js
+++ b/models/users.js
@@ -535,6 +535,12 @@ if (Meteor.isServer) {
   Users.after.insert((userId, doc) => {
 
     if (doc.createdThroughApi) {
+      // The admin user should be able to create a user despite disabling registration because
+      // it is two different things (registration and creation).
+      // So, when a new user is created via the api (only admin user can do that) one must avoid 
+      // the disableRegistration check.
+      // Issue : https://github.com/wekan/wekan/issues/1232
+      // PR    : https://github.com/wekan/wekan/pull/1251
       Users.update(doc._id, { $set: { createdThroughApi: '' } });
       return;
     }
--- a/models/users.js
+++ b/models/users.js
@@ -534,8 +534,10 @@ if (Meteor.isServer) {
 
   Users.after.insert((userId, doc) => {
 
-    if (doc.fromAdmin)
+    if (doc.fromAdmin) {
+      Users.update(doc._id, { $set: { fromAdmin: '' } });
       return;
+    }
 
     //invite user to corresponding boards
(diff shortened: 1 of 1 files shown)


xet7:
@soohwa 

With this new pull request:
- Wekan install with creating first user admin etc works
- When I use REST API, it looks like it did create new user
- But I can not login as new user with new user's username and password

Thanks, there's progress! :)

soohwa:
I will take a look

xet7:
@soohwa 

Now when I create new user with API, and login as new user, new user has admin rights. I tried to setup fromAdmin: false, but I did not yet find a way to create non-admin user. Is there a way, or can you add it?

soohwa:
I will take a look

soohwa:
I assume you have already a first user who is the admin (it is the desired behaviour). I cannot reproduce the creation of a second user who will have admin rights :-(

xet7:
Yes.

1) I did create at Wekan webpage first user that became admin.

2) Then I created new user with rest api.

3) But new user has admin rights.

4) How can I create normal user with rest api? User that doea not have admin rights?

4) How to create normal user wirh

soohwa:
The point 3 is really weird because there is nothing in the code which grant the new user as admin (isAdmin = true). And when I log in in the interface with the second user I'm not admin and when I retrieve the user's information via /api/users/:id there is no presence of isAdmin. So I really can't explain this. I can just suppose the environnement maybe is different and maybe the isAdmin field is present and it was set to true by default ... (but I'm pretty sure I'm wrong on this point) or maybe this line `const userCount = Users.find().count();` is not working properly ...

soohwa:
Anyone feels free to review in order to find this bug ...

nztqa:
My execution log.
"isAdmin" is not inserted.

1. fromAdmin": "true"
```
$ curl -H "Authorization: Bearer xl4zhal8HhMyktnlaQ9E7fTISTFduFb2uem1KZ9pAsm" \
  -H "Content-type:application/json" \
  -X POST \
  http://localhost:3000/api/users \
  -d '{ "username": "tester", "password": "tester", "email": "tester@example.com", "fromAdmin": "true" }'

{
  "_id": "AT6MfLgJ3QrP4DM8n"
}
```
result
```
{
    "_id": "AT6MfLgJ3QrP4DM8n",
    "createdAt": ISODate("2017-10-05T11:29:38.513Z"),
    "services": {
        "password": {
            "bcrypt": "$2a$10$jjNR4MuK7hw1a9l9gesdvOLjHwAspoSxcPO6WDaJi2y2DhUabPmmO"
        }
    },
    "username": "tester",
    "emails": [
        {
            "address": "tester@example.com",
            "verified": false
        }
    ],
    "fromAdmin": true,
    "profile": {}
}
```


2. fromAdmin": "false"
```
$ curl -H "Authorization: Bearer xl4zhal8HhMyktnlaQ9E7fTISTFduFb2uem1KZ9pAsm"
  -H "Content-type:application/json"
  -X POST \
  http://localhost:300/api/users -d '{ "username": "tester1", "password": "tester1", "email": "tester1@example.com", "fromAdmin": "false" }'

{
  "_id": "CiP6gx5F3sBASK4JA"
}
```
result
```
{
    "_id": "CiP6gx5F3sBASK4JA",
    "createdAt": ISODate("2017-10-05T11:30:41.688Z"),
    "services": {
        "password": {
            "bcrypt": "$2a$10$6ZwguctvWqo0TP31gSiad.P73ADiJdK2AykPGSG/sjBo0aysdlOo2"
        }
    },
    "username": "tester1",
    "emails": [
        {
            "address": "tester1@example.com",
            "verified": false
        }
    ],
    "fromAdmin": true,
    "profile": {}
}
```

soohwa:
@nztqa You're right: the only purpose of `fromAdmin` is to avoid "disableRegistration == true", not to grant admin rights like `isAdmin`. And yes, it does not need to be save. Thanks for your patch, i will include it soon.

nztqa:
isAdmin field = Administrator
fromAdmin field = avoid "disableRegistration == true"

Because the name is confusing, it is the source of misunderstanding.
So it's better to change to an appropriate namea nd  add the issue url and comments.

PR #1294, opened 2017-10-12
Fix and update Turkish Translations

vedia:
Fix and update Turkish Translations

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1294)


xet7:
Translations are done in Transifex, please add changes there:
https://www.transifex.com/wekan/wekan

Only English changes come as with pull requests when adding new features.

Thanks!

Issue #1311, opened 2017-10-26
Feature Request: WIP Soft Limits


darkuranium:
I think it would be nice to have the ability to make the WIP Limit on a list a "soft limit" --- that is, allow the user to go over the limit, but show some sort of a warning (e.g. different color of the column, different background, ...).

Most other Kanban boards --- at least the ones I've tried --- seem to use a soft limit by default (with no option for a hard limit), which makes them more flexible.

**Example:**
If I set the limit of a column to 3, I can at the moment not create more issues there, nor can I drag more into that list. However, sometimes, the situation may demand that I create (or move) more anyhow. For that, it would be nice to have the ability to do so *anyhow*, so that you can have e.g. "5/3" in progress (even if it's very undesirable).

xet7:
@amadilsons

Can you look at this?

amadilsons:
For sure, I will add a 'Soft Limit' option.
Sorry it took me so long to answer

darkuranium:
Amazing, thank you!

And better yet, I now know *exactly* which platform I'll use as a self-hosted alternative to Trello (no points for guessing).

amadilsons:
I've just submitted a new PR #1319 with this feature.
I've tried to test it for all the use cases I could remember but might have missed something!

Feel free to check it out and see if suits your needs :)

PR #1312, opened 2017-10-27
Permit editing the WIP limit

### LINKED ISSUE #1310: Issues editing WIP Limits
## Issue
**Impacted version**: 0.51

**Server Setup Information**:
Sandstorm

**Problem description**:
WIP Limits cannot be edited: When creating a WIP limit, it autopopulates the current number of cards in the list, and I am unable to change that text value. This pertains to #1278 and I am tagging @amadilsons as an FYI.

*Update:* On my work browser, I can see "up/down" arrows I couldn't see while testing Wekan 0.51 on my other device. The up/down arrows work correctly when visible, I expected to be able to change the number with my keyboard (which I believe should be an option). So this may be two issues: The number should straight-up be editable so that you don't have to spam click to get a high number, and we need to figure out when the buttons don't show for me and fix those.

ocdtrekkie:
This fixes #1310 and required for Edge users to be able to edit the WIP limit. (The up/down buttons appear to be "optional" according to the HTML5 spec, and can't be assumed to be present.)

I also believe I fixed a spelling error in the same section.

This PR is **NOT TESTED**

Also, @amadilsons for opinions.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1312)

### DIFF
--- a/client/components/lists/listHeader.jade
+++ b/client/components/lists/listHeader.jade
@@ -76,14 +76,14 @@ template(name="listDeletePopup")
 
 template(name="setWipLimitPopup")
   #js-wip-limit-edit
-    lable {{_ 'set-wip-limit-value'}}
+    label {{_ 'set-wip-limit-value'}}
     ul.pop-over-list
       li: a.js-enable-wip-limit {{_ 'enable-wip-limit'}} 
         if isWipLimitEnabled  
           i.fa.fa-check
     if isWipLimitEnabled
       p
-        input.wip-limit-value(type="number" value="{{ wipLimitValue }}" min="1" max="99" onkeydown="return false")
+        input.wip-limit-value(type="number" value="{{ wipLimitValue }}" min="1" max="99")
         input.wip-limit-apply(type="submit" value="{{_ 'apply'}}")
         input.wip-limit-error
         


nztqa:
@amadilsons And i have to add validation on input. 
See https://stackoverflow.com/questions/20964288/how-to-make-html-input-tag-only-accept-numbers

[Free IE and Edge Virtual Machines](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/)

ocdtrekkie:
So still ccing @amadilsons , I may have cause an... intriguing new bug. Now it's editable, and if you expand and collapse the WIP limit once or twice, it DOES get the current card number set in the WIP limit field... but the first time you enable the WIP limit on a list, it says 0 instead of the number of cards in the list.

ocdtrekkie:
Actually, nope, it wasn't me. Same thing happens sometimes on 0.51. No idea why.

amadilsons:
That happens because 0 is the default value for the limit when the list is created. Thinking about it, it isn't very smart..

Just to make sure I understand the problem, the value 0 only shows up when you enable the WIP limit for the first time on an empty list, right? Because that should be the only case where that happens.
If you enable the limit for the first time in a list, and that list already has cards in it, it should default to the number of cards already on the list. Is that what happens to you??

ocdtrekkie:
@amadilsons When I tried to enable a WIP limit on a list with several cards, the value 0 showed up with a red border around the box because it was too small. Clicking to disable the WIP Limit and then again to enable it again caused the box to properly populate with the number of cards in that list.

amadilsons:
So, I've just submitted a new PR for the feature request #1311 and at the same time tried to address the problem you described.

Unfortunately, I was not able to recreate the problem you mentioned!! The only situation I could see where the WIP value was 0, was the one I described in my previous comment. I might have also not understood your issue exactly..

Anyhow, whenever you have time, test my new PR #1319 for the problem you were having and check if it still occurs. I've tested it for the use cases I could remember, but might have missed something.

Issue #1255, opened 2017-09-26
Members do not get included on board import from WeKan


AllTinSnowyCold:
## Issue
**Impacted version**: v0.39
* Deployment Method: tar

As of v0.39, members do not actually get included on new boards when they're created via an import from WeKan JSON. This should also be an issue for importing from Trello, although I haven't tried it.

It looks like this mostly boils down to a couple issues in the "createBoardAndLabels" method of WekanCreator.

- const wekanId = wekanMember.userId;
    There is no "userId" on the wekanBoard.members objects, at least when importing from WeKan JSON. Instead this should probably be "wekanMember.id". I'm not sure if this part is also a problem for Trello.

- const wekanId = this.members[wekanId];
    In the transpiled version of this code that's in the WeKan tar, this is "var _wekanId = _this.members[_wekanId]", which is just going to end up being an undefined value, since _wekanId hasn't previously been set. Easily fixed by just using a different variable name for this wekanId.

- The reuse of the "wekanMember" variable name when determining whether the new board already has a given wekanMember.
    First the code checks that this "wekanMember" is falsy in an "if" conditional, but then it tries to access a property on the object in the "if" body, which should actually be on the shadowed version of "wekanMember". Should be fixed by just giving this "wekanMember" a different name.

I'm afraid I don't have an easy way to do a full build from the source here, but changing the above items in the pre-built app.js fixed the issue for me.

GhassenRjab:
This is odd because when exporting a Wekan board, members have `userId` field and not `id`

AllTinSnowyCold:
During the Wekan board import, the member objects have both `id` and `wekanId` for me, although they have the same value.

`{ isAdmin: false,
    isActive: true,
    isCommentOnly: false,
    id: '7DgBEK1hf4DQLHeG3',
    username: 'test',
    selected: true,
    wekanId: '7DgBEK1hf4DQLHeG3' }`

I'm just working from the tar with the minified client code, but it looks like this is probably due to the `getMembersToMap` function in wekanMembersMapper.js, which replaces the `userId` with `id` and potentially adds `wekanId`.

Tentoe:
the userId to id and  wekanId change is weird and happens [here](https://github.com/wekan/wekan/blob/01507576b1c3c3ad148969691d1e22bf7addbeb9/client/components/import/import.js#L37) I think. I don't know why that happens. As far as I can tell id is the old id and wekanId is the id the user is being mapped to.
Long story short. That wasn't the problem.

PR #1316, opened 2017-10-29
fixed Members do not get included on board import from WeKan #1255

Tentoe:
fixed multiple errors where constants were mistakenly overridden.(or
trying to get a property of an object after making sure it doesn't
exits)

renamed wekanBoard Parameter from createBoardAndLabels for clarity

reworked adding admin for importer

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1316)


Tentoe:
Didn't cover the case when only nonadmis are mapped.

Tentoe:
Covered case of no admin and no self mapping.

Now importer is always admin of the imported board so that no 
undeletable boards will be created.

Issue #1096, opened 2017-06-29
Maximum call stack size exceeded


GavinLilly:
Seemingly at random I find that Wekan will stop responding through the web interface despite the process still being active on the server. Checking the log I can see that I'm getting the following error repeating quite often:

Jun 29 15:05:22 Wekan[31666]: Exception in queued task: Error: Removed nonexistent document sdHaffBWkRddGhbct-AA8CNEENdy4xedee4
Jun 29 15:05:22 Wekan[31666]: at [object Object]._.extend.removed (packages/ddp-server/livedata_server.js:202:17)
Jun 29 15:05:22 Wekan[31666]: at [object Object]._.extend.removed (packages/ddp-server/livedata_server.js:394:10)
Jun 29 15:05:22 Wekan[31666]: at [object Object]._.extend.removed (packages/ddp-server/livedata_server.js:1281:19)
Jun 29 15:05:22 Wekan[31666]: at _cursor.observeChanges.removed (packages/cottz:publish-relations/lib/server/cursor/cursor.js:52:13)
Jun 29 15:05:22 Wekan[31666]: at packages/mongo/observe_multiplex.js:182:30
Jun 29 15:05:22 Wekan[31666]: at Array.forEach (native)
Jun 29 15:05:22 Wekan[31666]: at Function._.each._.forEach (packages/underscore.js:139:11)
Jun 29 15:05:22 Wekan[31666]: at Object.task (packages/mongo/observe_multiplex.js:176:9)
Jun 29 15:05:22 Wekan[31666]: at [object Object]._.extend._run (packages/meteor.js:807:18)
Jun 29 15:05:22 Wekan[31666]: at packages/meteor.js:785:14

and then at times the following message:
Jun 29 15:09:12 Wekan[31666]: RangeError: Maximum call stack size exceeded
or:
Jun 29 15:01:30 Wekan[31666]: Exception in setTimeout callback: RangeError: Maximum call stack size exceeded

For information the infrastructure is as follows:
Red Hat Enterprise Linux 7.3 with kernel 3.10.0-327.10.1.el7.x86_64
Wekan 0.26 installed using the manual procedure from the wiki
nodeJS v4.8.1
MongoDB 3.2.10 hosted on a seperate RHEL 7.3 instance
Wekan DB was migrated from an existing LibreBoard DB

tlmorgen:
[...]

PR #1325, opened 2017-11-06
Added people list in admin panel

thuanpq:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1325)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 tmp/
 node_modules/
 .vscode/
+.idea/
 .build/*
 packages/kadira-flow-router/
 packages/meteor-useraccounts-core/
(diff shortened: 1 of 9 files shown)


nztqa:
1. When publishing user data to the client, be extra careful not to include anything that client shouldn’t be able to see. [Always restrict fields](https://guide.meteor.com/security.html#fields)

1. Please add [pagination](http://www.discovermeteor.com/blog/template-level-subscriptions/) if that is possible. [We have over 1000 unique users](https://github.com/wekan/wekan/issues/1096#issuecomment-312187482)

thuanpq:
@nztq is just first commit, I will update soon

PR #1352, opened 2017-11-23
Revert 00a357cd7067cf5c8b055deacfe2e9bc28b06c5e to fix logging

pierreozoux:
I'm still not able to see any logs on my server.

https://12factor.net/logs recommends to just log to stdout, it has been the default for meteor, and was really happy so far.

Since this commit, I can't see logs.
Even if I see that the logger is initialized, I don't see any logs after.

I'm happy to put this again, but first, we'll need to test it a bit better.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1352)


xet7:
@huneau

It seems that your Winston logger and logging used by pierreozoux has conflicts. So your Winston logger seems to be going away if you don't get these working in harmony. I will merge this soon, just a warning.

xet7:
@huneau 

Problem is that Wekan demo keeps crashing, and it's very hard to debug without working logging.

xet7:
I don't what kind of logs you all use. I added those I know to:
https://github.com/wekan/wekan/wiki/Logs

PR #1353, opened 2017-11-24
Add disableCardRestoring & disableCardDeleting to allow only board admin to restore or remove archived cards

thuanpq:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1353)

### DIFF
--- a/client/components/settings/settingBody.jade
+++ b/client/components/settings/settingBody.jade
@@ -20,7 +20,7 @@ template(name="setting")
           if loading.get
             +spinner
           else if generalSetting.get
-            +general
+            +registration
           else if emailSetting.get
             +email
           else if accountSetting.get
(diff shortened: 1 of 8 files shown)


xet7:
@thuanpq 

As normal user I can still click List hamburger menu / More / Delete to delete card when in Admin Panel disable deleting is checked. That Delete link should not be visible when deleting is disabled.

thuanpq:
@xet7 the setting is for archived cards only, any users can delete non-archived cards if they want

xet7:
@thuanpq 

Can you explain what is actual use case where this would be useful ?

Currently this is global Wekan settings. This means that when Wekan will have organizations and teams, this setting will disable restoring and deleting archived cards for all organizations.

I would understand this better if this were per-board permission setting, but currently I'm not able to figure out the real purpose.

I would recommend that you try to implement Organizations and Teams before continuing this, because then at Admin Panel could probably be some more logical place for this setting.

PR #1362, opened 2017-11-27
Resolve "Frequent Subscriptions" problem that make Excessive CPU usage 

mfshiu:
I found the solution. The problem is "Frequent Subscriptions" in top of client\components\main\layouts.js

`Meteor.subscribe('boards');`
`Meteor.subscribe('setting');`
`Meteor.subscribe('user-admin');`

I move these code to top of client\components\boards\boardList.js

`BlazeComponent.extendComponent({`
`    onCreated() {`
`        Meteor.subscribe('boards');`
`        Meteor.subscribe('setting');`
`        Meteor.subscribe('user-admin');`
`    },`

CPU is no longer too high, from 100% down to 30% - 50%.
The subscriptions at top of layouts.js will make boards data sync every where, it's the reason that cause CPU heavy when user number is increased. To move it to creation time of boardList.js, there is the only place that boards data needed, all boards data will not sync when user enter a board or card. 
I will try to update the code or you can told me how to do that.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1362)


xet7:
Merged manualy at https://github.com/wekan/wekan/commit/7e4b42e9b3482460cac77a603809a8b5d4016477

PR #1381, opened 2017-12-06
Fixed - boards subscription should be placed at header for all other …

thuanpq:
…component can be used

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1381)


xet7:
Why? Boards subscription was moved in current place because of performance fix with large number of users https://github.com/wekan/wekan/pull/1363

xet7:
Actually, this could fix some bug you are seeing. I'll try to improve performance with upcoming other performance fix instead.

Issue #1328, opened 2017-11-10
Docs: How do/should attachments work?


richb-hanover:
**Update, and sort of solved:** See the next message to see my current thinking.

## Issue
**Impacted version**: Current (9Nov2017) Wekan on Docker from https://github.com/wekan/wekan-mongodb

**Server Setup Information**:

* Operating System: OSX 10.12.6
* Deployment Method(tar/sandstorm/snap/docker/etc /http frontend (nginx, httpd, ...)): docker image
* Node Version: default
* MongoDB Version: default
* ROOT_URL environment variable (Is there a subfolder?): localhost:8080

**Problem description**:

- It occurs to me that I would like to attach documents (images, Word or Excel) to a card, then let it percolate through the Wekan flow. I could then make cards with, say, the "Latest Projection spreadsheet" for others to study
- I don't see an obvious "front door" way to do this - it would be simple to drag a document to the card, and have it automatically added.
- I found the card's Edit Attachment, which leads to Computer/Clipboard/Drag and Drop, but don't see that the attachment has "stuck" to the card in any way that others could find the attachment.

Do attachments work like this? Am I doing it wrong? 

PS I should say that the underlying GUI of Wekan is _very_ sexy, and I like it a lot...

richb-hanover:
**Update: Never mind...** I don't know what I was doing wrong with my earlier tests. I see that Edit Attachment _does_ allow you to attach multiple attachments to a card.

That said, I still want to drag a document to a card as a shortcut to the Edit Attachment -> Computer/Clipboard/Drag&drop choice. This seems a sufficient "gesture" to indicate that I  want to add the file as an attachment.

richb-hanover:
One last thought: I was confused by "Edit attachments" menu label. If it had said, "Add attachments", I would have understood what it does immediately. (The resulting GUI only allow you to add attachments - not remove ("edit") them.)

xet7:
From @ocdtrekkie https://github.com/wekan/wekan/pull/1391

> I am not super fond of this change. It takes up like an inch and a half, since the current "Add attachment" button is a giant box. It pushes the most recent comment to nearly always be below the scroll point.
>
> Is it possible we could shorten how large this section is if there's no attachments, while still making the Attachments section visible all the time?
>
> I understand why. But as someone who rarely uses attachments in Wekan, the change has a very painful cost. I am not personally comfortable trying to improve it myself, but hoping perhaps my suggestion on how it can be improved might be useful to the person who made the PR.

[...]

PR #1391, opened 2017-12-10
Always display attachment section for uploading file quickly

thuanpq:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1391)

### DIFF
--- a/client/components/cards/cardDetails.jade
+++ b/client/components/cards/cardDetails.jade
@@ -85,13 +85,12 @@ template(name="cardDetails")
     hr
     +checklists(cardId = _id)
 
-    if attachments.count
-      hr
-      h2
-        i.fa.fa-paperclip
-        | {{_ 'attachments'}}
+    hr
+    h2
+      i.fa.fa-paperclip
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
I am not super fond of this change. It takes up like an inch and a half, since the current "Add attachment" button is a giant box. It pushes the most recent comment to nearly always be below the scroll point.

Is it possible we could shorten how large this section is if there's no attachments, while still making the Attachments section visible all the time?

xet7:
@ocdtrekkie 

Those changes have been made to make it faster to add a lot of attachments. If you have improvements to CSS, then submit pull request :)

ocdtrekkie:
I understand why. But as someone who rarely uses attachments in Wekan, the change has a very painful cost. I am not personally comfortable trying to improve it myself, but hoping perhaps my suggestion on how it can be improved might be useful to the person who made the PR.

xet7:
@ocdtrekkie 

Discussion moved to #1328

Issue #1414, opened 2018-01-03
Different icons for start date and due date


mmarschall:
Right now (version 0.63) start date and end date use the same icon in front of the date (\f017). 
It is created using time::before.

It would be great, if wekan would use a different icon for the start date (e.g. \f01d or similar). This would make it easier to see whether a date is a start date or a due date.

mmarschall:
I set up my devel environment and look at the source now.

PR #1420, opened 2018-01-04
Different icons for start date and due date

mmarschall:
Proposed fix for issue #1414 

I chose two new icons (one for start date, one for due date):
screen shot 2018-01-04 at 6 16 04 pm

An alternative approach could be:
screen shot 2018-01-04 at 6 11 51 pm

What do you think? Please let me know what I can improve in my solution.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1420)

### DIFF
--- a/client/components/cards/cardDate.js
+++ b/client/components/cards/cardDate.js
@@ -171,11 +171,12 @@ class CardStartDate extends CardDate {
   }
 
   classes() {
+    let classes = 'start-date' + ' ';
     if (this.date.get().isBefore(this.now.get(), 'minute') &&
         this.now.get().isBefore(this.data().dueAt)) {
-      return 'current';
+      classes += 'current';
     }
-    return '';
+    return classes;
   }
 
   showTitle() {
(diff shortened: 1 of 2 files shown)


davydov-vyacheslav:
IMHO, the alternative variant is better.  From end-user point of view, first case looks like something related to sign-in / sign-out

AlexanderZobkov:
I like the alternative approach as well. 
I think you can try to use consistent icons, for example:  
* use 'flag-o' as a start date.  
* use 'calendar-o' as a start date and 'calendar-check-o' as due date

Unfortunately, not much time/clock/duration related icons in font awesome.

mmarschall:
yes, true, not too much choice.

I thought about using consistent icons as well but I would prefer the icons to have different shapes to make them easier to distinguish on a glance.

I'll try all the proposals out and see what might work best.

mmarschall:
mixed icons (again):
screen shot 2018-01-04 at 6 11 51 pm


calendar icons:
screen shot 2018-01-05 at 10 02 33 pm


flag icons:
screen shot 2018-01-05 at 10 00 32 pm

and I tried it with hourglasses:
screen shot 2018-01-05 at 10 12 10 pm


I'll ask our UX designers on Monday, what they would recommend. My favorite currently is the version with the two calendar icons.

What do you think?

AlexanderZobkov:
I like 'calendar' and 'flag' options. 
Let's see what UX designers will recommend.

xet7:
Thanks! Original is most clearly understood for me.

PR #1496, opened 2018-02-21
Fix card shadow to cover the full page.

stefano-pogliani:
Hello @xet7,

I have recently updated my wekan installation to version 0.75 and noticed that the card-shadow no longer covered the page if you scroll down.
I believe this PR fixes the problem without introducing any side effects.

Hope this helps,
Thanks for Wekan!

### DIFF
--- a/client/components/boards/boardBody.styl
+++ b/client/components/boards/boardBody.styl
@@ -1,8 +1,11 @@
 @import 'nib'
 
 position()
-  if arguments[0] == cover
-    position: absolute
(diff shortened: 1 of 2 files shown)


therampagerado:
I'm on 0.76 and latest Chrome and don't have such issue. The shadow is fully visible when scrolling.

xet7:
@therampagerado 

The shadow fix is already in newest released Ubuntu Snap, and snap usually updates automatically, so if you hare using snap, you already have the fix.

Issue #552, opened 2016-03-22
Feature request: search items !


Horgix:
Hi,

When I used wekan it didn't have any search bar to look for existing items (may it be on a single board or accross multiple ones). I also don't see anything like that on the screenshot in the README : did I miss it somewhere, or would it be something that could be added ?

tiger2wander:
:+1

makufiru:
I'm working on implementing one for my local Wekan project.  How would you like it to work? Just searching descriptions or titles would be rather trivial.

tiger2wander:
I think it can be powerful with support simple syntax like in Gmail:
- Assignee
- Time: relative and exact
- Label (Color)
- Column
...

Some example:
`Label:"Feature" Assignee:"John"`
Search for cards which have label "Feature" and assigned to "John"

`Column:"On Hold" Time:"A week ago"`
Search for cards which in state "On Hold" and from a week ago till now

It can be more powerful with following improvement:
- Logical query like: "And", "Or" and parentheses group conditions
- Auto complete query like: keyword, time, user

fmonthel:
Search when you have more than 100 cards in the backlog start to be mandatory !
Search should be only on title and description (and maybe comments)
Other as label and user can be filter my2cents

ebxm:
Same issue here, hoping to see this feature come ASAP

quants-ch:
Any update on this?

xet7:
@quants-ch

First we need ideas how searching for all info on large Wekan database with lots ot cards does not slow down usage of Wekan for everyone.

One option would be to use mirroring to ToroDB and/or PostgreSQL so searching is much faster than in MongoDB:
https://github.com/wekan/wekan-postgresql

Mirroring still needs to be fixed to work with newest Wekan https://github.com/wekan/wekan/issues/787#issuecomment-312125720

rautamiekka:
^ Better not forget MySQL (or rather its superior forks Percona and MariaDB, although MariaDB is becoming something else but a MySQL fork). When everything's already running on MySQL, something not supporting it automatically becomes a non-option.

xet7:
ToroDB and ToroDB + PostgreSQL does not require any changes to Wekan.

I don't yet have info can that same be done with MariaDB etc.

If there needs to be changes, I'd like to know what is required.

I'm looking to add support for more databases, not forget any database.

xet7:
[...]

PR #1503, opened 2018-02-23
Add new search feature

GhassenRjab:
This PR adds search feature ( related to #552 )

A new search button is added inside an opened board.
On clicking on it, the search sidebar will open and will show the input where to put the search keyword
and the last added ten cards by default.
After typing the keyword, the user has to press enter to start the search.
The matched cards will appear, and by clicking on them, they will open Inside the board.
To make the search as fast as possible, the keyword is matched against the cards' titles and descriptions and only a maximum of ten matches will be shown.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1503)

### DIFF
--- a/client/components/boards/boardHeader.jade
+++ b/client/components/boards/boardHeader.jade
@@ -87,6 +87,10 @@ template(name="boardHeaderBar")
           a.board-header-btn-close.js-filter-reset(title="{{_ 'filter-clear'}}")
             i.fa.fa-times-thin
 
+      a.board-header-btn.js-open-search-view(title="{{_ 'search'}}")
+        i.fa.fa-search
+        span {{_ 'search'}}
+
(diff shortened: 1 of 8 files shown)


therampagerado:
Can we have settings where we can decide in how many cards to search? And also does it search in archived items?

xet7:
@therampagerado 

You can add new feature request as new GitHub issue, and anyone can add pull request with code that adds those features, with possibility of admin to limit results to 10 or any other number of results, and enable searching of archived cards.

But those are not part of original specs at #552 , because returning much more results would really slow down everything when there is thousands of users. This pull request implements those original specs, and for that @GhassenRjab gets all of the bounty for this search feature. @GhassenRjab already spent more time than expected with implementing this feature, it was not easy.

PR #1526, opened 2018-03-07
Fix swimlane header rotation on Google Chrome

lumatijev:
With Google Chrome `writing-mode: sideways-lr;` is marked as invalid property value and swimlane header is under swimlane menu icon.

After this change both Firefox 58 and Google Chrome 64 have properly rotated swimlane header.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1526)

### DIFF
--- a/client/components/swimlanes/swimlanes.styl
+++ b/client/components/swimlanes/swimlanes.styl
@@ -32,7 +32,8 @@
     border-bottom: 1px solid #CCC
 
     .swimlane-header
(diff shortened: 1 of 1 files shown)


GavinLilly:
This appears to still be an issue with IE11. It would seem that the CSS property is not supported in IE or Edge https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode

Issue #1483, opened 2018-02-14
Feature request : [Trello Like] Sortable checklist and items list


mlorgeoux:
## Feature request
**Impacted version**: 0.34.0

**Feature request description**:
It would be great if we can dispose a sortable checklist feature (as trello).
For the moment, a checklist can be added or removed in the create order, but it is not possible to :
- change the order of the checklists in a card
- sort the items position in a checklist without remove and recreate in the right order

xet7:
I think dragging checklists worked when @GhassenRjab added it at #876 but after @andresmanelli added swimlanes #955 there is some bug in migrations at https://github.com/wekan/wekan/blob/devel/server/migrations.js so dragging does not work anymore.

@andresmanelli @GhassenRjab 

Can you look at this?

xet7:
@mlorgeoux  

Are you running Wekan v0.34 ? Why not latest Wekan version?

mlorgeoux:
@xet7 : 
We still running on Wekan 0.34 because we did not take time to upgrade our inside platform
We will do it in the next few days !
Thanks

PR #1543, opened 2018-03-19
Checklist items sort fix, and checklist sort capabiliity

andresmanelli:
Based on #876 and #1483, I changed the model of checklist items to have their own collection. This way sorting can be implemented almost the same way as with cards/lists/swimlanes.

I also added checklist reorder capability within the cardDetails view.

Works in my local setup, please try it out!

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1543)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -131,6 +131,7 @@
     "AccountSettings": true,
     "Announcements": true,
     "Swimlanes": true,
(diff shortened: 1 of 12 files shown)


xet7:
@andresmanelli 

WOW this works so well, it's so cool ! I will merge soon, thanks a lot !

PR #1578, opened 2018-04-09
authentification oauth2

salleman33:
Hi, 

I wrote an oauth2 authentification patch for my organisation. It's not perfect : attributes may need to be adjust manually (username, fullname) to create a complete user profile with the right name, fullname, initials, etc. 
And to take off the secret configuration from file authentification.js... 
but it works ! :)

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1578)

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -31,6 +31,7 @@ kenton:accounts-sandstorm
 service-configuration@1.0.11
 useraccounts:unstyled
 useraccounts:flow-routing
+salleman:accounts-oidc
 
 # Utilities
 check@1.2.5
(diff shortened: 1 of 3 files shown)


xet7:
@salleman33 

Thanks a lot ! This is a very nice starting point for me.

I would really like if I could get more pull requests for auth like this, because it is enough that it works somehow. Standalone Wekan still has only password auth integrated, and I have not got any status info about LDAP etc.

hypery2k:
Any chance to get this merged?

xet7:
@hypery2k 

This PR needs settings added as environment variables as described in [LDAP issue](https://github.com/wekan/wekan/issues/119#issuecomment-393474106)

Issue #1652, opened 2018-05-23
Imported Boards won't open


vkb:
## Issue

**Server Setup Information**:

* Did you test in newest Wekan?: Yes
* Wekan version: 1.0.0
* If this is about old version of Wekan, what upgrade problem you have?:
* Operating System: Ubuntu Server 16.04
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): snap
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first):
* Node Version: (snap package)
* MongoDB Version: (snap package)
* ROOT_URL environment variable http(s)://(subdomain).example.com(/suburl): http://192.168.56.101:8080

**Problem description**:
Currently it is **not possible to open imported boards**. I considered problems with my entire setup, and therefore removed and freshly installed the wekan snap package, to get a new and clean environment.

*Steps to reproduce*
- Install a fresh setup of Wekan via snap
- Open the preexisting _Welcome Board_
- Add an additional testcard
- Export the board
- Re-Import the board
- Try to open the board

The board loading animation appears and nothing more happens. I added the export as an attachment:
[export_test.zip](https://github.com/wekan/wekan/files/2030228/export_test.zip)

Thank you for your work!

feuerball11:
Hi, That has probalby to do something with the CustomFields Feature, I'm Currently looking into this

vkb:
Thank you. I just noticed another detail:

URL of existing Card in preexisting _Welcome Board_:
![addressofcard](https://user-images.githubusercontent.com/458103/40414853-b09d593c-5e79-11e8-9690-c01d19a827cc.png)

URL of _Add Card_ button:
![addressofnewcard](https://user-images.githubusercontent.com/458103/40414901-d36f3b60-5e79-11e8-8011-0b9b4cd29b2c.png)

Same behavior when i create a new board. After switchting the port to 80, i can open the card. Unfortunately, it doesn't solve the loading issue of imported boards, but might represent another issue.

feuerball11:
@vkb your sekond problem sounds more like a configuration issue... anyway, to your first problem, i opend a pullrequest that should fix this, just wait for @xet7 to include this. It should fix the issue.

vkb:
Excellent, very much appreciated!

Issue #1651, opened 2018-05-22
Board view breaks after adding new card via REST interface


mariovor:
## Issue

**Server Setup Information**:

* Did you test in newest Wekan?: yes
* Wekan version: 1.0.0
* Operating System: Debian Jessie / Arch Linux
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): Docker / quay.io/wekan/wekan

**Problem description**:
Adding a new Card via the Rest interface leads a broken board view -> The board does not load anymore and gets stuck with the loading animation. However, I can see the new Card via the Rest interface, it is added to the database. 

I compared the Cards added via the web interface (a) and the rest interface (b):

a):
{
    "_id": "8QxavMczTrynh7xYT",
    "title": "New",
    "members": [],
    "labelIds": [],
    "customFields": [],
    "listId": "2QqaJ477zHvaZN7Am",
    "boardId": "eHKjiN5964Y8dNy7F",
    "sort": -1,
    "swimlaneId": "u4EEEdo59wTq5MmHK",
    "archived": false,
    "createdAt": "2018-05-22T18:42:03.033Z",
    "dateLastActivity": "2018-05-22T18:42:03.033Z",
    "isOvertime": false,
    "userId": "xxxx"
} 

b)

{
    "_id": "znJMnvr92GGJifHsv",
    "title": "New",
    "members": [
        "xxx"
    ],
    "listId": "2QqaJ477zHvaZN7Am",
    "boardId": "eHKjiN5964Y8dNy7F",
    "sort": 0,
    "swimlaneId": "u4EEEdo59wTq5MmHK",
    "archived": false,
    "createdAt": "2018-05-22T18:44:39.499Z",
    "dateLastActivity": "2018-05-22T18:44:39.499Z",
    "isOvertime": false,
    "swimlaneId": "u4EEEdo59wTq5MmHK",
    "userId": "xxxx"
}

We see, in b) are missing the labelIds and customFields. Without checking the Wekan code in detail, I could believe that could lead to problems.
As far as I can see in https://github.com/wekan/wekan/blob/devel/models/cards.js both elements are not added in 
`JsonRoutes.add('POST', '/api/boards/:boardId/lists/:listId/cards', function (req, res).` 
Would an expert agree with this analysis?

It requires a restore of the mongoDB in order to make Wekan work again.

Let me know if I can provide more information.

If nobody cares to much about the issue, I could try to fix it. But it could take a bit.

[...]

Issue #807, opened 2017-02-08
Add feature: Custom Fields


xet7:
_From @xet7 on January 10, 2017 16:50_

Similar to Trello feature:
http://blog.trello.com/trello-custom-fields

Existing discussion at original Wekan forum:
https://discuss.wekan.io/t/any-way-to-add-new-data-on-cards/362/2

This wish is from original Wekan Gitter user @TutloTutlo

I did not find any existing pull request for this yet from original Wekan.

Wefork already has Due Date, so this new feature Custom Fields is separate from that.

If somebody has more information about this, please add it so it's ready here when somebody starts implementing this.

This issue is related to #43 and making Wekan programmable.

_Copied from original issue: wefork/wekan#61_

xet7:
_From @ehausen on January 24, 2017 21:14_

In Trello custom fields are kind of a hack and only available per board. I'm thinking a better logic would be to define **Card Types** - like content types! This would very handy: _Project_ card type, a _Contact_ card type (for CRM purposes), etc...

Looking at Trello... Please someone (or me later) continue

- User can define custom field types
  - Field Name
  - Field type (text, number, checkbox, date, dropdown, location? card reference? list reference? link? )
  - Is it displayed on the Card Front Badge when card not open

- If fields are set they are listed at the top of the card

- Fields should integrate with search and filters (not in trello)

xet7:
_From @tugal on January 25, 2017 15:0_

Hi

i agree @ehausen, card types would be really interesting, letting every users to customize their card..

xet7:
_From @ehausen on January 25, 2017 15:4_

Well I was actually thinking more along the lines of defining these at team or organization level - when there will be implemented. Probably the best would be to have permission where the admin can define which roles can do what.

tugal:
Hi,

I'm trying to develop it.

**Design** 
Add new menu entry (I'm actually planning in provide a list of available custom field names)

<img width="358" alt="capture d ecran 2017-02-09 a 10 17 48" src="https://cloud.githubusercontent.com/assets/556990/22776719/0f2304be-eeb1-11e6-972f-926eb4c12fb4.png">

<img width="304" alt="capture d ecran 2017-02-09 a 10 18 27" src="https://cloud.githubusercontent.com/assets/556990/22776730/1ffb6600-eeb1-11e6-8983-44fe42393260.png">

Then, display it (quick UI only to POC)

[...]

PR #1653, opened 2018-05-23
Custom Fields Correction of Loading Board Issues

feuerball11:
This commit should fix a lot of issues regarding not opening boards after import / REST API:

#1652 
#1651 (Not sure, but could be, would nice if the issue author could test this after release) 
#807

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1653)


feuerball11:
Srsly? Indends again?

xet7:
I did leave codacy as it is. Having `npm run lint` without errors is more important anyway.

Issue #833, opened 2017-02-11
Add Feature: Delete Wekan board


HLFH:
Hi,

We should be able to delete Wekan board.
Archive feature is not enough.

Thanks,

HLFH:
Offtopic: can you add the milestone v0.12 @xet7 to Wekan and link https://github.com/wekan/wekan/issues/785 and https://github.com/wekan/wekan/issues/784? Thanks!

fmonthel:
Hi,

I developed below script to clean database (so here will delete board archived fore more than 1 month) : https://github.com/fmonthel/wekan-cleanup

xet7:
@fmonthel 

Does your script delete archived card attachments?

fmonthel:
Arf not, but I will add (as checklist)

xet7:
@fmonthel 

Thanks! I added link to your cleanup script to readme and changelog.

fmonthel:
wekan-cleanup fix to delete (at least into cfs.attachments.filerecord) upload data. Not sure how to proceed with other cfs.* collection , I need to look deeper
wekan-cleanup fix as well to delete checklists

Script is deleting as well board(s) with no cards into and created more than x days ago

xet7:
Issues related to Wekan cleanup Python script can be continued at:
https://github.com/wekan/wekan-cleanup

This issue is about deleting board in Wekan.

genecraig:
2 years and we cannot delete boards? ABSURD!

ocdtrekkie:
@genecraig This is an open source project. Nobody makes any money to work on it. The fact that you have the ability to create Wekan boards for free is something you should be appreciative of. If you'd like to contribute, do so. If you'd like to offer up money on here for someone to write a feature you want, someone might be interested in taking you up on that even.

But coming in to comment on the "absurdity" that a feature you want isn't in a program that you didn't pay for or contribute to in any way is, in fact, absurd. :)

genecraig:
@ocdtrekkie definition of trolling is to follow someone around online commenting on their comments. This is what you have just done from one thread to another. STOP!

And this is NOT a charity. You are using the public for free QA. We are doing part of your work for your projects and your compensation is allowing us to use the product in its half-baked state. If you don't want criticism then why did you put your projects on GitHub? It is NOT charity. You are so ungrateful for the work the public does on YOUR intellectual property without any payment.

ocdtrekkie:
[...]

PR #1679, opened 2018-06-05
Additional label colours & fields

rjevnikar:
Add additional label colours (from pull by JamesLavin) and add Assigned & Requested By text fields.
Add delete board feature (#833)

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1679)


xet7:
@rjevnikar 

Thanks a lot for your amazing work!

It would be much easier if you could create separate pull request for each feature, because some of these features are ready, and some are not. That way I could test and merge them separately. Having them together in one pull request means that all features need to be ready before merging, and it's harder to test in what feature each bug is.

1) Additional label colors: Seems to work, I'll need to test with export/import later.

2) Board Delete: Seems to work, but it should be in All Boards / Recycle Bin - at right side of Restore Board buttons. delete-board button seems to not be translated, there is still text delete-board when Wekan is running, instead of text Delete Board.

3) Assigned & Requested By text fields: add buttons does not work, when I have existing Wekan data and added your feature, maybe [migration is missing](https://github.com/wekan/wekan/issues/807#issuecomment-390385631).

Wekan card originally looks like this:
before

After adding Assigned & Requested By text fields it looks like this:
after

So layout is broken.

xet7:
After migration added, Reguested By / Assigner By Add-buttons don't work yet.

xet7:
@rjevnikar 

Thanks, you have great progress!

Now in browser right client / Inspect Element / Console, there is these error about multiple templates:
multiple-templates

xet7:
@rjevnikar

Thanks, great progress!

- Card layout is now fixed
- You did add Delete Board button to All Boards / Recycle Bin, but it should be on right side, like this: `Board name [Restore Board] [Delete Board]`, and Delete Board button does not work yet.
- Assigned By / Received By Add-buttons do not work yet
- Delete Board button is not yet removed from single Board hamburger menu

rjevnikar:
Thanks for your patience @xet7 .
I think I have addressed everything, but now Travis CI build has failed.
I ran eslint using npm, and didn't get any errors.

xet7:
@rjevnikar 

That Travis build error is my fault, I copied newest Sandstorm fork version of Node.js from Sandstorm to Wekan server, and forgot to update checksum. I will fix it.

Issue #1693, opened 2018-06-13
1.04 - after update to this version the yellow due date warning is missing.


therampagerado:
## Issue

**Server Setup Information**:

* Did you test in newest Wekan?: currently using
* Wekan version: 1.04
* If this is about old version of Wekan, what upgrade problem you have?:
* Operating System: ubuntu
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): snap
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): no
* Node Version: 8.11.1
* MongoDB Version: latest
* ROOT_URL environment variable http(s)://(subdomain).example.com(/suburl): no

**Problem description**:
- when I came to work today the kanban has updated to the latest as usual and the yellow due date warning is missing. They stay grey as those far in time. The red warning is here.

What can have caused this issue?

xet7:
@therampagerado 

Maybe it's this in progress work https://github.com/wekan/wekan/pull/1685 by @rjevnikar 
I presume he will do more changes later.

therampagerado:
OK but I can't see why you have to implement a broken fix when it's WIP... Please can you revert this in 1.05 and include it again when he's ready?

xet7:
@therampagerado 

You can add pull request to change it.

BTW, Maybe I wrote wrong, @rjevnikar has name Rebecca in GitHub name, so could be she.

rjevnikar:
Do you have an end date set?

xet7:
@therampagerado 

Please read #1668 about what is currently in progress to be fixed.

therampagerado:
@rjevnikar  - no I don't use end dates because I don't need them.

Until now we used only due date which was highlighting the more urgent orders nicely (sometimes strangely but ...). Now the ones with passed dates are red (as they were before) but those due 2 or less dates are no longer yellow, they stay grey and then turn red when the time passes.

rjevnikar:
ok, that is because of the logic
`if ((this.now.get().diff(this.date.get(), 'days') >= 2) &&
       (this.date.get().isBefore(this.data().endAt)))
      classes += 'long-overdue';
    else if ((this.now.get().diff(this.date.get(), 'minute') >= 0) &&
       (this.date.get().isBefore(this.data().endAt)))
      classes += 'due';
    else if ((this.now.get().diff(this.date.get(), 'days') >= -1) &&
       (this.date.get().isBefore(this.data().endAt)))
[...]

PR #1698, opened 2018-06-13
Fix data colour changes on cards.

rjevnikar:
Addressing both #1685 and #1693

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1698)


xet7:
@rjevnikar 

Maybe in some other pull request you can leave out translation files from wekan/i18n/ directory? Because usually only changes to wekan/i18n/en.i18n.json is needed in pull request, when new strings are added. You don't need to change this pull request though, I'll merge it soon.

PR #1701, opened 2018-06-14
Repair of CardDetail for Mobile View (Hotfix)

feuerball11:
This should fix cardDetail on Mobile View. 
A thing that bugged me for a while, becouse it realy kills mobile usage. 

Please note: Its only a hotfix.


jetnet:
Thanks a lot for the fix! It works now, but there is a one small issue - when you open a card, you have to use **two** fingers to scroll, one finger scroll does not work on Android/Chrome.

xet7:
@jetnet 

I don't know how that can be fixed. One finger is used for clicking etc. You can add new issue, because here at merged pull request it gets lost.

PR #1710, opened 2018-06-19
Fixed advanced filter description.

zypA13510:
The syntax in advanced filter description is wrong (assignment operator = instead of equality comparison operator ==).

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1710)


zypA13510:
@xet7 Someone please review this?

xet7:
@zypA13510 

I need to add these changes manually to every translation myself sometime, because otherwise every translator needs to translate this again.

zypA13510:
@xet7 then should I just change = -> == in every language?

xet7:
@zypA13510 

It's better that I do it, because only I can change base english language etc.

PR #1717, opened 2018-06-19
Submit inline form on click outside

pravdomil:
make sense?

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1717)

### DIFF
--- a/client/lib/inlinedform.js
+++ b/client/lib/inlinedform.js
@@ -81,10 +81,10 @@ EscapeActions.register('inlinedForm',
 
 // submit on click outside
 document.addEventListener("click", function(evt) {
-  const formIsOpen = currentlyOpenedForm.get() && currentlyOpenedForm.get().isOpen.get();
+  const openedForm = currentlyOpenedForm.get()
   const isClickOutside = $(evt.target).closest(".js-inlined-form").length === 0;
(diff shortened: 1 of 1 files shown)


zypA13510:
I think the argument for both behaviors is equally true:
On one hand, I've seen many new users trying to save by clicking outside, so it is reasonable to make clicking outside to save. OTOH, if the user expects it to save, when it is not, the user can always re-enter what they have just entered. But if the user expects it to discard, when it is actually saved, the old data is lost forever and there will be no way to recover. (It is much more likely to remember the new input than the old data)

I think the current behavior in "description" field is fine. It saves the input into a temporary "unsaved description" and asks the user whether they would like to discard it.

pravdomil:
gif

finder macos

pravdomil:
gif

finder mac os

zypA13510:
@pravdomil But I'm pretty sure you can find plenty of web apps these days that do the opposite.
And in the renaming example, you can always undo. But in Wekan you cannot (if it saves when you intended to discard).

pravdomil:
> I think the current behavior in "description" field is fine. It saves the input into a temporary "unsaved description" and asks the user whether they would like to discard it.

this PR is especially for edit `Cart Title` in `Board View` and `Card Detail`

PR #1592, opened 2018-04-18
Board-As-Card / Card-As-Card

andresmanelli:
This PR hopefully implements the import feature of a board or a card (from another board) to an existing board.

It aims to fix #492 and maybe #709 and #1068.

What is implemented:

- Two way data binding for:
  - Checklists: from original card, or own checklists if imported board
  - Description: from original card/board
  - Member assignment: for card _and_ board
  - Comments
  - Activities
  - Attachments
  - Dates / spentTime

What is missing:
  - ~Add link to imported card / board to open that card / board~
  - Bugfixes when necessary :D

### Screenshot
seleccion_002

@xet7 Please try this and let me know if I can add anything else.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1592)

Issue #1430, opened 2018-01-15
Trello Import // Board not Found


generic-user-name:
## Exporting Data from grain that has Board not Found

https://github.com/wekan/wekan/wiki/Export-from-Wekan-Sandstorm-grain-.zip-file

## Issue
**Impacted version**: 0.63.0

**Server Setup Information**:

* Operating System:  Ubuntu 14.04.5 LTS
* Node Version: v4.8.4
* MongoDB Version:2.6.7

**Problem description**:
- Encounterd a problem with the Import of a board from Trello
- At first it looks like its working, and everything gets imported as expected
- after I close the board and try to reopen it again I get the Message "Board not Found"

xet7:
@generic-user-name 

How did you install Wekan? I have not seen this error before.

generic-user-name:
Hi,

thanks for the quick reply,

WeKan is currently running under sandstorm.io

xet7:
@generic-user-name 

Please add Wekan grain debug log, that you see with terminal/display looking button.

generic-user-name:
...** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns
** Starting Mongo...
2018-01-15T16:38:53.365+0000 I STORAGE  Engine custom option: log=(prealloc=false,file_max=200KB)
about to fork child process, waiting until server is ready for connections.
forked process: 10
child process started successfully, parent exiting
** Starting Meteor...
> Starting board-background-color migration.
> Finishing board-background-color migration.
> Starting lowercase-board-permission migration.
> Finishing lowercase-board-permission migration.
> Starting change-attachments-type-for-non-images migration.
> Finishing change-attachments-type-for-non-images migration.
> Starting card-covers migration.
> Finishing card-covers migration.
> Starting use-css-class-for-boards-colors migration.
> Finishing use-css-class-for-boards-colors migration.
> Starting denormalize-star-number-per-board migration.
> Finishing denormalize-star-number-per-board migration.
> Starting add-member-isactive-field migration.
> Finishing add-member-isactive-field migration.
> Starting add-sort-checklists migration.
> Finishing add-sort-checklists migration.
> Starting enforce-public-visibility-for-sandstorm migration.
> Finishing enforce-public-visibility-for-sandstorm migration.
** SANDSTORM SUPERVISOR: Grain shutdown requested.
** SANDSTORM SUPERVISOR: Starting up grain. Sandbox type: userns
** Starting Mongo...
[...]

PR #1723, opened 2018-06-24
Nested tasks

TNick:
Adds the ability to add cards inside other cards. Newly added card is inserted into another board.

The implementation modifies the database as follows:

- **Cards**: adds a field for parent card (_parentId_) and a sorting member for the order of cards inside parent card (_subtaskSort_);
- **Boards**: 
    - each board has another board where it deposits subtasks added to cards inside it (_subtasksDefaultBoardId_). One is created if none exists.
    - any board can be the destination for subtasks from any board; inside it there is a default list that receives incoming cards (_subtasksDefaultListId_); it will be created it none is set.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1723)


client/components/cards/checklists.jade
```diff
@@ -75,7 +74,7 @@ template(name="checklistItems")
       +inlinedForm(classNames="js-edit-checklist-item" item = item checklist = checklist)
         +editChecklistItemForm(type = 'item' item = item checklist = checklist)
       else
-        +itemDetail(item = item checklist = checklist)
+        +cjecklistItemDetail(item = item checklist = checklist)
```
xet7: is here typo? cjecklistItemDetail ?

client/components/cards/checklists.jade
```diff
@@ -84,7 +83,7 @@ template(name="checklistItems")
           i.fa.fa-plus
           | {{_ 'add-checklist-item'}}...
 
-template(name='itemDetail')
+template(name='cjecklistItemDetail')
```
xet7: is here typo? cjecklistItemDetail ?

client/components/cards/checklists.js
```diff
@@ -204,7 +204,7 @@ Template.checklistDeleteDialog.onDestroyed(() => {
   $cardDetails.animate( { scrollTop: this.scrollState.position });
 });
 
-Template.itemDetail.helpers({
+Template.cjecklistItemDetail.helpers({
```
xet7: is here typo? cjecklistItemDetail ?

client/components/cards/checklists.js
```diff
@@ -223,4 +223,4 @@ BlazeComponent.extendComponent({
       'click .js-checklist-item .check-box': this.toggleItem,
     }];
   },
-}).register('itemDetail');
+}).register('cjecklistItemDetail');
```
xet7: is here typo? cjecklistItemDetail ?

i18n/es.i18n.json
```diff
@@ -247,7 +247,7 @@
...
+    "advanced-filter-description": "El filtrado avanzado permite escribir una cadena que contiene los siguientes operadores: == != <= >= && || ( ) Se utiliza un espacio como separador entre los operadores. Se pueden filtrar todos los campos personalizados escribiendo sus nombres y valores. Por ejemplo: Campo1 == Valor1. Nota: Si los campos o valores contienen espacios, deben encapsularse entre comillas simples. Por ejemplo: 'Campo 1' == 'Valor 1'. Para omitir los caracteres de control único (' \\/), se usa \\. Por ejemplo: Campo1 = I\\'m. También se pueden combinar múltiples condiciones. Por ejemplo: C1 == V1 || C1 == V2. Normalmente todos los operadores se interpretan de izquierda a derecha. Se puede cambiar el orden colocando paréntesis. Por ejemplo: C1 == V1 && ( C2 == V2 || C2 == V3 ). También se puede buscar en campos de texto usando expresiones regulares: C1 == /Tes.*/i",
```
xet7: Only i18n/en.i18n.json has additions at GitHub, others are translated at https://transifex.com/wekan/wekan . Please remove other translation files. Here's how to remove commits:
https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-pull-request

i18n/fr.i18n.json
```diff
@@ -247,7 +247,7 @@
...
+    "advanced-filter-description": "Le filtre avancé permet d'écrire une chaîne contenant les opérateur suivants : == != <= >= && || ( ). Les opérateurs doivent être séparés par des espaces. Vous pouvez filtrer tous les champs personnalisés en saisissant leur nom et leur valeur. Par exemple : champ1 == valeur1. Remarque : si des champs ou valeurs contiennent des espaces, vous devez les mettre entre apostrophes. Par exemple : 'champ 1' = 'valeur 1'. Pour échapper un caractère de contrôle (' \\/), vous pouvez utiliser \\. Par exemple : champ1 = I\\'m. Il est également possible de combiner plusieurs conditions. Par exemple : f1 == v1 || f2 == v2. Normalement, tous les opérateurs sont interprétés de gauche à droite. Vous pouvez changer l'ordre à l'aide de parenthèses. Par exemple : f1 == v1 and (f2 == v2 || f2 == v3). Vous pouvez également chercher parmi les champs texte en utilisant des expressions régulières : f1 == /Test.*/i",
```
xet7: Only i18n/en.i18n.json has additions at GitHub, others are translated at https://transifex.com/wekan/wekan . Please remove other translation files. Here's how to remove commits:
https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-pull-request

i18n/ja.i18n.json
```diff
@@ -2,18 +2,18 @@
...
     "act-addAttachment": "__card__ に __attachment__ を添付しました",
-    "act-addChecklist": "added checklist __checklist__ to __card__",
-    "act-addChecklistItem": "added __checklistItem__ to checklist __checklist__ on __card__",
+    "act-addChecklist": "__card__ に __checklist__ を追加しました",
+    "act-addChecklistItem": "__checklist__ on __card__ に __checklistItem__ を追加しました",
     "act-addComment": "__card__: __comment__ をコメントしました",
     "act-createBoard": "__board__ を作成しました",
     "act-createCard": "__list__ に __card__ を追加しました",
-    "act-createCustomField": "created custom field __customField__",
+    "act-createCustomField": "__customField__ を作成しました",
```
xet7: Only i18n/en.i18n.json has additions at GitHub, others are translated at https://transifex.com/wekan/wekan . Please remove other translation files. Here's how to remove commits:
https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-pull-request

i18n/ja.i18n.json
```diff
@@ -2,18 +2,18 @@
...
     "act-createList": "__board__ に __list__ を追加しました",
     "act-addBoardMember": "__board__ に __member__ を追加しました",
-    "act-archivedBoard": "__board__ moved to Recycle Bin",
-    "act-archivedCard": "__card__ moved to Recycle Bin",
-    "act-archivedList": "__list__ moved to Recycle Bin",
-    "act-archivedSwimlane": "__swimlane__ moved to Recycle Bin",
+    "act-archivedBoard": "__board__ をゴミ箱に移動しました",
+    "act-archivedCard": "__card__ をゴミ箱に移動しました",
+    "act-archivedList": "__list__ をゴミ箱に移動しました",
+    "act-archivedSwimlane": "__swimlane__ をゴミ箱に移動しました",
```
xet7: Only i18n/en.i18n.json has additions at GitHub, others are translated at https://transifex.com/wekan/wekan . Please remove other translation files. Here's how to remove commits:
https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-pull-request

i18n/ru.i18n.json
```diff
@@ -247,7 +247,7 @@
...
+    "advanced-filter-description": "Расширенный фильтр позволяет написать строку, содержащую следующие операторы: ==! = <=> = && || () Пространство используется как разделитель между Операторами. Вы можете фильтровать все пользовательские поля, введя их имена и значения. Например: Field1 == Value1. Примечание. Если поля или значения содержат пробелы, вам необходимо взять их в кавычки. Например: «Поле 1» == «Значение 1». Для одиночных управляющих символов ('\\ /), которые нужно пропустить, вы можете использовать \\. Например: Field1 = I \\ m. Также вы можете комбинировать несколько условий. Например: F1 == V1 || F1 = V2. Обычно все операторы интерпретируются слева направо. Вы можете изменить порядок, разместив скобки. Например: F1 == V1 && (F2 == V2 || F2 == V3). Также вы можете искать текстовые поля с помощью regex: F1 == /Tes.*/i",
```
xet7: Only i18n/en.i18n.json has additions at GitHub, others are translated at https://transifex.com/wekan/wekan . Please remove other translation files. Here's how to remove commits:
https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-pull-request

models/cards.js
```diff
@@ -271,14 +318,33 @@ Cards.helpers({
...
 
 Cards.mutations({
+  applyToKids(funct) {
+    Cards.find({ parentId: this._id }).forEach((card) => {
+      funct(card);
+    });
+  },
+
   archive() {
+    this.applyToKids((card) => { return card.archive(); });
```
xet7: Please change all mentions of Kids => Children

i18n/ru.i18n.json
TNick: I think that one of your commits got caught up in the process (09c8f29)

xet7: @TNick 

Ok, those are my changes, nevermind.

TNick: @xet7 too late now... :) 
If it is Ok I can drop this branch and squash it into a single commit

xet7: @TNick 

I prefer separate commits with descriptions, like this is currently. I do not like squashing.

xet7:
@TNick 

Oh no, sorry, I think this got messed up with some commits by somebody from https://github.com/wekan.py Python port. You should try to squash your commits and remove all the extra stuff, or add a new pull request.

xet7:
@TNick

Sure you can also remove extra stuff directly in this pull request too, it's fine :+1:

xet7:
@TNick 

Please fix merge conflicts. Thanks!

d3dbit:
@TNick , @xet7 ,
How can impact this to the Wekan app upgrade process? 
__"The implementation modifies the database"__
And how can impact it to a Wekan app restore from a JSON backup or mongo-dump backup (a backup from an older Wekan version)?
Thanks.

xet7:
@d3dbit 

For migrating old to new Wekan database structure, [migrations](https://github.com/wekan/wekan/blob/devel/server/migrations.js) needs to be added, so database schema is upgraded.

Mongorestore and upgrading is not impacted, if migrations code exists.

Trello import is impacted, if Trello JSON includes nested tasks. Those need to be converted to Wekan nested tasks.

Wekan JSON import/export needs to have nested tasks imported/exported, or ignored when importing older Wekan JSON.

Database schema and import/export code is here:
https://github.com/wekan/wekan/tree/devel/models

With this implementation of nested boards, if I understand correctly, it duplicates all boards so they have subboards containing all nested tasks. For example, if Wekan has 100 boards, it would be duplicated to 200 boards. I don't know how this would be visible to users, because I have not tested this yet, because of merge errors.

xet7:
I have also not looked how this would compare to Linked Cards and Linked boards https://github.com/wekan/wekan/pull/1592 , is there some overlap in funtionality, and which one would be better as design to base these features on.

TNick:
To the point that 100 boards need 100 sub-boards: this *can* be the case
but is not a condition. Any board (including itself) can be default
destination for current board. So you can have 3 main boards and one board
which receives subtasks from all 3.

Also worth pointing out that subtasks of cards in a board can exist in
various boards including current one. Not all subtasks in a board are in
same "sub-board". There is a *default* board where new sub-tasks are
deposited.

I did not evaluate the impact over sandstorm either, btw.

TNick:
### Wekan app restore from a JSON backup
Current implementation does nothing about this. It should preserve the "vertical" links between cards implemented in this pull in a manner consistent with "lateral" links implemented by #1592 

Things to store-retrieve: 

- cards: 
  - link to parent card
  - sort number for subtasks
- boards
  - allow subtasks or not
  - default board for subtasks of cards in this board
  - default list for incoming subtasks (when this list is a default board for another board)
  - the way parent task of cards in this boards is presented

The change in export.js is actually a **_bug_**, as we should only export the id of the card for cards in current board. Also, a full export should only have ids.

### Trello Import

I haven't looked at the code in there and is unlikely that I will have time to do so.

### Wekan Import

This is something that needs implemented. Should be easy so will probably add a commit this week.
One nasty thing is importing a board with subtasks in another board.
Is it OK to create boards when importing boards? #sandstorm

### Sub-boards and Sandstorm

With regards to @xet7 comment about sub-boards: I forgot to mention that - if a default board is not set - when the user adds a subtasks a new board is created. This is fine in plain implementation but sandstorm only works with one board. Possible solutions:

- in sandstorm the board is always its own default board => cards will be deposited in same board
- I don't think that there is anything preventing the database in sandstorm to have multiple boards, it's just a gui problem. So create new board to be used as default board for new tasks in main board and add a way to switch between main tasks and subtasks. Either:
    - limit the nesting level to 1 (subtasks cannot have their own subtasks) or 
    - do something smarter like allowing multiple boards (significant change, probably). 

### Linked cards #1592 

I suggest changing `type` to `linkType` in models/cards.js
Once implemented code needs to be added to take subtasks into consideration as linked data.
Other than that I see no reason for concern.

xet7:
At Sandstorm only one board is visible, with board name sandstorm. Importing has [board not found bug](https://github.com/wekan/wekan/issues/1430). Each board is in separate grain, that increases security. I could enable seeing all boards in Sandstorm grain, although that would alter current user experience a lot, then probably all those not found boards could be seen. Getting Sandstorm bugs fixes would require a lot of time, because a lot of it is in Sandstorm internals etc. Most of my current customer work is for Standalone Wekan.

sfahrenholz:
Hello together,

i hope my question is ok, is there any chance, that this feature comes in the next release? This were really nice. Thanks!

TNick:
Right now I see a strange bug: parent is not shown if you do not click first on the board where the parent belongs. This means that I fail to understand some internal logic.

One thing that I want to add is ability to show archived tasks as checked, with those not archived being unchecked.

FYI I implemented this feature because I need it for my business, I use it daily and looks good, except previously mentioned quirks.

d3dbit:
@TNick 
Thanks for your work.

Samunosuke:
Thanks for the great work.
I tried the new sub task feature and the first thing which did miss was to see the name of the list from the subtasks.
wekansubtaskfeature

This would allow to show the progress from the subtask without opening the "sub board" where the subtask is located.
E.g. 
MySubTask 1       Delete/View
>>In implementation
MySubTask 2       Delete/View
>>Finished

zypA13510:
+1 for the great work @TNick.

However, I tried the feature, there is a potentially confusing issue:
If no board is selected to keep the subtask under _subtasks settings_, which is the default, unlinking the subtask with its parent creates unexpected result - if unlinked, the subtask would simply disappear from the board, and all activities related to it become "undefined".

By linking, I mean to **change card's parent** of an existing task, like this:
capture

To solve this issue, the user must manually select a board instead of _(none)_ under **deposit subtasks to this board**, all disappeared cards would re-appear in the board selected.
capture1

P.S. I believe it would be better to use "save" instead of "deposit" (in the second image), just my personal opinion tho.

TNick:
Thanks for testing & reporting.
Will look into that. The two parts should be independent; "deposit subtasks" regards only future actions and should change nothing about existing cards.

I have no preference of "deposit" over "save". If others don't have any issue with that I suggest a PR.

gkarachuk:
Subtasks are a great feature, thanks for all the work.
Does an icon like the checklist icon/numbers that appear on the cards exist for subtasks?  I could not find it in any settings, and it is hard to keep track of which cards have subtasks.

I am hoping to use the subtasks to create a milestones board with each card having subtasks to be completed to complete the milestones.  I think I prefer this because the subtask cards are "active" while the checklist items are a bit hidden and "passive".

Thanks for the great work.

Issue #808, opened 2017-02-08
Add feature: Calendar View


xet7:
_From @xet7 on January 10, 2017 17:8_

Wish originally from @TutloTutlo

It's this Trello feature:
http://help.trello.com/article/811-viewing-cards-in-a-calendar-view
http://blog.trello.com/enable-trello-calendar-mode

There's also possibility use Trello with Google Calendar:
http://help.trello.com/article/829-using-trello-with-google-calendar

This will include CalDav support.
Resource: [Introducing CalDav](http://blogs.nologin.es/rickyepoderi/index.php?/archives/14-Introducing-CalDAV-Part-I.html)

In Sandstorm Wekan could work with some Sandstorm calendar app instead.

_Copied from original issue: wefork/wekan#62_

phillc73:
Related to #424 

I think CalDAV support would provide the most flexibility. 

Resource: [Introducing CalDav](http://blogs.nologin.es/rickyepoderi/index.php?/archives/14-Introducing-CalDAV-Part-I.html)

NoodleBB:
Calendar view is missing. :+1:

steven-tib:
+1 for the calendar view! I just discovered it on trello and would be awesome to have it in Wekan (making it even greater!!!) 
It would open new perspectives for the use of wekan in two current projects.

NoodleBB:
> It would open new perspectives for the use...
It's the main reason, why i can't use wekan in reallife. 😞

pmatts:
Easier than webdav would be to maintain an ics file.  This could then be subscribed to from a multitude of clients (including android via icsdroid).

This would be the most useful implementation for me anyways.

Creanimo:
+1
I urgently need a calendar view as well. I think a calendar (especially with drag and drop) could turn wekan into a very effective project management software.

I am quite new to github and I saw that sometimes people put a bounty on implementing features. Is this a possibility here?

xet7:
@Creanimo 

Yes it's possible to put bounty:
https://www.bountysource.com/teams/wekan

xet7:
@Creanimo

Yes it's possible to put bounty:
https://www.bountysource.com/teams/wekanteam

xet7:
There's some setup of bountysource still going on, so please wait until I have configured it correctly.

xet7:
https://github.com/wekan/wekan/wiki/FAQ#will-my-feature-request-be-implemented

killua-eu:
[...]

PR #1728, opened 2018-06-26
Calendar

TNick:
Adds a calendar view.

This is work in progress. Please comment and contribute.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1728)

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -77,7 +77,6 @@ email@1.2.3
 horka:swipebox
 dynamic-import@0.2.0
 staringatlights:fast-render
-staringatlights:flow-router
 
 mixmax:smart-disconnect
 accounts-password@1.5.0
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -84,3 +84,4 @@ accounts-password@1.5.0
 cfs:gridfs
 browser-policy
 eluck:accounts-lockout
+rzymek:fullcalendar
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -85,3 +85,4 @@ cfs:gridfs
 browser-policy
 eluck:accounts-lockout
 rzymek:fullcalendar
+momentjs:moment@2.22.2
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -103,6 +103,7 @@ mixmax:smart-disconnect@0.0.4
 mobile-status-bar@1.0.14
 modules@0.11.0
 modules-runtime@0.9.1
+momentjs:moment@2.8.4
 mongo@1.3.1
 mongo-dev-server@1.1.0
 mongo-id@1.0.6
@@ -139,6 +140,7 @@ reactive-var@1.0.11
 reload@1.1.11
 retry@1.0.9
 routepolicy@1.0.12
+rzymek:fullcalendar@3.8.0
 service-configuration@1.0.11
 session@1.1.7
 sha@1.0.9
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -157,7 +157,6 @@ srp@1.0.10
 standard-minifier-css@1.3.5
 standard-minifier-js@2.2.3
 staringatlights:fast-render@2.16.5
-staringatlights:flow-router@2.12.2
 staringatlights:inject-data@2.0.5
 stylus@2.513.13
 tap:i18n@1.8.2
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -103,7 +103,7 @@ mixmax:smart-disconnect@0.0.4
 mobile-status-bar@1.0.14
 modules@0.11.0
 modules-runtime@0.9.1
-momentjs:moment@2.8.4
+momentjs:moment@2.22.2
 mongo@1.3.1
 mongo-dev-server@1.1.0
 mongo-id@1.0.6
--- a/client/components/boards/boardBody.jade
+++ b/client/components/boards/boardBody.jade
@@ -25,3 +25,5 @@ template(name="boardBody")
           +swimlane(this)
       if isViewLists
         +listsGroup
+      if isViewCalendar
+        +fullcalendar(calendarOptions)
--- a/client/components/boards/boardBody.js
+++ b/client/components/boards/boardBody.js
@@ -114,6 +114,62 @@ BlazeComponent.extendComponent({
     }
   },
 
+  calendarOptions() {
+    return {
+      id: 'calendar-view',
+      defaultView: 'basicWeek',
+      header: {
+        left: 'title',
+        center: 'agendaDay,listDay,timelineDay agendaWeek,listWeek,timelineWeek month,timelineMonth timelineYear',
+        right: 'today prev,next',
+      },
+      views: {
+        basic: {
+          // options apply to basicWeek and basicDay views
+        },
+        agenda: {
+          // options apply to agendaWeek and agendaDay views
+        },
+        week: {
+          // options apply to basicWeek and agendaWeek views
+        },
+        day: {
+          // options apply to basicDay and agendaDay views
+        },
+      },
+      themeSystem: 'jquery-ui',
+      height: 'parent',
+      /* TODO: lists as resources: https://fullcalendar.io/docs/vertical-resource-view */
+      navLinks: true,
+      nowIndicator: true,
+      businessHours: {
+        // days of week. an array of zero-based day of week integers (0=Sunday)
+        dow: [ 1, 2, 3, 4, 5 ], // Monday - Thursday
+        start: '8:00',
+        end: '18:00',
+      },
+      locale: TAPi18n.getLanguage(),
+      events(start, end, timezone, callback) {
+        const currentBoard = Boards.findOne(Session.get('currentBoard'));
+        const events = [];
+        currentBoard.cardsInInterval(start.toDate(), end.toDate()).forEach(function(card){
+          events.push({
+            id: card.id,
+            title: card.title,
+            start: card.startAt,
+            end: card.endAt,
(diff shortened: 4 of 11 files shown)


xet7:
@TNick 

Does this show card's Received/Start/Due/End dates in calendar?

xet7:
@TNick 

I'll try it when I have some free time. Thanks!

xet7:
Calendar issue at #808 . It also has bounties.

TNick:
@xet7 uses start and end to display the card as an interval and shows the title of the card at this point. 

A link to the card is inserted but does nothing because the code does not know to switch to another view.

An easy next step is type use the card detail template to show cardDetail in a popup when the user clicks a card.

Then changing default view to our minicard would be nice.

Cards that only have start or end should be presented at that time with a 1 hour duration.

Cards that have no time info should be presented in a separate side list so that the user can drag them into place.

TNick:
Also, it would be possible to show same card more than once:

- once if there is a received time with a distinct color
- once the start-end interval
- finally if there is a due date => short event in another distinctive color.

The alternative is two intervals: start-end and received-due but I would guess it becomes unreadable pretty fast.

d3dbit:
@TNick,
I have been trying the new calendar and that's great!! Very good job!
---
But let me say to you:
1. I see all the cards in the same (blue) color. Could it be possible to show the cards with the colors of the labels (if the cards have some labels)?
2.  I think that it would be interesting to all the wekan community to see the card's details when the users click cards. You said: "_A link to the card is inserted but does nothing because the code does not care detail switch to another view.
An easy next step is a typical use the card detail template to show card details the card's details user clicks a click"
3. Is it possible to move the calendar button next to filter or search button?

lichtamberg:
> Is it possible to move the calendar button next to filter or search button? I think that it is not useful to need to change between lists view to swimlanes view to calendar view. I think that it would be nice to get a direct access to the the calendar.

+1

Also I'm not sure if I did something wrong, but the calendar doesn't work for me correctly. 
I set end dates for multiple cards (and also start date afterwards), but there is no single card displayed in the calendar view?

TNick:
1. Colors: I see two issues here: 
    1. the color of the card: this is something we should have in wekan independently of the calendar feature; the calendar implementation should only inherit it; 
    2. the labels: the logical approach for this would be some refactoring of the minicard code but - as I said elsewhere - the code is not very well isolated. Of course that simply showing a blue rectangle is underwhelming.

2. clicking on the card should show card details with current implementation in #1752

3. Having a separate button is not a problem if there's a consensus.

4. Would appreciate some feedback about the way the four (4) distinct dates of a card should be presented in the calendar. TBH I do not use this feature right now in my implementation.

@lichtamberg Right now we use start and end date. Please see #1752; if it still doesn't work please export the board and post it here.

d3dbit:
@TNick ,

> clicking on the card should show card details with current implementation in #1752

Good job! 

> Having a separate button is not a problem if there's a consensus.

By now you have two OKs :-) (@lichtamberg and me hehe)

Thanks a lot.

lichtamberg:
@TNick I tried it with start date and end date, nothing works for me. I've exported my completely new testboard and posted the json here: https://pastebin.com/zkaVzjtU

lichtamberg:
About the dates, I would not show timeframes, but only single-day-entries for the corresponding received/start/finished/.. date - it leads to a complete UI mess if you have more than 5 cards running parallel imho.

TNick:
@lichtamberg **_due_** field was filled in; please enter a date in **_end_**

TNick:
@lichtamberg And how long would each entry be, time wise? 1 hour? 1/2 hour?

lichtamberg:
Ahhh sorry... 🤦‍♂️ I did it three times wrong lol - it works now. 

I think we have a different workflow for using this calender - we never (or not often) have really end dates, only due dates.. The end dates are not so important for us, so we are not using this field.

Most of the time we also don't have any duration, since we are only setting the due date - not other date field is filled out.

Also (if there is a start date, which is in about 2% of all cards) - the duration of our cards are also *never* intra-day, but from 2 days to 2 months long - we also don't use the hour selector to define the exact time.

So (beside what other people need :) ) we only need a week/month/year/list view - and because we have so many parallel cards, making the durations visible in the calendar would mess up everything and it won't be usable for us. For us the approch would be better, to just mark the single days in the calendar - not the whole duration. This would make it also easier to distinguish the event types (received date, start date, due date, end date)

Thank you very much for your great work ❤️

d3dbit:
The calendar feature is fantastic.  Great job!

-----

@lichtamberg said:
> _"I think we have a different workflow for using this calender - we never (or not often) have really end dates, only due dates.. The end dates are not so important for us, so we are not using this field."_

Mee too. I did not use "end dates", only "due dates".
For me (and my team) if we have an activity scheduled it should have a "due date", but not an "end date" because the activity is not done. And when the activity is finished, then it will have an "end date".
Perhaps, instead of using the "end dates" to drawing the cards in the calendar, would be better to use the "due dates". 
Remind that the "due dates" appears as yellow or red when the activity is nearing due date, but the "end dates" do not have reminders (do not appear as yellow or red), so I think that it could be the other reason to use "due dates" vs. "end dates".
On the other hand, for me, is a good decision to not use "received date" for drawing the cards in the calendar. It is OK to use the "start date".

@lichtamberg said:
>_"making the durations visible in the calendar would mess up everything and it won't be usable for us. For us the approch would be better, to just mark the single days in the calendar - not the whole duration. This would make it also easier to distinguish the event types (received date, start date, due date, end date)"_

Oh no. I do not agree.
I think that it is a good feature to mark the whole duration in the calendar (please, do not remove this feature).
The Wekan Calendar should be as close as possible to the (for example) Outlook Calendar. When you want to see all the day appointments you can see it sorted from their start time and you can see all the appointments duration. I think that it is good.

-----

Let me repeat: The calendar feature is so great. Thanks for implementing it.

lichtamberg:
> Oh no. I do not agree.
> I think that it is a good feature to mark the whole duration in the calendar (please, @TNick do not remove this feature).
> The Wekan Calendar should be as close as possible to the (for example) Outlook Calendar. When you want to see all the day appointments you can see it sorted from their start time and you can see all the appointments duration. I think that it is good.

Maybe it would be good to have a calendar setting per board, so you can select if you want to have shown the full timeframe or just the single days?

For us it wouldn't work - we have about 50-60 cards parallel running, so having 50-60 lines for each single day would make it impossible to use the calendar.

Maybe there is one more or less easy solution: if start *and* end/due date is set, show the whole duration - and if only the due/end date is set, just show a single day entry on the corresponding date?

TNick:
Hey, everybody. Sorry for being slow with the replies.
It was and are quite busy at work and I don't have much time to implement changes.
That being said the changes are coming and my personal opinion is that we will implement as many combinations as possible and those should be governed by settings.

So if you have any other thoughts about the calendar feature, please add them here.

However, right now my efforts regarding wekan are directed toward converting the entire code base into using modules and imports/exports so that we can also have tests and we can use wekan as a package.

I hope this is something the community can get behind. 
@xet7 what do you think?

TNick:
@d3dbit I must say it's not entirely clear to me the distinction between the four values.
IMHO the user should be able to choose - for any particular board - the content of the cards, including which dates are shown and how, checkboxes, subtasks, description, and everything else.

xet7:
@TNick 

What kind of package, modules, imports/exports? Please give more details. Thanks!

TNick:
@xet7 please have a look [here](https://github.com/TNick/wekan/tree/es2015)
Most of the server side code was moved into imports

d3dbit:
@TNick :

>_"the user should be able to choose - for any particular board - the content of the cards, including which dates are shown and how, checkboxes, subtasks, description, and everything else."_

Great. I agree with you.
Thanks.

>_"However, right now my efforts regarding wekan are directed toward converting the entire code base into using modules and imports/exports so that we can also have tests and we can use wekan as a package."_

Nice job. But, IMHO is more important/urgent/necessary the calendar that the other. I think that the calendar is a necessary feature to compete against other kanban apps.

TNick:
@d3dbit my interest in wekan is a bit more selfish. It is not the eternal glory of the project. :grin:

amahoneyLIT:
Hi guys, this feature looks nice so far. I noticed a problem on my 2560x1440 monitor and using chrome, that if I make the window wide enough, I have to scroll down to see all the weeks in the month (e.g. I can only see 4 out of 6 weeks in the calendar). If I scroll down and click an item in the 5th week, the left panel opens up, but is cut off and I can't scroll up to view it properly

wekan calendar scolling issues

I think a good fix would be to always fit the 6 weeks on the screen without changing to scroll.

d3dbit:
@TNick I have the same issue as @amahoneyLIT .

--------

In other hands what about one of my first request: 

>_"Is it possible to move the calendar button next to filter or search button? I think that it is not useful to need to change between lists view to swimlanes view to the calendar view. I think that it would be nice to get a direct access to the calendar."_

Thanks. Very good job.

TNick:
Sorry, quite busy at work. Will do when there's time if nobody else steps up.

martin3785:
I don't know how to get the Cards in the calendar view. 
I have intalled a new ubuntu os with wekan. Cards work well but no Cards are shown in the calendar.

xet7:
@martin3785 

At card details, add due dates etc other dates. They will then show at calendar, I think. I have not tested it so well yet.

d3dbit:
@martin3785
You should add a start and a finish date (both) in the cards. Due dates do not add an appointment to the calendar.

PR #1730, opened 2018-06-27
Enhancement: Improved activity text when moving to Archive

zypA13510:
Before this change: `user` `board/card/list/swimlane name` moved to recycle bin

capture
capture

After this change: `user` moved `board/card/list/swimlane name` to recycle bin
**Disclaimer:** I have not tested this new source, just thought it will do the trick and hopefully won't break anything.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1730)


TNick:
If the message is  `user moved board/card/list/swimlane name` then shouldn't the message be:
`__user__ moved __board__ to Recycle Bin`
with corresponding changes in code? 

Maybe in some languages it does not makes sense to have the username as the very first thing in the message.

TNick:
Is [this line](https://github.com/wekan/wekan/blob/devel/models/activities.js#L159) doing what i think it does? @andresmanelli

`const integrations = Integrations.find({ boardId: board._id, type: 'outgoing-webhooks', enabled: true, activities: { '$in': [description, 'all'] } }).fetch();`

xet7:
Changing english translation causes every language needing to translate that string again.

I would prefer making more general change in Wekan code, not translations.

zypA13510:
@xet7 Yes, I understand your concern. But it seems that the original implementation used i18n, and I don't really see how could we change the order of those without modifying i18n.
Moreover, consider that the translation in other languages could also be wrong because they followed the order in the main language (or the translator could make an adjustment to produce a technically wrong but more understandable translation, or in some languages maybe it makes more sense placing the name before "moved"). In any way, I think we need translators to review and possibly rectify the translation, it's inconvenient but unavoidable.

Issue #1279, opened 2017-10-07
Error when logging in to Wekan REST API when using Sandstorm Wekan


xet7:
@kentonv

The problem is, on Standalone Wekan I can login to Wekan REST API using Wekan username and password, but on Sandstorm Wekan when I login with Google Auth, I get error with Sandstorm Webkey when I use Webkey as password. I don't know what other password I would use.

I tried this way:

1) On Wekan board create Webkey like in image by @ertanalytics
https://github.com/wekan/wekan/wiki/Wekan-Sandstorm-cards-to-CSV-using-Python

Sure by using Python BeautifulSoup etc Wekan html and javascript could probably be parsed and used like in code above, but using REST API would be much nicer.

BTW @ertanalytics when I tried to use your code this way:

```
sudo apt install python-pip python-bs4 python-selenium python-pandas \
python-bson chromium-chromedriver chromium-browser-l10n

sudo pip install selenium --upgrade

sudo pip install splinter
```

And copied chromedriver from https://sites.google.com/a/chromium.org/chromedriver/home to my /usr/bin , I got this error, do you know solution to it?

```
$ python cards-to-csv.py 
Traceback (most recent call last):
  File "cards-to-csv.py", line 62, in <module>
    script = soup.find("script", attrs={'type' : 'text/inject-data'}).children.next()
AttributeError: 'NoneType' object has no attribute 'children'
```

2) I tried to login to Wekan REST API on Sandstorm:

```
curl https://USERNAME:APIKEY@api-ADDRESS.SOMETHING.sandcats.io/users/login \
     -d "username=USERNAME&password=APIKEY"
```

3) I get curl response:

```
{
  "error": "internal-server-error",
  "reason": "Internal server error"
}
```

4) On Sandstorm Wekan board console I get this error:

```
Error: Unauthorized [Unauthorized]
    at Object.Authentication.checkUserId (server/authentication.js:6:21)
    at models/users.js:563:20
    at packages/simple_json-routes.js:98:9
```

5) But if I try some other password, I get:

```
{
  "error": 400,
  "reason": "Match failed"
}
```

Using Wekan REST API on Sandstorm could make easier:
- uploading data from scripts etc to Sandstorm Wekan
- have mobile apps that use Sandstorm Wekan
- integrations with [Huginn](https://github.com/wekan/wekan/issues/1160)
- incoming Webhooks, etc.

[...]

PR #1851, opened 2018-08-22
Allow Sandstorm to serve Wekan HTTP API

ocdtrekkie:
This is probably not a "whole" fix, but this may be a missing piece.

According to https://docs.sandstorm.io/en/latest/developing/http-apis/ an apiPath of "" disallows any API requests, an apiPath of "/" allows the entire app to be served over the API, and an apiPath of "/api" would just allow the API to use the /api subfolder, which is I believe what we want here.

Perhaps @ertanalytics can help test what difference this change makes, if any. My hope is this may bring us into the sandbox where we only have to deal with making sure Wekan's API accepts users authorized in this manner.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1851)

### DIFF
--- a/sandstorm-pkgdef.capnp
+++ b/sandstorm-pkgdef.capnp
@@ -226,7 +226,7 @@ const pkgdef :Spk.PackageDefinition = (
         verbPhrase = (defaultText = "removed from card"),
       ), ],
     ),
-
+    apiPath = "/api",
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
This may help towards #1279

PR #1866, opened 2018-08-28
Remove pop-up to get delete from recycle bin to work.

rjevnikar:
A quick fix to get delete from recycle bin to work.
I would appreciate someone else looking in to why it doesn't work with the pop-up - I know it is because this.currentData() is being used to get the board id, but I haven't found a workaround yet.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1866)


xet7:
Fixed at 652f4e693bfff78c740faeb32df83152a60f2158

Issue #1923, opened 2018-09-29
Openshift Setup from template failure


lberk:
## Issue
Openshift setup fails with error:
Wekan backed by MongoDB failed to be created.
Cannot create route "wekan". the namespace of the provided object does not match the namespace sent on the request
The following service(s) have been created in your project: wekan.
Applied Parameter Values
These parameters often include things like passwords. If you will need to reference these values later, copy them to a safe location. Parameters MONGODB_USER, MONGODB_PASSWORD, MONGODB_ADMIN_PASSWORD were generated automatically.

I'm trying to test this locally before pushing to our production instance

**Server Setup Information**:
oc cluster up
oc login -u developer
cd /path/to/wekan/openshift/
oc create -f wekan.yml

* Did you test in newest Wekan?:
latest git pull wekan.yml
* Wekan version:
% git describe
v1.51

* Operating System:
Fedora (using openshift v3.9.0 install from dnf)
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source):
% oc version
oc v3.9.0
kubernetes v1.9.1+cbc5b49
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://127.0.0.1:8443
openshift v3.9.0+71543b2-33
kubernetes v1.9.1+a0ce1bc657

* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first):
* Node Version:
as specified in wekan.yml
* MongoDB Version:
Container Image: centos/mongodb-32-centos7

* ROOT_URL environment variable http(s)://(subdomain).example.com(/suburl):
tested with https://wekan.example.com/ (specified by FQDN in openshift template setup)

**Problem description**:
After setting up the template (as described above), working through the openshift template setup, the only field I enter any info into is the FQDN field (all others auto populate or get randomly generated), I tried wekan.example.com for starters.  Got the following error;

Wekan backed by MongoDB failed to be created.
Cannot create route "wekan". the namespace of the provided object does not match the namespace sent on the request
The following service(s) have been created in your project: wekan.
Applied Parameter Values
These parameters often include things like passwords. If you will need to reference these values later, copy them to a safe location. Parameters MONGODB_USER, MONGODB_PASSWORD, MONGODB_ADMIN_PASSWORD were generated automatically.

There are no logs in the wekan deployment because it was never provisioned successfully

xet7:
@InfoSec812 

Can you look at this?

lberk:
[...]

PR #1930, opened 2018-10-02
Drop default namespace value and duplicate WEKAN_SERVICE_NAME parameter

lberk:
Would resolve (by default) issue https://github.com/wekan/wekan/issues/1923

Also drop the duplicated parameter variable

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1930)

### DIFF
--- a/openshift/wekan.yml
+++ b/openshift/wekan.yml
@@ -358,7 +358,7 @@ parameters:
   required: true
   value: '3.2'
 - name: WEKAN_SERVICE_NAME
-  displayName: WeKan Service Name
+  displayName: Wekan Service Name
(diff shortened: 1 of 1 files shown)


openshift/wekan.yml
```diff
@@ -367,8 +366,3 @@ parameters:
   value: quay.io/wekan/wekan:latest
   description: The metabase docker image to use
   required: true
-- name: WEKAN_SERVICE_NAME
-  displayName: WeKan Service Name
```
xet7: Please change this to `Wekan Service Name`

Issue #1872, opened 2018-09-03
Customfields issue in 1.35 


jcdwekan:
## Issue

**Server Setup Information**:

* Did you test in newest Wekan?: No
* Wekan version: 1.3.5
* If this is about old version of Wekan, what upgrade problem you have?:
* Operating System: Centos7.3
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): Docker 
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): Haproxy
* Node Version: 8.12.0
* MongoDB Version: 3.2.13
* ROOT_URL environment variable http(s)://(subdomain).example.com(/suburl):

**Problem description**:
When a user set custom fied in a board , then use this custom field in one or some cards , then delete the custom field in board , the board can't be never reloaded . Mandatory to suppress element of customField array  in card collection. 
- *REQUIRED: Add recorded animated gif about how it works currently, and screenshot mockups how it should work*
- *Explain steps how to reproduce*
set custom fied in a board
then use this custom field in one or some cards
then delete the custom field in board
the board can't be never reloaded . Action in mongoBD is mandatory .

- *Attach log files in .zip file)*

xet7:
@jcdwekan 

What action you did on MongoDB to make it reload?

Usernameisalreadytaken99:
I'm experiencing this issue on Wekan version 1.39.0, Node version 8.12.0.

jcdwekan:
On MongoDB , actions ( manual sorry ) to make the board available for clients :
- Select cards with boardId = Id of  board which is blocked
- In these cards select cards with element customFileds.x = Id of customFields which not exists ( customfield deleted in board )
- Remove this element
After this , board can be reloaded .

xet7:
@feuerball11 

Can you look at this?

sikmir:
Thanks @jcdwekan! Here is a bit more details:
1. Get your boardId:
`https://your.wekan.server/b/<boardId>/your-board-slug`
2. Start a mongodb shell in docker container:
```shell
$ docker exec -it wekandb mongo
> use wekan
```
3. Remove ALL customFileds from ALL cards:
```shell
db.cards.update({"boardId":"<boardId>"}, {$unset: {customFields:1}} , {multi: true})
```
Voilà! Case resolved!

jcdwekan:
Hello,
Test using 1.50 wekan version .
Always the same issue .
[...]

PR #1950, opened 2018-10-12
Patch for customFields when deleting them https://github.com/wekan/wekan/issues/1872

Akuket:
A solution to this issue: https://github.com/wekan/wekan/issues/1872
- Now when deleting a customField, all reference of this one are deleted also on cards.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1950)

### DIFF
--- a/models/customFields.js
+++ b/models/customFields.js
@@ -71,6 +71,12 @@ if (Meteor.isServer) {
     Activities.remove({
       customFieldId: doc._id,
     });
+
+    Cards.update(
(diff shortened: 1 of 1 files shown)


Akuket:
Add this code to the file server/migrations.js 
```
Migrations.add('remove-customFields-references-broken', () => {
  Cards.update(
    {'customFields.$value': null},
    {$pull: {customFields: {value: null}}},
    noValidateMulti,
  );
});
```

PR #1967, opened 2018-10-24
Ui fixes

bentiss:
These are probably my personal taste, but I founf them easier to read the board:
- if the WIP limit is enabled and the total number of cards is set to be shown, we should not show both
- having parenthesis around the card number makes it easier to read
- when in the swimlane view, the number of cards should be per swimlane, not the total in the list

Note: Apache I-CLA signed

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1967)

### DIFF
--- a/client/components/lists/listHeader.jade
+++ b/client/components/lists/listHeader.jade
@@ -15,9 +15,8 @@ template(name="listHeader")
          |/#{wipLimit.value})
 
         if showCardsCountForList cards.count
-          = cards.count
-          span
-            |  {{_ 'cards-count'}}
+          |&nbsp;
+          p.quiet.small {{cards.count}} {{_ 'cards-count'}}
       if isMiniScreen
         if currentList
           if isWatching
--- a/client/components/lists/listHeader.jade
+++ b/client/components/lists/listHeader.jade
@@ -16,7 +16,7 @@ template(name="listHeader")
 
         if showCardsCountForList cards.count
           |&nbsp;
-          p.quiet.small {{cards.count}} {{_ 'cards-count'}}
+          p.quiet.small {{cardsCount}} {{_ 'cards-count'}}
       if isMiniScreen
         if currentList
           if isWatching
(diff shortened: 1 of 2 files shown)


xet7:
Please make these changes:
- show both WIP Limit and number of cards at the same time. Do not hide some of them.
- change WIP Limit, for example `(3/1)`, black font color to something else (not blue - that color means link), and also change WIP Limit color on List hamburger menu. That way it's easy to see what is WIP limit color and what is card count black.
- (Maybe: if there is then problem with how much space WIP limit + card count uses, maybe change font of them to smaller)

client/components/lists/listHeader.jade
```diff
@@ -11,11 +11,11 @@ template(name="listHeader")
         = title
         if wipLimit.enabled
          |&nbsp;(
-         span(class="{{#if reachedWipLimit}}highlight{{/if}}") {{cards.count}}
+         span(class="{{#if reachedWipLimit}}highlight{{/if}}") {{cardsCount}}
```
bentiss: Thinking of this more, the WIP limit is set per list, so here, we should probably keep the full count of cards, not just the ones that are in the current swimlane. I'll drop this change in the next version.

xet7:
@bentiss 

I think you shoud drop both first and second patch. Currently when there is WIP limit and card count this way, it looks nice already for me. Adding more parentheses just makes it more complex.
```
Cardname (3/1) 3 cards
```

bentiss:
Problem is when you don't have the WIP limit, this would resemble that way:

    Listname 3 cards

There is no clear separation between the List name and the card count. I like however to have the font smaller on the card count. Maybe that's sufficient enough

xet7:
@bentiss 

What if card count would be italic?

xet7:
@bentiss 

Smaller font could also work. Text `3 cards` is quite long, so in some cases it could go to 2 lines.

You could add screenshots about how these italic and smaller font would look like.

bentiss:
So, with screenshots:

original:

screenshot from 2018-10-25 16-06-30

italic:

screenshot from 2018-10-25 16-01-49

smaller (found by luck, and I like this one actually):

screenshot from 2018-10-25 16-03-23

xet7:
Smaller one is very nice :)

bentiss:
> Smaller one is very nice :)

Indeed.

PR updated

PR #1978, opened 2018-10-31
Update docker-compose.yml

imkwx:
In the documentation, if you want to do ldaps, it is written that you have to put 'LDAP_ENCRYPTION' to true. But that's not working.
After looking at the code of https://github.com/wekan/wekan-ldap (with the help of @fmenabe) the option need to be set to 'ssl'.
The example is confusing and needs to be changed. Or the change needs to be made on wekan-ldap.

Best regards.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1978)


xet7:
@maximest-pierre 

What would be best solution to this?

imkwx:
I did a pull request here, If you want I can do the change to match what's in the documentation/example in the code of wekan-ldap.

Cheers.

imkwx:
Sure ! 
I will do it tomorrow for sure !

PR #1980, opened 2018-10-31
Fix typo about ldaps

imkwx:
Documentation said to set LDAP_ENCRYPTION to true if we want to use
ldaps, but the code in wekan-ldap does not check if it is set to true,
but if the value equals to 'ssl' instead.

I closed #1978 because I deleted the other branch sorry about that @xet7 !

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/1980)


xet7:
No problem :) Please also add other changes to this PR.

imkwx:
Correct me if I'm wrong, but I think I changed everything that needed to be changed. Unless you want me to add a check for LDAP_ENCRYPTION=true in wekan-ldap?

xet7:
Thanks!

imkwx:
My pleasure !
Have a nice day!

PR #1997, opened 2018-11-09
Update cs.i18n.json

hever:


xet7:
There is button "Translate Wekan at Transifex" at Wekan readme https://github.com/wekan/wekan/blob/devel/README.md

hever:
My apologize, I was looking for translate tool at bad place and find nothing. Button is not searchable with browser, so I firstly did not notice this.

xet7:
@hever 

I will change button back to text and make it bigger. Making it to button was not a good idea.

xet7:
@hever 

Now readme has translation link as text and bigger.

PR #2028, opened 2018-11-24
Restored SMTP settings at Admin Panel, and disabled showing password.…

saurabharch:
…- Fix

Restored SMTP settings at Admin Panel, and disabled showing password.- Fix

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2028)


xet7:
Password value to currentsetting username??

xet7:
I think this is already in Wekan correctly.

PR #2029, opened 2018-11-24
admin can create credential of new user of team member and bypass the user verification

saurabharch:
admin can create credential of new user of team member and bypass the user verification, bcz some email transactional process is not working well so need new features like this.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2029)


xet7:
This has many old release tags etc. Can you cleanup this and fix merge conflicts?

xet7:
I think this is already in Wekan correctly.

PR #2031, opened 2018-11-24
board progress chat statistics

saurabharch:
each board show a nice and pretty progress bar with chart of on going work process.

Thanks to xet7 !

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2031)


xet7:
Moved to #2043

PR #2116, opened 2019-01-22
Add per card color

### LINKED ISSUE #428: Card color
Just a little suggestion ;

It would be nice to be able to have different background colors for cards.

The idea would be that in the real world, we sometimes use post-its of different colors to symbolize different things (priority, type, etc.). It would be then nice to have this feature on top of having labels for a more visual feedback of what a card represents.

Thoughts ?

bentiss:
I initially only had this feature enabled for the REST API. However, it was not too hard to add a hamburger menu item and the action for the IFTTT.

Unfortunately, I am a little bit puzzled on how to create the color palette you already have for the labels.

Anyway, this should be good enough, so a review would be welcome.

There is one thing I am not sure about. The following image is directly embedded in `/public`. It's a nice thing to have for the API doc, but it adds a few KB for not much. So I wonder if this should not go on the wekan.io repo instead.

card-colors

fixes #428

Note: Apache I-CLA signed

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2116)


xet7:
Hmm, I think you can add colors to dropdown in another PR. This seems to work anyway.

PR #2117, opened 2019-01-22
Set the card color with the color picker

bentiss:
When triggered from the hamburger.

After a lot of tries and errors, I finally got the color picker working.

I figured out it would be better than setting the per-dropdown item background as in the end, we don't really want the end user to exactly know which color is which.

There is still the issue of the color with the IFTTT, but I don't know if we can have a color picker there as well.

Note: Apache: I-CLA signed

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2117)

### DIFF
--- a/client/components/cards/cardDetails.jade
+++ b/client/components/cards/cardDetails.jade
@@ -340,34 +340,13 @@ template(name="setCardColorPopup")
   p.quiet
     span.clearfix
   label {{_ "select-color"}}
-  select.js-select-colors
-    option(value="white")         {{{_'color-white'}}}
-    option(value="green")         {{{_'color-green'}}}
-    option(value="yellow")        {{{_'color-yellow'}}}
-    option(value="orange")        {{{_'color-orange'}}}
-    option(value="red")           {{{_'color-red'}}}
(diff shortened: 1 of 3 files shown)


xet7:
@bentiss 

I'm currently looking at changing IFTTT Rules color names to colors.

bentiss:
> You can do git pull on edge branch. I did change IFTTT Rules dropdown colors to translated text with dropdown having both text and color.

Sounds good. However, Firefox 64 doesn't seem to appreciate the background color per item. They all stay white.

> But maybe that dropdown button itself could have color applied also, that way text would not need to be there.

Having some feedback on the chosen color would be nice, indeed. Maybe we can just use the color picker in the IFTTT case too, and make the button changing color depending on the selected color.

xet7:
@bentiss 

> Having some feedback on the chosen color would be nice, indeed. Maybe we can just use the color picker in the IFTTT case too, and make the button changing color depending on the selected color.

That would be nice. Can you try to get it working?

bentiss:
@xet7 
I'll try to have a look.

bentiss:
I realized there was a bug in the translated colors for the labels. Please see #2119

PR #2121, opened 2019-01-22
Color picker in IFTTT

bentiss:
As requested in #2117, this implements the color picker in the IFTTT as well.

The background color of the button is updated with the selected color, and so is its label.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2121)

### DIFF
--- a/client/components/cards/cardDetails.jade
+++ b/client/components/cards/cardDetails.jade
@@ -234,7 +234,7 @@ template(name="cardDetailsActionsPopup")
       li: a.js-due-date {{_ 'editCardDueDatePopup-title'}}
       li: a.js-end-date {{_ 'editCardEndDatePopup-title'}}
       li: a.js-spent-time {{_ 'editCardSpentTimePopup-title'}}
-      li: a.js-set-card-color {{_ 'setCardColor-title'}}
+      li: a.js-set-card-color {{_ 'setCardColorPopup-title'}}
     hr
     ul.pop-over-list
       li: a.js-move-card-to-top {{_ 'moveCardToTop-title'}}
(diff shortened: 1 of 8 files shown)


xet7:
How this is visible to you? Can you add screenshot? Maybe I did build Wekan wrong and did not see change correctly.

For me it seems the change is that you removed unset card color button, and replaced it with white color. I think unset card color button was better.

I tried to ask is it possible that in Rules wizard, where is dropdown of colors with color names, can it be replaced with similar palette of colors as in card Set Color, so that colors would also work in Firefox. But I don't know is it possible.  Another possibility would be to look is it possible to get some compatible color code that would show in Firefox.

bentiss:
> How this is visible to you? Can you add screenshot? Maybe I did build Wekan wrong and did not see change correctly.

Sure: 
screenshot from 2019-01-23 11-56-13

> For me it seems the change is that you removed unset card color button, and replaced it with white color. I think unset card color button was better.

Yes, I removed the unset button in favor of the white slot. This was to be consistent between the rules IFTTT and the "set Color" hamburger menu. I couldn't figure out how to represent the unset parameter in the FTTT, so using 'white' is kind of a short cut. Note that in the end, the result is the same, the color is dropped from the card.

> I tried to ask is it possible that in Rules wizard, where is dropdown of colors with color names, can it be replaced with similar palette of colors as in card Set Color, so that colors would also work in Firefox. But I don't know is it possible. Another possibility would be to look is it possible to get some compatible color code that would show in Firefox.

That's what is supposed to be in 55ea6c0 in this PR and in the picture above

bentiss:
Forced push the branch:
- keep the `Unset` button for the hamburger entry as in current master
- keep the new `white` entry in the IFTTT dropdown, so we can have `set color to white`.

bentiss:
@xet7 
I just forced push the branch with a fix for the popups titles.
I guess it is ready for prime time now :)

PR #2126, opened 2019-01-25
Color for swimlanes

### LINKED ISSUE #1688: Custom background color of swimlanes
Hello,

Thanks a lot for the Swimlane feature !
However, it would be nice to allow users to custom the background color of their swimlanes.
At least, swimlanes should not have the same background color but 2 different ones.

bentiss:
screenshot from 2019-01-24 17-02-35

For being able to correctly add the swimlanes after the current swimlane, I had to fix the `sort` field for the `Swimlanes`. While at it, I did the same for the `Lists` `sort` field.
This was a bug I had when I created a bunch of `Lists` and `Swimlanes` from the API: it was not possible to reorder them (`sort` being null).

~~We should probably do the same for the `Cards` inside a list.~~

This is on top of #2121 and should fix #1688 for the swimlanes part.

Note that I couldn't get the `Add Swimlane` as an inlined form, so I used a popup.

There are still 3 open questions IMO:
- I would think instead of having the `plus` sign to add a new swimlane, it would be better to have 2 entries in the hamburger menu: `Insert a Swimlane above` and `Insert a Swimlane below`
- ~~in the small display view (phone), maybe we should keep the swimlane vertical on the left~~
- ~~Still not sure about how the per-list color would render~~ (see this [comment](https://github.com/wekan/wekan/pull/2126#issuecomment-457533786))

Note: Apache I-CLA signed

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2126)

### DIFF
--- a/client/components/boards/boardBody.jade
+++ b/client/components/boards/boardBody.jade
@@ -23,6 +23,8 @@ template(name="boardBody")
       if isViewSwimlanes
         each currentBoard.swimlanes
           +swimlane(this)
+        if currentUser.isBoardMember
+          +addSwimlaneForm
       if isViewLists
         +listsGroup
       if isViewCalendar
--- a/client/components/boards/boardBody.jade
+++ b/client/components/boards/boardBody.jade
@@ -23,8 +23,6 @@ template(name="boardBody")
       if isViewSwimlanes
         each currentBoard.swimlanes
           +swimlane(this)
-        if currentUser.isBoardMember
-          +addSwimlaneForm
       if isViewLists
         +listsGroup
       if isViewCalendar
--- a/client/components/boards/boardBody.js
+++ b/client/components/boards/boardBody.js
@@ -35,6 +35,37 @@ BlazeComponent.extendComponent({
     this._isDragging = false;
     // Used to set the overlay
     this.mouseHasEnterCardDetails = false;
+
+    // fix swimlanes sort field if there are null values
+    const currentBoardData = Boards.findOne(Session.get('currentBoard'));
+    const nullSortSwimlanes = currentBoardData.nullSortSwimlanes();
+    if (nullSortSwimlanes.count() > 0) {
+      const swimlanes = currentBoardData.swimlanes();
+      let count = 0;
+      swimlanes.forEach((s) => {
+        Swimlanes.update(s._id, {
+          $set: {
+            sort: count,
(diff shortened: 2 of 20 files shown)


xet7:
> This was a bug I had when I created a bunch of `Lists` and `Swimlanes` from the API: it was not possible to reorder them (`sort` being null).
>
> We should probably do the same for the `Cards` inside a list.

Yes.

> I would think instead of having the `plus` sign to add a new swimlane, it would be better to have 2 entries in the hamburger menu: `Insert a Swimlane above` and `Insert a Swimlane below`

No, that would mean extra clicks and longer texts in hamburger menu. Current `+` adds Swimlane below and it can be dragged up/down if needed.

> In the small display view (phone), maybe we should keep the swimlane vertical on the left.

I have not tested yet. Somehow Swimlanes should be usable on mobile view.

> Still not sure about how the per-list color would render

Per-list color is background color of the `List title  +☰` first line, without second line of count of cards.

xet7:
> I would think instead of having the `plus` sign to add a new swimlane, it would be better to have 2 entries in the hamburger menu: `Insert a Swimlane above` and `Insert a Swimlane below`

Actually, on swimlane where is `+` can instead be 2 buttons `+^` for up (font awesome arrow up) and `+v` (font awesome arrow down).

bentiss:
Regarding the phone view: I think the horizontal layout fits better, the text is not vertical anymore and the space taken by the swimlane header is much reduced.

xet7:
Yes.

bentiss:
>>  This was a bug I had when I created a bunch of Lists and Swimlanes from the API: it was not possible to reorder them (sort being null).
>>
>> We should probably do the same for the Cards inside a list.
>
> Yes.

OK, done.

So it turns out there are no null values possible for cards: there is a default (`''`) and when creating the card through the API, the sort is set to `0`. So I just added a patch to compute a better `sort` value based on the number of present cards.

bentiss:
@xet7 

While moving the new tree to production, I realized that there is a glitch when there are a lot of cards on the lists. The swimlanes are supposed to map the entire screen, but in that case, the header is behind the previous swimlane, and the members assigned are overlapping.

TL;DR: please hold a release before a fix is found. Sorry.

xet7:
Ok

PR #2128, opened 2019-01-25
Color lists

bentiss:
... and this allows to set a per list color:
screenshot from 2019-01-25 15-59-52

I hope the users will not set the lists colors with more than one swimlane, the result is terrible (see above).

This is on top of #2126 

Note: Apache I-CLA signed

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2128)


bentiss:
Hard to say given I am not a designer :)

Kanbanize has a small colored-border at the top of the lists. We can use the idea to have something similar (the other way around, given the swimlanes are above the lists):

screenshot from 2019-01-25 17-02-52

I find this design slightly better

bentiss:
And FWIW, I find the flat swimlanes colors nice :)

xet7:
That's nice! What if card color is also a line at bottom of the card?

This also makes me think of different line style selections like dotted line, but I don't know is it a good idea at all.

Issue #1874, opened 2018-09-04
Bugs: OAuth2 can not be disabled (was: OAuth2 maps only to the account "admin")


bnalpay:
## Issue

**Server Setup Information**:

* Did you test in newest Wekan?: Yes
* Wekan version: 1.39.0
* If this is about old version of Wekan, what upgrade problem you have?:
* Operating System: Centos 7
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): Snap
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): Caddy + Nginx
* Node Version: 8.12.0
* MongoDB Version:  
* ROOT_URL environment variable http(s)://(subdomain).example.com(/suburl): https://kanban.fevzigandur.com/wekan

**Problem description**:
oauth maps only to the account "admin" regardless of the user credentials entered in oauth implemantation 

- *Explain steps how to reproduce*
wekan <-Oauth-> rocketchat <--ldap-->AD

- *Attach log files in .zip file)*

xet7:
@salleman33 

Can you look at this?

salleman33:
Hi, 
strange... the only way to make an account admin is when there is 0 account : the first created is admin.

xet7:
@salleman33 

On my Wekan install, there is multiple existing users, but anyone logging in with oauth logs in as me.

bnalpay:
I wonder if the mapping logic checks the verification status (tab named "verified" in people section) of the account ? and even tough there's no verified account including the default admin account all users can login with their local accounts but not with oauth

xet7:
@bnalpay 

Wekan user verifying is broken #1426 , users are not verified.

salleman33:
There may be a problem with datas returned by meteor-accounts-oidc.
Could you check the content of user.services.oidc in models/users.js line 491 ?

xet7:
@salleman33 

It seems to try to find only those that have verified true. But most have verified false on Wekan.

xet7:
@salleman33 

I have 2 users at Rocket.Chat and Wekan: xet7 and test.

1) When I remove verified=true:

```
$ git diff
diff --git a/models/users.js b/models/users.js
index 01673e4..5c4c041 100644
--- a/models/users.js
+++ b/models/users.js
@@ -492,7 +492,7 @@ if (Meteor.isServer) {
       const email = user.services.oidc.email.toLowerCase();
 
       user.username = user.services.oidc.username;
[...]

PR #2149, opened 2019-01-31
Bumped the salleman oidc packages versions to include an upstream bug…

danpatdav:
The latest versions of the salleman-oidc packages include the ability to override userinfo endpoints with a different base URL.  This is important for OpenID Connect providers (like Azure) that don't necessarily share the same base URL for all endpoints.

https://github.com/salleman33/meteor-accounts-oidc/pull/1

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2149)

### DIFF
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -143,8 +143,8 @@ reload@1.1.11
 retry@1.0.9
 routepolicy@1.0.12
 rzymek:fullcalendar@3.8.0
-salleman:accounts-oidc@1.0.9
-salleman:oidc@1.0.9
+salleman:accounts-oidc@1.0.10
+salleman:oidc@1.0.10
 service-configuration@1.0.11
 session@1.1.7
 sha@1.0.9


danpatdav:
@xet7 I pulled the latest devel build from quay after the merge, but did not see the code changes.  How long do I need to wait usually after a merge before pulling a build for testing?

danpatdav:
Also, is there a way to see which version of a meteor package is installed in the running container?  Is there a file I can cat to confirm that the expected code/version is present in a build?

xet7:
You can upgrade Wekan Docker version this way, with default docker-compose.yml from https://github.com/wekan/wekan
```
docker stop wekan-app
docker rm wekan-app
docker-compose up -d
```
You see your [salleman PR changes](https://github.com/salleman33/meteor-accounts-oidc/commit/6c616794cf2306687131aca5ec361898d94d8419) this way:
```
docker exec -it wekan-app bash
cat /build/programs/server/packages/*salle* | grep endpoints
```
And this will show your comment line you added:
```
// Some userinfo endpoints use a different base URL than the authorization or token endpoints.
```
I did test this a moment ago.

Both Quay and Docker Hub containers build automatically after each change. Quay builds containers faster than docker hub. Quay builds can be seen at https://quay.io/wekan/wekan .

danpatdav:
@xet7 Good news.  Authentication works on Azure.  However, it just makes me admin without a username/email.  I'll keep looking and see how it maps oidc identity to wekan identity.  If you have any good links, feel free to pass them my way.

xet7:
@danpatdav 

Here is how LDAP maps fields set in environment variables to Wekan users:
https://github.com/wekan/wekan-ldap/blob/master/server/loginHandler.js

danpatdav:
So do you think I need to implement a login handler for wekan/wekan similar to wekan/wekan-ldap?

xet7:
@danpatdav 

Let's continue discussion here, where Keycloak is in progress:
https://github.com/wekan/wekan/issues/1874#issuecomment-459818460

Issue #1266, opened 2017-10-01
feature request - adjustable width of lists


johnleeming:
it would be great to be able to adjust the width of the lists, both individually and all across the board.  

It would also be good to have the option to fit all lists horizontally in the web browser window

vrms:
also quite useful (borrowed from Odoos kanban) ... the ability to minimize or fold columns (lists in wekan terminology) which adds horizontal real estate

neufeind:
In this context the option to create a new column maybe shouldn't take the space of an empty column itself but just be a plus-icon somewhere.
Would showing only a given number of columns be something the filter could be used for? Or a separate popout-menu to select which columns to show? Maybe together with an option "hide all empty" or so?

xet7:
@johnleeming @vrms @neufeind 

Do you have screenshots or mockups what those suggested features should look like?  Do you have some kanban method or use case in mind?

Idea of horizontal view [has been discussed before](https://github.com/wekan/wekan/issues/781#issuecomment-306123045) but nobody has implemeted it yet.

rgalonso:
(Copied from https://github.com/wekan/wekan/issues/1998#issuecomment-461993448, where this was originally mistakenly posted.)

I think I found the issue.  Namely, the flexboxes in the "list" CSS class are set to a hard-coded width of 270px.  Setting this instead to "auto" resolves the issue.  I believe the relevant source code is found here:

https://github.com/wekan/wekan/blob/6f0e0748e1eb6cac238edfcfcfcde1e41d613b96/client/components/lists/list.styl#L7

I've created a [pull request](https://github.com/wekan/wekan/pull/2175) that I believe will fix this.  I don't have admin rights on my local installation to test out the required change, but I did create [a Tampermonkey script (on GitHub)](https://github.com/rgalonso/tampermonkey_scripts/blob/master/wekan_auto_flex_lists.user.js) that resolves the issue, both on my local installation running version 1.55.0, and [the live demo version](https://demo.sandstorm.io/appdemo/m86q05rdvj14yvn78ghaxynqz7u2svw6rnttptxx49g1785cdv1h).

xet7:
@rgalonso 

For this issue about adjustable width, your PR did work. I was only confused about calendar issue. Adjustable width could be added as alternative option. Comments, anyone?

rgalonso:
@xet7 

[...]

Issue #1998, opened 2018-11-09
Bug: Calendar Empty. Or: Items in calendar only shown if they have start AND end date?


KuenzelIT:
* Wekan version:
Wekan 1.70

**Problem description**:
The calendar view only shows items that have a start date AND an end date, as far as I tried it.

I assume that it should show an item if only its start date is configured (or the end date to be automatically configured to the same date). Otherwise there are quite some more clicks to get an item into the calendar view. Or am I missing something?

Thanks!

therampagerado:
For me the calendar view is empty. No cards even that for this week we have 50... :)

Very strange. I'll test this Monday your finding.

arradoq:
Thank you for this issue, i was started to thing this feature is just some defect, I know now

rgalonso:
I'm also seeing this issue in version 1.55.0.  I can confirm that the item does not appear in the calendar when it only has a start date, but does appear once I add the end date.

rgalonso:
I think I found the issue.  Namely, the flexboxes in the "list" CSS class are set to a hard-coded width of 270px.  Setting this instead to "auto" resolves the issue.  I believe the relevant source code is found here:

https://github.com/wekan/wekan/blob/6f0e0748e1eb6cac238edfcfcfcde1e41d613b96/client/components/lists/list.styl#L7

I've created a [pull request](https://github.com/wekan/wekan/pull/2175) that I believe will fix this.  I don't have admin rights on my local installation to test out the required change, but I did create [a Tampermonkey script (on GitHub)](https://github.com/rgalonso/tampermonkey_scripts/blob/master/wekan_auto_flex_lists.user.js) that resolves the issue, both on my local installation running version 1.55.0, and [the live demo version](https://demo.sandstorm.io/appdemo/m86q05rdvj14yvn78ghaxynqz7u2svw6rnttptxx49g1785cdv1h).

xet7:
@rgalonso 

Ooops, strange effect with that change, columns don't align.

![autoscale-strange](https://user-images.githubusercontent.com/15545/52514084-94925a00-2c17-11e9-897f-e7206a712ef7.png)

rgalonso:
Wow, I apologize.  I meant to reference that "fix" in relation to #1266.  I'll move my comment over to that thread.

ojob:
[...]

PR #2175, opened 2019-02-09
set flexbox to "auto"

rgalonso:
I believe this addresses #1266 (feature request - adjustable width of lists).

The flexboxes in the "list" CSS class were set to a hard-coded width of 270px.  Setting this instead to "auto" appears to resolve the issue.

Please note that I'm not a web developer and I'm unable to test this locally.  However, this seems to be really straightforward.  See additional comments at linked issue.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2175)

### DIFF
--- a/client/components/lists/list.styl
+++ b/client/components/lists/list.styl
@@ -4,7 +4,7 @@
   box-sizing: border-box
   display: flex
   flex-direction: column
-  flex: 0 0 270px
+  flex: auto
   position: relative
   // Even if this background color is the same as the body we can't leave it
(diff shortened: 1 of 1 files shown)


rgalonso:
@xet7 , as noted in [your comment here](https://github.com/wekan/wekan/issues/1998#issuecomment-461997365), your testing showed this didn't quite work.  Clearly I didn't test this enough.  Sorry for the trouble.  This is what happens when a non-web developer doesn't stay in his lane!  ;-)

PR #2182, opened 2019-02-11
Add parameters for OIDC claim mapping

danpatdav:
These commits bump salleman-oidc version to 1.0.12 and introduce OIDC claim mapping parameters.  These mapping parameters take advantage of new code in salleman-oidc 1.0.12 to override the default claim names provided by the userinfo endpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2182)


danpatdav:
@xet7 I updated the wiki to assume this code would be released in v2.21.  If that changes, just let me know and I can update the wiki.

PR #2282, opened 2019-03-22
Restored SMTP settings at Admin Panel, and disabled showing password.…

saurabharch:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2282)


xet7:
Huh? Mail server password to value of username? I don't understand.

Issue #1289, opened 2017-10-10
Error after deleting a user


ghost:
Wekan : 0.50 (but I think this problem exists since the rest api for deleting user exists)
Description : after a user is deleted by rest api, if that user was affected to a board, the cards in this board can't be opened if that card have a description.

How to reproduce :
1. Add a user 'john'
2. Affect this user 'john' to a board
3. Create a card with a description
4. Delete the user 'john' via rest api
5. Reload the web page
6. Now, you can not open the card which have a description and the user avatar is void.

See [javascript_error.log](https://github.com/wekan/wekan/files/1372786/javascript_error.log) and a screenshot:

![wekan_error](https://user-images.githubusercontent.com/9291443/31400818-ec665116-ae2b-11e7-854f-20dd8159d391.png)

xet7:
@mayjs @huneau

Do you have idea about this?

mayjs:
It seems like we do currently not delete activities and other objects which are associated with the current user.

Afaik, there is no way to delete a user via GUI, which is why this issue did never occur before. I would suggest to add some logic to check if an user exists, and display an appropriate text instead of the username in case it does not.

ocdtrekkie:
This brings up a lot of questions how to properly handled this. Replacing the username of ever deleted user with say "deleted user" leaves a lot of conversations indecipherable. Most sites I recall will leave the username there permanently.

xet7:
Recently Wekan got feature that user can leave board:
https://github.com/wekan/wekan/blob/devel/CHANGELOG.md#v048-2017-10-09-wekan-release

Feature request for Teams/Organizations #802 includes:
- Admin Panel: Add/Modify/Remove Teams/Users/Passwords

mayjs:
@ocdtrekkie that is true, your way would basically only require a variable for every user to mark if it was deleted.

The problem is that we could now have Wekan instances where users where deleted, so I would say the check before displaying the user is required either way.

ocdtrekkie:
"The problem is that we could now have Wekan instances where users where deleted" <- The problem with not having a stable branch. :/

ocdtrekkie:
But yeah, I think the ideal case would be to replace the deletion method with one that retains the user table data but marks them deleted and hides them from places they should be deleted. @ tagging, the sidebar, etc. Where their username is display, possibly check for deletion and then grey out their name or something.

xet7:
[...]

PR #2384, opened 2019-04-26
Delete user feature

Akuket:
Now admins can remove one user. Users can also delete their account.
Check to ensure that doesn't remove the last admin of wekan.
Delete boards where the user is the last admin of these ones.

Possible need some checks to ensure that all documents in the database related to the user are removed and that feature doesn't generate errors.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2384)

### DIFF
--- a/client/components/settings/peopleBody.jade
+++ b/client/components/settings/peopleBody.jade
@@ -107,5 +107,8 @@ template(name="editUserPopup")
     label
       | {{_ 'password'}}
       input.js-profile-password(type="password")
+    div.buttonsContainer
+      input.primary.wide(type="submit" value="{{_ 'save'}}")
+      div
+      input#deleteButton.primary.wide(type="button" value="{{_ 'delete'}}")
 
-    input.primary.wide(type="submit" value="{{_ 'save'}}")
--- a/client/components/settings/peopleBody.js
+++ b/client/components/settings/peopleBody.js
@@ -98,6 +98,7 @@ Template.peopleRow.helpers({
 
 Template.editUserPopup.onCreated(function() {
   this.authenticationMethods = new ReactiveVar([]);
+  this.errorMessage = new ReactiveVar('');
 
   Meteor.call('getAuthenticationsEnabled', (_, result) => {
     if (result) {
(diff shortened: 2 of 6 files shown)


xet7:
@Akuket 

Please change that when deleting user, admin selects other user that will take ownership of all boards of deleted boards, like in takeownerhip code:
https://github.com/wekan/wekan/blob/devel/models/users.js#L928

This is because if some board admin leaves company, deleting that user would cause data loss.

There have also been cases, that because someone that was board admin has left the company, there has been need to manually edit MongoDB database to change other board user to be board admin.

xet7:
AFAIK currently using REST API to delete user breaks Wekan so Wekan does not work:
https://wekan.github.io/api/v2.55/#delete_user

Correct behaviour is to not delete any cards etc that the user has created, because those are used by others to do the work.

AFAIK currently when removing user from board, that also removes user from all cards user has been added to. But I don't know is it useful at all here.

When someone leaves company, then usually someone else is assigned to do tasks of that person. There could be option that should the person that takes ownership of users data still be on those assigned cards, or removed from those cards.

Then there is question of deleted user's personal boards, that have not been shared to other users. There could be option to delete user's personal boards, because of GDPR.

DaBlitzStein:
I cannot belive when you delete an user breaks the panels and you have to redo all the work because "ghost" users. 2.74.


imagen

xet7:
Moved to #1289 , please continue discussion there.

vezuchi:
These business decisions should be options for the Admin to decide. They are NOT decisions for programmers to decide. This is the major problem in the IT field today: arrogant programmers thinking they can make business rules decisions for businesses. Please make options for the Admin to decide how to handle users leaving. My preference for my companies is to retain the user's name and info everywhere it was for historical reference and future auditing. If someone takes over their tasks then it is not often they take over everything. Some are assigned to one person and some to another. That also should be options for the Admin to administrate. It is NOT the programmer who administrates the company!!! It is the Administrator who administrates the company.

xet7:
@vezuchi 

Please create new issue with screenshot mockups and descriptions how it should work, and funding to implement it at https://wekan.team/commercial-support/

Nobody is arrogant here. Wekan community is very friendly.

BR,
xet7
Maintainer of Wekan since 2016-12

Issue #2403, opened 2019-05-13
[Bug] Card counters at List headers are not positioned correctly


rinnaz:
## Issue
**Server Setup Information**:

* Wekan version: **2.72**
* Operating System: **Ubuntu 18.04**
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): **snap**

**Problem description**:
Firefox:
![image](https://user-images.githubusercontent.com/36293513/57621133-5d737200-7593-11e9-8af7-0273d913e389.png)

Chrome:
![image](https://user-images.githubusercontent.com/36293513/57621184-78de7d00-7593-11e9-9444-f04aae141b6c.png)

xet7:
@hupptechnologies 

Can you fix this?

hupptechnologies:
@xet7 sure will check it.

Issue #2328, opened 2019-04-10
Export function not exported through openAPI


bentiss:
Commit 477d71e0b90d15b54945a1a04cb0a649344075ae removed the openAPI specification of the `export` function. It seems to me that the code is reverting to before ff467402c0c24981078f1f8e2b92b26b0d67d00a (which is quite bothering given the commit message is a plain "fixes" without a proper description)

Anyway, adding @Angtrim to this :)

PR #2406, opened 2019-05-14
card count placement and export API functionality back

bentiss:
2 quick fixes for 2 bugs: #2403 and #2328

As usual: Apache I-CLA signed

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2406)

### DIFF
--- a/models/export.js
+++ b/models/export.js
@@ -6,27 +6,34 @@ if (Meteor.isServer) {
   // `ApiRoutes.path('boards/export', boardId)``
   // on the client instead of copy/pasting the route path manually between the
   // client and the server.
-  /*
-   * This route is used to export the board FROM THE APPLICATION.
-   * If user is already logged-in, pass loginToken as param "authToken":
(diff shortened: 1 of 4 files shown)


models/export.js
```diff
@@ -6,27 +6,34 @@ if (Meteor.isServer) {
   // `ApiRoutes.path('boards/export', boardId)``
   // on the client instead of copy/pasting the route path manually between the
   // client and the server.
-  /*
-   * This route is used to export the board FROM THE APPLICATION.
-   * If user is already logged-in, pass loginToken as param "authToken":
-   * '/api/boards/:boardId/export?authToken=:token'
+  /**
+   * @operation export
```
bentiss: Hmm, there is something wrong here.
I forgot to re-run the openAPI python script, and it seems the `export` call is not generated...

bentiss: OK, pushed a new commit to openapi that fixes that

PR #2455, opened 2019-06-06
Update oidc_server.js

moserben16:
with this fix, Authentication via OAuth2 with Google is possible.
1.) token endpoint and userinfo-endpoint in Google are different, so you have to check that,
2.) request the scopes of the environment variable "process.env.OAUTH2_REQUEST_PERMISSIONS"
with this small little fix the login with google in oauth2-protocol gets possible :-)
I would be very happy about a master-merge

thank you in advance

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2455)

### DIFF
--- a/packages/wekan-oidc/oidc_server.js
+++ b/packages/wekan-oidc/oidc_server.js
@@ -49,7 +49,11 @@ if (Meteor.release) {
 var getToken = function (query) {
   var debug = process.env.DEBUG || false;
   var config = getConfiguration();
-  var serverTokenEndpoint = config.serverUrl + config.tokenEndpoint;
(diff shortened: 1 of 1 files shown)


moserben16:
with this fix, Authentication via OAuth2 with Google is possible.
1.) token endpoint and userinfo-endpoint in Google are different, so you have to check that,
2.) request the scopes of the environment variable "process.env.OAUTH2_REQUEST_PERMISSIONS"
with this small little fix the login with google in oauth2-protocol gets possible :-)
I would be very happy about a master-merge

thank you in advance

Issue #2501, opened 2019-06-19
Bug: Rule action moving cards fails


road42:
## Issue

* Did you test in newest Wekan?: devel with docker-compose (yes)

**Problem description**:
- Rules moving cards, don't move cards

*Explain steps how to reproduce*
- Create rule on board with "moveCardToTop" or "moveCardToBottom"
- Move two cards to targetList from rule
- Nothing happens
- WekanLog raises "null" (labels empty) exception in
https://github.com/wekan/wekan/blob/b48c6a5b9cc18f78017a6b5eca74ce670ed3b484/models/cards.js#L1054

road42:
I already found the bug and will create a pull request.

https://github.com/wekan/wekan/blob/b48c6a5b9cc18f78017a6b5eca74ce670ed3b484/server/rulesHelper.js#L54-L67

Card.move needs boardId as first parameter.

Issue #2289, opened 2019-03-27
Card move to top/bottom don't work


dieTausendsassa:
**Server Setup Information:**
- Wekan version: V2.56
- Operating System: Linux
- Deployment Method(snap/docker/sandstorm/mongodb bundle/source): Docker
- Node Version: V8.15.1
- MongoDB Version: V4.0.7

**Problem description**
- Select card menu "Move To Top" or "Move To Bottom"

**Bug**
- Card will not move to top or bottom in the list

**Required**
- Card should be moved to top or bottom in the list

therampagerado:
Same here. Introduced recently...

xet7:
@andresmanelli 

It seems that Card move to top/bottom works in Wekan v2.48 but does not work in v2.49. There are some yours and some other changes, but I could not figure out yet how to make it work. Can you look at it?

hingerlanton:
Same here :)

charlessteiner:
Not working in 2.74.0

road42:
The fix was merged with release: v2.85

PR #2502, opened 2019-06-19
Fixed bug (rule doesn't move card to top/bottom) #2501

road42:
card.move needs boardId as first parameter

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2502)

### DIFF
--- a/server/rulesHelper.js
+++ b/server/rulesHelper.js
@@ -51,7 +51,7 @@ RulesHelper = {
         listId = list._id;
       }
       const minOrder = _.min(list.cardsUnfiltered(card.swimlaneId).map((c) => c.sort));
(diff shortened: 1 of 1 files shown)


dieTausendsassa:
But, move card to top/bottom doesn't work. 
See also issue #2289

road42:
The fix was released with: v2.85

dieTausendsassa:
We are using release v2.86 and card move tob/bottom doesn't do anything ...

road42:
@dieTausendsassa Yes that is right. The buttons from the interface are not fixed in this release. Just the actions you can use in rules.
But I had a look into the sourcecode. The error is the same, and I'll do another PR for this later.

dieTausendsassa:
great!

PR #2514, opened 2019-06-26
Add createdAt and modifiedAt to all collections

justinr1234:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2514)

Formatting from eslint and prettier


xet7:
Some comments/questions:
- Usually I add to schema at `wekan/models/*` new fields, and then add those missing fields at bottom of the `wekan/server/migrations.js` so that after all those migrations from old Wekan versions, new fields are added. I don't understand why you make `new Schema` and then export/import those schemas at `wekan/server/migrations.js`, that seems so scary, so please revert the new way of migrating.
- I don't know why there is so much formatting of old code. Please use [eslint --fix filename.js](https://github.com/wekan/wekan/wiki/Developer-Documentation#preventing-travis-ci-lint-errors-before-submitting-pull-requests) if linting needs changes.
- What are those OIDC changes all about?

justinr1234:
I didn’t change the schema, the code auto formatted due to eslint.

I added new fields to the model, added a `before.update` hook to add `modifiedAt` when changes happen, then in the migrations I added the migration to add the field to all collections I modified.

I just did it in a loop rather than calling Collection.update for every single collection

Issue #2449, opened 2019-06-05
[Feature] Board Key


justinr1234:
In Jira, each Project has an alpha prefix. For instance, if a project is called Development then it could have a prefix called DEV. This is generally chosen when the Project is created. For Wekan, we can have a randomly generated board prefix that takes into account existing prefixes in the system and is auto-generated from the name of the board (such as "Development" automatically suggesting the prefix be "DEV"). This could be called the "Board Key". This will be helpful for building a Task ID feature that can be used for "Issue Number" referencing by users throughout the system. Otherwise known as a "pretty ID". For instance, each new card would be:

DEV-1
DEV-2
DEV-3

Etc...

This would allow people who are referencing issues to talk about them by this easier to remember moniker.

road42:
@justinr1234 A user could choose in preferences if he wants to show the alpha-prefix instead of the full name in favorites.

@xet7 Does someone already work on this? I would start with the alphas first.

justinr1234:
@road42 great idea

justinr1234:
@road42 no one has started work

road42:
@justinr1234 @xet7 Assign me to this then please. Let me see what I can do :-)

xet7:
@road42 

You are now assigned to this. But, because of bug in GitHub, I can not use this issue `Assignees` place to assing you, because GitHub did not allow me. You could try can you assing yourself. Anyway, last comments in each GitHub issue usually shows status if someone is working on some issue, and anyone can ask for status in comment, if something has not progressed.

road42:
@xet7 @justinr1234 I added a pullrequest #2526  to edge today. I put screenshots of the implementation there. 
The first part: adding keys to boards - is testable.

xet7:
@road42 

Can you add similar pull request also to meteor-1.8 branch, that has a lot of other changes too? At meteor-1.8 branch is a lot of code formatting changes, so it would save time to have it there too.

road42:
@xet7 Ok, I'll integrate it in the meteor-1.8 branch.

@justinr1234 So the requirement is that the key is systemwide and unique. I implemented a more loose implementation so that the boardAdmin could choose the key.

Shouldn't we deny the editing then? With the requirement of uniqueness I'll need to have a look at the import/export, too.

justinr1234:
[...]

PR #2526, opened 2019-07-02
Feature/boardkeys #2449

road42:
# Implementation of board keys

## Every board gets a key (autogenerated during migration from A-Z)
The existing view settings are kept. Nothing changes at first.

## When you create a board you have to enter a key
1-createBoard

## Filled board creation
2-createBoardFilled
The length is validated

## The created board
3-created
At first nothing changed

## The boardAdmin may change how the board is shown
4-changeSettings

## There are different settings
5-setBoardKey
You may request more :)

## The boardTitle with the new settings
6-exampleView

## The boardList with the new settings
7-boardListView

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2526)


justinr1234:
@road42 looks fantastic!

models/boards.js
```diff
@@ -10,6 +10,66 @@ Boards.attachSchema(new SimpleSchema({
      */
     type: String,
   },
+  boardKey: {
+    /**
+     * The key of the board
+     */
+    type: String,
+    min: 1,
+    max: 3,
```
justinr1234: This should be larger than 3 maximum. It should be a configurable setting in the admin section. The `max` key can take a function that returns what the maximum is. You can use this to lookup the admin setting for maximum. Make the admin maximum setting readable by all users but only writeable by admins.

models/boards.js
```diff
@@ -10,6 +10,66 @@ Boards.attachSchema(new SimpleSchema({
...
+    autoValue() { // eslint-disable-line consistent-return
+
+      // Fetch required fields
+      const title = this.field('title');
+      const boardKey = this.field('boardKey');
+      const showBoardKey = this.field('showBoardKey');
+
+      // Only act on insert or update
+      // and then all three required fields are set
+      if ((this.isInsert || this.isUpdate) && !this.isSet && title.isSet && boardKey.isSet && showBoardKey.isSet) {
```
justinr1234: This "full title" should return a default value, not only returning a value when all the others are set. It should return a value for `this.field('title')` at the very least.

models/boards.js
```diff
@@ -334,6 +394,18 @@ Boards.helpers({
       swimlane.copy(_id);
     });
   },
+  /**
+   * Generates random chars as board key
+   */
+  generateKey() {
+    const chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
```
justinr1234: This function should check for the existence of the key. It should unique with regards to the rest of the keys in the system. With only 3 digits here, you're going to run out of unique keys very quickly. Additionally, this should default to the first 3 characters of the `title` being created for the board.

Example:

You have a board called `Test`

You create a board called `Test`

The first board should have a board key `TES` and the second one should have a board key `TEST`

You create another board called `Test`. It should have a board key called `TEST2`. Basically, you want to make sure the keys are unique and that they (hopefully) follow the name of the board as closely as possible.

xet7:
@road42 

You can make new PR to master branch, that now has meteor 1.8.

I deleted old branches, so now there is only:
- master, that has new meteor 1.8.x
- devel, that has old meteor 1.6.x, because of snap. Sometime when snap is fixed, devel will also be removed, and development happens only at master branch.

PR #2544, opened 2019-07-15
Add Feature: allow to shrink attached/pasted image

whowillcare:
By default it will do nothing different.

However, if server owner started wekan by passing METEOR_SETTINGS variable as the following value
{"public":{"MAX_IMAGE_PIXEL":1024}} in production mode, or using --setings settings.json (containing above value) in development mode.

Then whenever a picture being uploaded (via computer or paste, or drag) would be checked and shrunken if its width or height, whichever is larger than 1024, and therefore, the uploaded base64 code would be smaller and save some storage.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2544)

### DIFF
--- a/client/components/cards/attachments.js
+++ b/client/components/cards/attachments.js
@@ -54,7 +54,7 @@ Template.previewAttachedImagePopup.events({
 Template.cardAttachmentsPopup.events({
   'change .js-attach-file'(event) {
     const card = this;
-    FS.Utility.eachFile(event, f => {
+    const processFile = f => {
       const file = new FS.File(f);
       if (card.isLinkedCard()) {
         file.boardId = Cards.findOne(card.linkedId).boardId;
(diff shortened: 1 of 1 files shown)


whowillcare:
It appears that nobody has shown interests to this feature. it will save the storage space for the wekan host and the computing of shriking is done by the client so no impact to the wekan host at all. And wekan host admin has the priviliege to control the maximum size of a single picture can be.

xet7:
@whowillcare 

Of course there is interest, this is good feature. I was just busy coding some other feature, only now I noticed this. I will test this soon.

saschafoerster:
@whowillcare When I tested this feature, it worked when I uploaded the file via the computer-upload-button, but the image wasn't resized, when I used the drag-and-drop-field. Can you replicate this as an issue?

PR #2547, opened 2019-07-18
Makes LDAP sync work.

### LINKED ISSUE #2354: Import LDAP Users
This doesn't seem to be the same issue as #14, but correct me if I'm wrong.

We've got Wekan set up and working with LDAP auth, but importing all of the users in the user base DN doesn't seem to work:

```
ldap-background-sync                              true
ldap-background-sync-import-new-users             true
ldap-background-sync-interval                     100
ldap-background-sync-keep-existant-users-updated  true
```

Is there a way to force the sync? Logs aren't showing any attempts at syncing, although any login related log messages are there and show no errors.

We're running this on CentOS 7.5 via snapd.

pshunter:
This is an attempt to fix #2354. While this code basically does what it is intended to, it still registers as an ugly hack and should be reviewed by an experienced Meteor developer. Specifically, I had to use SyncedCron.SyncedCron to access said package functions due to unknown import intricacies. Therefore I think #2354 should not be closed.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2547)


xet7:
@pshunter 

Your pull request has old Wekan commits, release tags etc, so revieving it is too hard, I would need new clean PR to see changes more clearly.

Please:
1. Clone your current forked repo
2. Delete your forked repo from GitHub, but not from your local copy.
3. Make new for of wekan/wekan repo.
4. Clone your new fork.
5. Add your changes to files.
6. git add --all, git commit
7. Submit new pull request.

PR #2577, opened 2019-08-01
Add settings field for user authentication field

Trekky12:
In the current implementation the user authentication is required to be with the 'uid' field. This PR adds a setting for this field.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2577)


xet7:
Please make new pull request to master branch. devel branch is old and not in use and will be deleted sometime.

PR #2607, opened 2019-08-10
 Bugfix: 2560, 2604 - enable mixed mode mongodb attachment and filesys

whowillcare:
fixed bug: unable to disable richer comment editor
add features:

On wekan master specifying ATTACHMENTS_STORE_PATH, it will try its best keeping original attachments, only newer attachments will be stored into specified path.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2607)


snap-src/bin/config
```diff
@@ -60,6 +60,10 @@ DESCRIPTION_WITH_API="Enable/disable the api of wekan"
 DEFAULT_WITH_API="true"
 KEY_WITH_API="with-api"
 
+DESCRIPTION_METEOR_SETTINGS="Meteor client settings, a JSON string, {\"public\":{}}"
```
xet7: @whowillcare 

Oh no I accidentally merged this. Please remove this METEOR_SETTINGS JSON string completely. In general, settings are normal strings for each settings, it is set as environment variable at serverside, then it is read server side by `wekan/server/something.js` where public variable is finally set. Then at `wekan/client/` it's read from public variable.

xet7: Hmm I think I'll already remove this myself.

whowillcare:
good. sry for that.

PR #2672, opened 2019-09-02
Devcontainer: Removed MAIL-Vars, Add PATH with meteor to Envionment

road42:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2672)


xet7:
Huh? Why?

road42:
Hi, @xet7 I updated the .devcontainer environment in vscode to the latest version and integrated your updates.

For development I removed the useless MAIL_VARS in docker-compose. Everyone can extend the docker-compose with the own configuration (https://code.visualstudio.com/docs/remote/containers#_extending-your-docker-compose-file-for-development)

I'm also writing a wikipage about using the devcontainer && debugging feature in vscode for local development.

.devcontainer/docker-compose.yml
```diff
@@ -29,9 +29,6 @@ services:
     environment:
       - MONGO_URL=mongodb://wekandb-dev:27017/wekan
       - ROOT_URL=http://localhost:3000
-      #- MAIL_URL=smtp://user:pass@mailserver.example.com:25/
```
xet7: Please add back mail settings. Mail also is tested in development.

.devcontainer/Dockerfile
```diff
@@ -149,6 +149,8 @@ ENV PATH=$PATH:$HOME/.meteor/
 # Copy source dir
 USER root
 
+RUN echo "export PATH=$PATH" > /etc/environment
```
xet7: Why replace all of environment?

.devcontainer/docker-compose.yml
road42: Of course it is. But there are some problems, when the variable is set in the `docker-compose.yml` which is in source code management.

- When I insert my data here, the file is modified and git sees it as changed and I have to be careful not to insert my personal data here.
- As soon as the variable is set (even with demodata) wekan tries so send the mail which gives an exception in the docker console, but not the mailoutput. So in development ist useful to see the mail, without sending it.

There is a way to set the variables without getting into the risk to commit personal data:

- We add the reference to an `.devcontainer/docker-compose.extend.yml` file to the `devcontainer.json`
- We add the extend-file to `.gitignore` so that nobody exposes personal data to github.
- An empty extend file only needs `version : '3.7'`

If someone needs to modify or overwrite the minimal-default config, he can add the required data to the extend file, which modifies the `docker-compose.yml`:

```yml
version: '3.7'
services:
  wekan-dev:
    # ADD Custom vars out of git
    environment:
      - MAIL_URL=smtp://user:pass@mailserver.example.com:25/
      - MAIL_URL=smtp://<mail_url>:25/?ignoreTLS=true&tls={rejectUnauthorized:false}
      - MAIL_FROM=Wekan Notifications <noreply.wekan@mydomain.com>
```

.devcontainer/Dockerfile
road42: you are right, I fixed that. Also the $HOME-var was invalid for root.

PR #2715, opened 2019-09-16
fix some issues found

whowillcare:
1. in richer editor, @user might not get pickup correctly, if it's formated
2. table content should have word-wrap
3. two-way hooks locking mechanism will fail sometime, therefore, change all comment insert or update to direct, which means it won't invoke any hook.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2715)


justinr1234:
@xet7 did you test this?

xet7:
@justinr1234 

Yes, running Wekan still did work for me. Of course, I can not guarantee that there is no bugs in Wekan.

justinr1234:
What is the repercussion of the change to the webhooks?

@xet7 @whowillcare

whowillcare:
there shouldn't be any. This only impacts two-way webhook, and it's fixing
issue not cause issue. The previous version might have race condition that
would cause rebounce between wekan and webhooks. Now, the fix is smoothing
the process and eliminating the rebounce.

xet7:
@justinr1234 

Two-way hooks were added at Wekan v3.36, they are at Admin Panel.
https://github.com/wekan/wekan/blob/master/CHANGELOG.md#v336-2019-09-05-wekan-release

This fix is related to locking and timing of two-way webhook, for error conditions about if webhooks are run too fast, so here is fixes to delays used etc, so the webhook works well enough.

In mongodb database changes, for example some migrations use "direct" mongodb way of making database changes, instead of normal way. End result is the same, changes are made to database.

PR #2756, opened 2019-10-11
Feature added and bug fixed

### LINKED ISSUE #2093: Wekan export internal server error
When I try to export a board I get an internal server error and a file with a size of 0 bytes

Tested in Firefox and Chrome

Running on an Ubuntu 18.04 Server

whowillcare:
Add Features: allowing lists to be sorted by modifiedAt when not in draggable mode

Buf Fix #2093: the broken should be prior to file attachment feature introduced, and tested export board is working

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2756)


whowillcare:
The origianl version of client/components/swimlanes/swimlanes.js has issue, also fixed in my pull request, somehow it got conflict. please check

models/boards.js
```diff
@@ -409,6 +409,21 @@ Boards.helpers({
   },
 
   lists() {
+    const enabled = Meteor.user().hasShowDesktopDragHandles();
```
xet7: showDesktopDragHandles in separate feature, it should not be used for sorting

models/cards.js
```diff
@@ -1695,6 +1695,23 @@ if (Meteor.isServer) {
...
+        // change list modifiedAt
+        const modifiedAt = new Date();
+        const boardId = list.boardId;
+        Lists.direct.update(
+          {
+            _id: list._id,
+          },
+          {
+            $set: {
+              modifiedAt,
```
xet7: I'm not sure what this does, sorting should not make changes to modified dates

models/swimlanes.js
```diff
@@ -174,6 +174,21 @@ Swimlanes.helpers({
   },
 
   lists() {
+    const enabled = Meteor.user().hasShowDesktopDragHandles();
```
xet7: hasShowDesktopDragHandles is separate feature, it should not affect any sorting, that should be separate option at lists menu

xet7:
With this current PR, it's not possible to create board, it keeps spinning and does not show new board.

xet7:
It would be easier to test if each fix or feature would be separate PR. Here many are mixed, so I'm not sure which changes belong to which fix or feature.

whowillcare:
try to separate ech pr, but it doesn't allow me to create multiple pr for
different reason somehow. I need to double check more.

xet7:
Maybe some of these could help:
https://stackoverflow.com/questions/8450036/how-to-open-multiple-pull-requests-on-github

models/boards.js
whowillcare: this is by logical, if user has chosen using draghandle to sort his list, then you don't need to add other sort method.

models/cards.js
whowillcare: this allows when user has chosen an EndAt date of a list, but didn't want to archive the list, we could change the list modifiedAt date to affect the sort order instead of pulling cards in each list to do a hug comparision, which is too costy.

whowillcare:
Made more changes enabling normal users can sort and filter lists under a board:
1. Previous setup of wekan, cannot filter list by title, so I added a function in "Filter" section
2. Adding new features allowing user to be able to sort lists by title, modfiedAt, and user manually-set orders decesnding or ascending
3. Allowing board admin to star list(s), so those lists will stay at the top, on mater what sort orders user has chosen.

whowillcare:
And these two conflicts will only be resolved by you. swimlanes.js is due to the previous developer's typo, so he made a varible defined a three times, which dosn't make sense atll.
for users.js, I added a extra field in profile to store user's prefenerce for sorting. Don't know why it cannot merge.

models/export.js
```diff
@@ -134,6 +140,11 @@ export class Exporter {
     const getBase64Data = function(doc, callback) {
       let buffer = new Buffer(0);
       // callback has the form function (err, res) {}
+      const tmpFile = path.join(
+        os.tmpdir(),
+        `tmpexport${process.pid}${Math.radom()}`,
```
xet7: Random ?

whowillcare:
It's for creating temporary file name, not a good solution, doesn't have a
built-in solution, so it's a nofrill solution

xet7:
Yes, but is it a typo? Is it really `Math.radom` ? Or should it be `Math.random` ?

xet7:
Creating board does not work.

2does-not-create-board

whowillcare:
Random indeed. I will take further look. Thx

whowillcare:
it looks like your compiled version has issue. hasShowDesktopDragHandles is
a helper function in User Collection indeed. and I can create board without
any issue, in testing environment or compiled version. I need to dig more
evidence.

xet7:
I tried building your fork of Wekan directly, and it seems to work. So it's all my fault for merging so that it got broken. Sorry.

Does your sorting code keep original order of lists?

Is sorting per list setting or per board setting?

whowillcare:
Per board

xet7:
Merged manually at https://github.com/wekan/wekan/commit/7d6d3af54a2fc1fb68634725eb754b22f02fd430

PR #2763, opened 2019-10-24
Enhancement: set card times more sensible using the 'Today' button in…

liske:
The 'Today' button in the datepicker component uses 08:00 as the default time value. IMHO it would be more sensible to use the current time for the *received*, *start* and *end* fields. For the *due* field it might be more reasonable to use a time in the afternoon (hard-coded 17:00).

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2763)

### DIFF
--- a/client/components/cards/cardDate.js
+++ b/client/components/cards/cardDate.js
@@ -105,7 +105,7 @@ Template.dateBadge.helpers({
 // editCardReceivedDatePopup
 (class extends DatePicker {
   onCreated() {
-    super.onCreated();
+    super.onCreated(moment().format("YYYY-MM-DD HH:mm"));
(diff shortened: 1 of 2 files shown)


saschafoerster:
@liske Thanks for changing this behavior, which I am happy to see. Only the hardcoded timing for due date doesn't really fit our needs. Due time is for us a signal to do a job today, so 9:00 would fit much better to our needs. 

That's why we set NOTIFY_DUE_AT_HOUR_OF_DAY=8 on the early morning, which is also the default setting for e-mail notifications. 

Maybe instead of hard coding you could create a setting like the notify-due-at-hour? Thanks!

Issue #2772, opened 2019-10-30
New List item moved from right to left


derbolle:
After our Wekan Update from Version 3.44 to 3.49 we noticed that the New-List Field moved from the right all the way to the left. 

After reverting back to Wekan 3.44 the Behaviour switched back to like it was before

**Server Setup Information**:

* Did you test in newest Wekan?: yes

* Wekan version: 3.49

* Operating System: Docker Swarm on Debian 10

* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): Docker Swarm with docker-compose file

* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): Traefik reverse Proxy

* Node Version: 8.16

Mongo Version 4.0.11

* Wekan only works on newest desktop Firefox/Chromium/Chrome/Edge/Chromium Edge and mobile Chrome. What webbrowser version are you using? Firefox 69 and Chrome

jhgovsi:
Can confirm the same!

xet7:
@derbolle 

There is no option. I will try to revert this.

Issue #2771, opened 2019-10-30
Card and List move no longer possible


derbolle:
## Issue
Tonight we updated Wekan 3.44 to 3.49. After the Update we noticed that we cannot move cards or lists anymore. After reverting back to 3.44 it worked again.

**Server Setup Information**:

* Did you test in newest Wekan?: yes

* Wekan version: 3.49

* Operating System: Docker Swarm on Debian 10

* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): Docker Swarm with docker-compose file

* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): Traefik reverse Proxy

* Node Version: 8.16

Mongo Version 4.0.11

* Wekan only works on newest desktop Firefox/Chromium/Chrome/Edge/Chromium Edge and mobile Chrome. What webbrowser version are you using? Firefox 69 and Chrome

jhgovsi:
I can confirm the same. Workaround (I hope this is not intended feature from now on):

**Change Settings -> Show desktop drag handles**

Otherwise, card dragging is disabled on desktop!

NoxInmortus:
Hello, I do reproduce the issue, and I confirm it is resolved by enabling "Show desktop drag handles".
Even if it works after that, there is now a sort of cross on every card, and it's reducing card names visibility.

Also, I can drag and drop only clicking on the cross, not anywhere on the card like before. I understand this feature (to not move cards by accident), but something feel a bit like a step back :/

alexanderadam:
Is there a PR for this fix in the work?

xet7:
@alexanderadam 

There will not be PR, because I'm maintainer of Wekan, and will commit fix directly to repo master branch when it's ready.

alexanderadam:
Okay, just to be sure: so the only possibility is to get notified by subscribing this issue, right?

jhgovsi:
any news on progress of fixing this rather critical (regarding the user acceptance within our team) issue?

bkraul:
@jhgivsi It's a little bit of bad timing. I was actually in the middle of introducing wekan to our team, but the  moving the cards is central to the usability of the project...Thanks for looking into this, @xet7

derbolle:
Hello jhgovsi, 
As stated by xet7 the 3.49 should Work. If not: we Use version 3.44 wäre dragging works as intended


> Thanks for testing, I will look how it works on those browsers.
> 
> You are running newest commits from master branch.
[...]

PR #2774, opened 2019-10-30
The last merge didn't seem to happen correctly?

whowillcare:
Added some recent fix of typos

including radom to random

and add feature that user modifies card's startat, dueat, endat and then it will affect the list's last ModifiedAt timestamp, which is a compromised way to sort lists without aggregation all cards that could be too expensive, and current meteor version even didn't support certain mongodb aggregation syntax

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2774)

### DIFF
--- a/models/cards.js
+++ b/models/cards.js
@@ -1699,9 +1699,8 @@ if (Meteor.isServer) {
       if (list) {
         // change list modifiedAt, when user modified the key values in timingaction array, if it's endAt, put the modifiedAt of list back to one year ago for sorting purpose
         const modifiedAt = new Date(
-          new Date(value).getTime() - (action === 'endAt')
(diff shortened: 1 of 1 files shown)


xet7:
I did already change `radom` to `random`.

This pull request has changes that are already in Wekan. Merging this would mess up release tags etc.

Please close this pull request, rename your local repo directory, delete your fork, make new fork of Wekan, add your new changes to there, making new pull request.

Remaining issues are:
- #2772 where I did succeend in reverting your changes that new list button is at left. Remaining is fixing that new lists are created to right, not left.
- #2771 card and list move is not possible

xet7:
Actually, just a moment, I'll test this.

PR #2854, opened 2019-12-17
Allow card & checklist API creation for authorized board members

Robert-Lebedeu:
Use Utils function to check user membership to board and his role.
Fix checkBoardAccess authentication check. (User is board member check was broken)

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2854)

### DIFF
--- a/models/cards.js
+++ b/models/cards.js
@@ -2003,8 +2003,15 @@ if (Meteor.isServer) {
     req,
     res,
   ) {
-    Authentication.checkUserId(req.userId);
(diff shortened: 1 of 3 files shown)


xet7:
Please include REST API commands that you used to test (for example with curl/php/python/js) that:
- NoComments user can not get content of any comments and activities via REST API
- CommentOnly user can only add comments, and not change anything else at board

Related https://github.com/wekan/wekan/wiki/REST-API-Role

PR #2897, opened 2020-01-22
Implemented Nextcloud OAuth2 Hack References Issue #1874

bogie:
In reference to: https://github.com/wekan/wekan/issues/1874#issuecomment-475640011

I finally got my dev environment running, applied the change and tested:

Login works, however the profile data is not being applied. Is there some sort of additional logic we need to implement in [oidc_server.js](https://github.com/wekan/wekan/blob/master/packages/wekan-oidc/oidc_server.js) similarly to ldap auth?

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2897)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -342,6 +342,31 @@ services:
       # Tthe claim name you want to map to the email field:
       #- OAUTH2_EMAIL_MAP=email
       #-----------------------------------------------------------------
+      # ==== OAUTH2 Nextcloud ====
+      # 1) Register the application with Nextcloud: https://your.nextcloud/settings/admin/security
+      #    Make sure you capture the application ID as well as generate a secret key.
+      # 2) Configure the environment variables. This differs slightly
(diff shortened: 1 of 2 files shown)


xet7:
@bogie 

I think those details depends in what settings those should be mapped to:
https://github.com/wekan/wekan/wiki/OAuth2#3-snap-settings-change-to-it-from-above-client-id-secret-server-url-and-web-origin-namespace-for-rules-function-above

bogie:
Hmm, it gets correctly mapped to the serviceData object and also to

`options: { profile: profile }`

however it does not appear to get applied to the user's profile correctly. I do not understand where [OAuth.registerSerivce()](https://github.com/wekan/wekan/blob/master/packages/wekan-oidc/oidc_server.js#L3) gets called from.

Maybe the problem is due to the fact that the admin user was created by oauth2. I will do some more testing.

PR #2898, opened 2020-01-23
fixed board export with attchment

izadpoor:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2898)


izadpoor:
please review

Issue #2901, opened 2020-01-27
Proposition of Dockerfile simplification with docker swarm secret management


mduret:
The main idea is to separate the build and the run part:  [link](https://docs.docker.com/develop/develop-images/multistage-build/)

And with the rewrite of the Dockerfile, it's easy to put docker swarm secret management for ENVVAR which contain password: [link](https://docs.docker.com/engine/swarm/secrets/)

PR #2902, opened 2020-01-27
simplification of Dockerfile with docker swarm secret management for …

mduret:
…password ENVVARs

Hello,

it is a proposition about this: https://github.com/wekan/wekan/issues/2901

I made some modifications on Dockerfile for my use of wekan in my company.

I hope it can help.

Have a good day,

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2902)


xet7:
Pro:
- Docker image size changed from 217.1 MB to 145.1 MB

Con:
- Dockerfile does not have default values set, for example `WITH_API=true` so API and Export Board does not work.
- Number of vulnerabilities changed from 0 to 49 vulnerabilities, as detected by Quay, because changes in base image etc.

Are you planning on adding more fixes to this?

mduret:
Thanks for looking at my proposal,

I will fix theses vulnerabilities in few days: [link](https://quay.io/repository/wekan/wekan/manifest/sha256:3a658985750b14164102c1243d1cc977a026b05fa5677583f152f4c9fd1e9fb7?tab=vulnerabilities)

About the default values, I think that it is not the best place to put default value in Dockerfile but maybe I am wrong because I don't know very well the wekan project. I suggest to put default values in the configuration management code section: in the aim that the default values do not depend of a particular deployment way. If you think it is make sens, I can take a look and propose a solution.

Have a good day,

xet7:
If you are still working on this, please add a new pull request.

PR #2907, opened 2020-01-29
Layout changes and implementing the new design for the card

2020product:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2907)


2020product:
Hi
I created the pull request "Layout changes and implementing the new design for the card"
I hope it can help.
Please check this pull request.
Thanks and regards.

client/components/main/editor.js
```diff
@@ -120,6 +120,7 @@ Template.editor.onRendered(() => {
     autosize($textarea);
     $textarea.escapeableTextComplete(mentions);
   };
+  Meteor.settings.public.RICHER_CARD_COMMENT_EDITOR = false;
```
xet7: This line should be removed, it's setting it to always false.

docker-cleanup.sh
```diff
@@ -0,0 +1,8 @@
+sudo docker kill $(docker ps -q)
```
xet7: This file should be removed. Docker cleanup is not in any way related to layout changes.

public/api/wekan.html
```diff
@@ -1458,12 +1458,12 @@
   opacity: 0.5;
 }
 </style>
-    
+
```
xet7: Changes to this file should be removed. Wekan API docs are generated with script, they should be not modified manually.

rebuild-wekan.bat
```diff
@@ -1,10 +1,5 @@
 @ECHO OFF
 
-REM NOTE: You can try this to install Meteor on Windows, it works:
```
xet7: These changes are in latest Wekan git repo, added by me, you should not remove them. You should check that you are up-to-date with latest changes at git repo.

xet7:
What is that wekan-mongodb submodule? You should remove it.

xet7:
In general, with submodules make it difficult to clone repo on Windows. That's why they are not used in Wekan repo.

xet7:
Here's how to remove commits:
https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-pull-request

client/components/cards/cardDetails.jade
```diff
@@ -143,85 +172,85 @@ template(name="cardDetails")
...
             a.fa.fa-times-thin.js-close-inlined-form
         else
-          a.js-open-inlined-form
+          a.description-item.add-description.js-open-inlined-form
             if getDescription
               +viewer
                 = getDescription
             else
-              | {{_ 'edit'}}
+              | {{_ 'Add a more detailed description'}}
```
xet7: You have not yet added this translation also to `wekan/i18n/en.i18n.json`.

xet7: Although I don't know why this should be changed?

xet7: I think this change should be removed.

2020product: I will fix it

xet7:
Your changes permanently hides Wekan features on card. Hiding some feature should be optional per board, and possible to be done only by BoardAdmin users.

2020product:
https://youtu.be/7KG-tQ_QP_4
I have changed this features on card.
"Your changes permanently hides Wekan features on card. Hiding some feature should be optional per board, and possible to be done only by BoardAdmin users."
so I can not understand well

xet7:
Hmm, just a moment, I'll think what would be the best way to implement this.

xet7:
@2020product 

I'm currently making some fixes to this, after that I will merge this.

2020product:
So what can I do?

xet7:
@2020product 

You don't need to do anything, I will finish this today.

2020product:
Okay, thanks for your time

xet7:
I don't have enough time to finish those changes today.

You should make these changes:

At cardDetails.js , have these for checking do those exists:
```
  memberSelected() {
    if (this.getMembers().length === 0) {
      return false;
    } else {
      return true;
    }
  },

  labelSelected() {
    if (this.getLabels().length === 0) {
      return false;
    } else {
      return true;
    }
  },

  assigneeSelected() {
    if (this.getAssignees().length === 0) {
      return false;
    } else {
      return true;
    }
  },
```
This way you can show those on card that have some content, but not show empty one, like this assignee, at cardDetails.jade:
```
    if assigneeSelected
      .card-details-item.card-details-item-assignees
        h3
          i.fa.fa-user
          card-details-item-title {{_ 'assignee'}}
        each getAssignees
          +userAvatarAssignee(userId=this cardId=../_id)
          | {{! XXX Hack to hide syntaxic coloration /// }}
```
You should have paperclip Font Awesome icon on adding attachment.

If you hide something, like Subtasks, you should have it visible only when there is some subtasks, and have adding it at card hamburger menu.

xet7:
When you move something from card to card hamburger menu, you should have original Font Awesome icons there, like this:
```
  if canModifyCard
    unless currentUser.isWorker
      hr
      ul.pop-over-list
        li
          a.js-members
            i.fa.fa-users
            | {{_ 'members'}}
        li
          a.js-labels
            i.fa.fa-tags
            | {{_ 'labels'}}
        //li
        //  a.js-attachments
        //    i.fa.fa-paperclip
        //    | {{_ 'attachments'}}
        li
          a.js-custom-fields
            i.fa.fa-list-alt
            | {{_ 'custom-fields'}}
        li
          a.js-received-date
            i.fa.fa-sign-out
            | {{_ 'card-received'}}
        li
          a.js-start-date
            i.fa.fa-hourglass-start
            | {{_ 'card-start'}}
        li
          a.js-due-date
            i.fa.fa-sign-in
            | {{_ 'card-due'}}
        li
          a.js-end-date
            i.fa.fa-hourglass-end
            | {{_ 'card-end'}}
        li
          a.js-spent-time
            i.fa.fa-clock-o
            | {{_ 'card-spent'}}
        li
          a.js-set-card-color
            i.fa.fa-paint-brush
            | {{_ 'setCardColorPopup-title'}}
```

xet7:
But then we get into the problem, that with these changes card is less crowded, but card hamburger menu is more crowded.

At Trello, also those Members/Labels etc are hidden at start, when nothing is added to card yet at start.

xet7:
I think there is no need to add more translated strings, you should revert your translation changes.

Above I have changed current ones to use even shorter translation strings.

xet7:
There is also now the question, that when more in moved from card to card hamburger menu, it adds requirement for more clicks to use that feature, and makes using Wekan slower.

2020product:
Hi @xet7 
How are you?
I have changed the following code in carddetails.js such as // if assigneeSelected    
but I need to clear the requirements.
"But then we get into the problem, that with these changes card is less crowded, but card hamburger menu is more crowded.
At Trello, also those Members/Labels etc are hidden at start, when nothing is added to card yet at start."
"There is also now the question, that when more in moved from card to card hamburger menu, it adds requirement for more clicks to use that feature, and makes using Wekan slower."
Thanks

xet7:
@2020product 

With your changes, you did hide features like Subtasks permanently. Some Wekan users still use that feature.

If you would like to hide some feature, there should be on-off switch to turn that feature back on, to Subtasks being visible. Like there is switch for system messages.

xet7:
It's not OK to disable some existing feature permanently.

xet7:
@2020product 

By default, all existing fields at card (like Received/Start/End/Due Date/etc) should be visible like it is currently before your changes. You should add code for hiding those fields, with slider or at menu checkbox what is visible, and save that to database to board settings.

Only BoardAdmin user should be able to see those switches, to switch off Received Date and other fields. You check in .jade files those permissions, like ` if currentUser.isBoardAdmin`. You can search for similar code with `cd wekan && ./find.sh isBoardAdmin`.

2020product:
so you mean I have to make the button to hidden or display the icon such as start, end and so on?

xet7:
@2020product 

Yes.

2020product:
Screen Shot 2020-01-31 at 5 38 11 AM

I have to add the "Date Settings"  to settings menu like the above image

2020product:
please let me know about this

xet7:
In Dates Settings, please change icon `fa-sitemap` to `fa-calendar`. Otherwise, it's OK. Thanks!

xet7:
For each date setting like Start Date, you can have same icons as there currently is on card.

2020product:
yes, understand

zmaren:
Hello @xet7, I'm Zeljko from crewnew.com, Kaspar talked with you. Can you explain to me what are the next steps for us to have this change on the server? You will build snap and we just need to update, or we need to do something else? I installed Wekan through the snap package.

2020product:
Hi @xet7 
Sorry, but I closed the pull request.
because of @kaspar
good luck for this task

Issue #796, opened 2017-02-08
Add feature: Vote on cards, number of votes, average as result


xet7:
_From @xet7 on December 14, 2016 19:24_

Original issue: https://github.com/wekan/wekan/issues/130

_Copied from original issue: wefork/wekan#46_

killua-eu:
Added a bounty on this issue. Please see https://www.bountysource.com/teams/wekan/issues ...
I'd prefer if both +/- wotes are allowed. Having alternative + while not clotting the user interface would be absolutely awesome.

killua-eu:
Having +/-/whatever votes on comments would be nice too.

xet7:
I don't know how to reach someone that has youtube video of vote feature:

https://youtu.be/8jQPqcNgKA0

onthisearth:
@xet7 I contacted the Youtuber. Let's see, if he'll contact me :)

d3dbit:
Hello all,
I think that the idea to vote and display the number of votes in the cards would be a nice feature.
Thanks for your effort.

lynchmaniac:
Hi,
is anyone have some news about this feature ?
The feature on the video looks great.

pablanopete:
@xet7 did you ever get in contact with the youtuber? maybe this would be worth a Gitcoin bounty; but I don't know how to define that scope in terms of how much time it could take.

NicoP-S:
@xet7 I think this can be closed.

PR #2994, opened 2020-04-07
Add Card vote Feature

NicoP-S:
I added a card voting feature. #796 
Can someone please review and test it?

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2994)


client/components/cards/cardDetails.styl
```diff
@@ -330,3 +330,11 @@ card-details-color(background, color...)
 
 .card-details-indigo
   card-details-color(#4b0082, #ffffff) //White text for better visibility
+
+.voted
+  opacity: .7
+.vote-title
+  display: flex;
```
xet7, line 337: Please remove `;` from end of this, stylus does not have those.

client/components/cards/cardDetails.styl
```diff
@@ -330,3 +330,11 @@ card-details-color(background, color...)
...
+  justify-content: space-between;
```
xet7, line 338: Please remove `;` from end of this, stylus does not have those.

client/components/cards/cardDetails.styl
```diff
@@ -330,3 +330,11 @@ card-details-color(background, color...)
...
+.vote-result
+  display: flex;
```
xet7, line 340: Please remove `;` from end of this, stylus does not have those.

models/cards.js
```diff
@@ -304,6 +304,38 @@ Cards.attachSchema(
       optional: true,
       defaultValue: '',
     },
+    vote: {
```
xet7, line 307: For these new database fields, please also add migrations to Cards table/collection to bottom of `wekan/server/migrations.js`.

xet7:
Optional if you like, not required:
You could export Trello JSON, and look can you import votes from Trello to Wekan here:
https://github.com/wekan/wekan/blob/master/models/trelloCreator.js#L258

xet7:
@NicoP-S 

I did add most of those changes myself, and merged your vote feature. Thanks!

I did not add import from Trello.

FuXXz:
I am very grateful for such efforts and that you have added it. It is really very nice.
Also for the reason that I can not contribute much because i  have only HTML and CSS knowledge :(

But the first tests showed that the feature is missing something.

Colleagues argued very quickly who voted yes or no. That means trouble :)

So, without any information about who voted, this function is severely restricted

xet7:
@FuXXz 

When I'm voting for president, my vote is not public. Similarly in Wekan, I don't know why it should be known who voted what.

xet7:
@FuXXz 

Your colleagues should just stop arguing and accept the results of voting :)

NicoP-S:
Maybe it would be good to add a checkbox if the vote is public or not.
This should not be changeable after the vote Haß started.

xet7:
@NicoP-S 

Yes it would be nice if you have time to add it like this, default unchecked:
```
[_] Public
```

FuXXz:
@NicoP-S  unbelievable good and fast

Issue #2995, opened 2020-04-08
Filter by Assignee


darren-teo:
I guess this is more of a feature request. I'm able to filter cards by person if he/she is a member of the card, however I'm not able to filter if that person is an Assignee in the card.

Can the filtering be extended to assignee as well?

**Server Setup Information**:

* Wekan version: 3.86
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): docker
* MongoDB Version: 4.2.5

daniel-eder:
Seems like a vital feature to me, implemented in https://github.com/wekan/wekan/pull/2997

daniel-eder:
@xet7 this can be closed. Thanks!

darren-teo:
Thank you. I can see the commit in 3.91 onward.

PR #2997, opened 2020-04-08
Add filter option for assignee

daniel-eder:
Implements the filtering for assignees requested in https://github.com/wekan/wekan/issues/2995.
Works like the member filter, but filters for assignees instead.

No filter:
No Filter
Filtering assignee: 
Assignee Filter

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/2997)


TheMasterFX:
Great! Is it possible to use the according string for Assignee for the language? Thank you so much

Issue #252, opened 2015-08-29
Feature Request: Option for seeing all public boards on All Boards page


flippingtables:
For someone that is hosting an instance of Wekan as an internal Kanban board, it would be nice for all users to be able to see all public boards.

The way it works now, is that if you want to see the boards of other users, you will need to either send them the public link, or to add each user to the board.
It would be nice to be able to see all public boards directly in the All Boards page when you log in.

yipcma:
Up this feature. or else what makes public public other than access rights?

yipcma:
Please prioritize this. It would be very difficult to use wekan in an organizational setting without such feature.

flippingtables:
Thats what I'm thinking. Actually, it may even be better to implement some sort of GROUP functionality. 

E.g.:
Organization hosts a Wekan instance internally
Org has several internal groups.
Users are added to groups.
Boards can be declared to belong to a Group.
All users in the same group can view the group.

This would eliminate the need to manually add each individual user that should have access to the board.

Roitoh:
Would also like to see this implemented. We are testing Wekan as an organizational program at work and it'd be nice to not have to specify users every time I make a board.

AndyatFocallocal:
bump

simopaasisalo:
A simple mid-level publicity options would be enough for our use case: every registered user should be able to see every public board, but the boards should not be visible to outside users like the current `Public`-setting

koorg:
I think the way "public" boards are not seen by *any* existing user (they need to be invited to see the board), is more a bug than a feature, it doesn't "feel" public... We don't need groups at the moment, only "public" (seen/editable by any user) boards and "private" boards, as it is meant to work at the moment :)
Auto-adding is perhaps not the best way to implement it but we're looking into scripting auto-add for public boards as a dirty workaround... which is really not the way it should be...

jacques-:
We have a "master" board and just linking to all other boards from that master board - it is super dirty.

A page with list of public boards would be so much better.

TiibCD:
UP +1 to this feature. I agree I think the way public boards work at the moment doesn't feel "public" like we are used to.
The fact that we can share a public board via a link is great but, as a logged user to see all the public boards and my private board seems logic to me.
Any idea if someone took this feature on ?

bboy-space:
Up ?

bboy-space:
On the user's homepage of wekan, I suggest to he sees all public boards (with a limit and the usual "expand arrow" .. ),
[...]

PR #3008, opened 2020-04-11
Public boards overview

NicoP-S:
Added a public boards overview page #252 
Used the same components like the 'home' route with a different board query.

Please have a look on the changed boards publication.
server/publications/boards.js 

image

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3008)


TheMasterFX:
Wow, I waited so long for this! Thank you

PR #3014, opened 2020-04-13
hide password auth with PASSWORD_LOGIN_ENABLED variable

salleman33:
Hi, 
Here is a workaround to disable/hide the password login form. It coult be use when ldap,cas or Oauth2 methods are used.
I looked to completely disable password meteor account but it seems pretty difficult... 

Regards

### DIFF
--- a/client/components/main/layouts.js
+++ b/client/components/main/layouts.js
@@ -31,6 +31,11 @@ Template.userFormsLayout.onCreated(function() {
       return this.stop();
     },
   });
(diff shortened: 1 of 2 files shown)


xet7:
@salleman33 

Please add settings to Docker/Snap/Sandstorm/sh/bat like at https://github.com/wekan/wekan/commit/9d13001b903f9ec50f5fa3a4bdbacae32b27ac65

PR #3024, opened 2020-04-17
Improve card layout on mobile devices

marc1006:
Before:
before

After:
after

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3024)


marc1006:
Thanks.

Issue #2535, opened 2019-07-09
Feature: Starting day of the week (Sunday vs Monday) 


ykorzikowski:
There is currently no way to set the starting day of the week (Sunday vs Monday) and it defaults to Sunday on date edit view. It would be great to add an option to configure this. 

See: 

![Bildschirmfoto 2019-07-09 um 20 13 10](https://user-images.githubusercontent.com/20556547/60912609-0aadf200-a286-11e9-9f4f-1f0ef82410bc.png)

gruessung:
+1

xet7:
Target bounty 200 euro, then I will start implementing this.

https://www.bountysource.com/issues/76694702-feature-starting-day-of-the-week-sunday-vs-monday
https://wekan.team/bounties/index.html

saschafoerster:
We also had several issues with items accidentally put on a Sunday, because they thought it is Monday in the first line. I think this is not really a feature to be targeted by money, but rather a bug to be solved by anyone who is able. But anyway, we should support Wekan and it's development. :) So here is my +1 for this!

ykorzikowski:
Just made some research. Maybe https://www.npmjs.com/package/js-datepicker => setting

```
  startDay: 1, // Calendar week starts on a Monday.
```

will do the trick?

xet7:
Sure, with enough Wekan contributors I would get pull requests to all issues and Wekan would be finished :) Anyone can help and add pull requests.

saschafoerster:
@ykorzikowski Okay, let's try the commit here? :) https://github.com/wekan/wekan/blob/devel/client/lib/datepicker.js
(I really have no clue, where to add the option. Maybe here?)

```
 onRendered() {
    const $picker = this.$('.js-datepicker').datepicker({
      todayHighlight: true,
      todayBtn: 'linked',
      language: TAPi18n.getLanguage(),
      startDay: 1,
    })
```

xet7:
@saschafoerster 

I did try adding that `startDay: 1` but it did not make any difference.

saschafoerster:
So I warted to give 15$ via bountysource to this feature request, my PayPal-account shows it as paid, but I am not on the list as a backer... :( https://www.bountysource.com/issues/76694702-feature-starting-day-of-the-week-sunday-vs-monday

xet7:
@saschafoerster 

When did you add that bounty? Maybe it's not immediately visible, and takes some days.

saschafoerster:
[...]

Issue #1574, opened 2018-04-05
Wekan Design: Comment section on card, etc


TiibCD:
Hi guys!
I have another request! When you have some checked lists with items on a card, the comments in the activity section are hard to notice if you don't scroll down. (see screenshots below)

**Request**
* Could it be possible to have an option to hide the Attachments on cards ? 
* Or could it be possible to show the number of comments beside the "Activity" title ?

Thanks for listening and for the great work

**Server Setup Information**:

* Operating System: **Ubuntu**
* Deployment Method: **VirtualBox appliance**
* Wekan version : **v0.78**
* Node Version: **v8.9.3**

![to do - wekan - mozilla firefox](https://user-images.githubusercontent.com/24451989/38375038-847fd7de-38f5-11e8-915f-ee3f15c9105e.jpg)

![to do - wekan - mozilla firefox_2](https://user-images.githubusercontent.com/24451989/38375046-88886e2c-38f5-11e8-8dd4-2e1346ed5ca4.jpg)

edouardpa:
Hi everyone !

I would have loved it too, if we could hide Attachements when it's empty.

Thanks for your work !

d3dbit:
+1 to @eduardpa 's comment

It could be a nice idea to hide 'Attachements' section when it is empty

Jay2k1:
I would actually prefer to have the comments/activity section in a second pane to the right, so that the card details canvas has two panes, the left one like it is now (minus activity) and then the activity feed to the right of that. this way, you wouldn't have to scroll down for comments (which is super slow for me btw, looks as if there's some JS "smooth scrolling" foo involved).

FuXXz:
The layout is unfortunately a static neck break. In the 21st century, it should be more flexible, fluid, and intuitive.
It could be that easy
With a split layout, which I can adjust individually. With the bar in the middle I can slide flexibly and choose the relationship between comments and content
Wekan should be able to do that
![Anmerkung 2020-04-03 193553](https://user-images.githubusercontent.com/9010594/78389168-9325ad00-75e2-11ea-93df-3a2c9e932609.png)

xet7:
@FuXXz

I tried a little about that. Currently it's not flexible resizeable at middle, and opens on left only for cards at left so would need to be changed to a popup. Here is current In-Progress code as a starting point for someone interested in implementing this. 

[diff.txt](https://github.com/wekan/wekan/files/4432598/diff.txt)

[...]

PR #3032, opened 2020-04-21
Implement option to change the first day of the week in user settings

marc1006:
Implements https://github.com/wekan/wekan/issues/2535.

Screenshots:
screen
screen_2

It is a first version and hopefully it makes sense what I have done :)

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3032)

### DIFF
--- a/client/components/main/popup.styl
+++ b/client/components/main/popup.styl
@@ -135,6 +135,10 @@ $popupWidth = 300px
   margin-bottom: 8px
 
 .pop-over-list
+  li
+    display: block
+    clear: both
+
   li > a
     clear: both
     cursor: pointer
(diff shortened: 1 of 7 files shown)


marc1006:
It might also be useful to change the default value from Sunday to Monday.

xet7:
@marc1006 

Yes, please change default value to Monday.

marc1006:
@xet7 Okay, I will add a separate commit for changing the default.

xet7:
@marc1006

Those are all different settings, you should not combine them.

marc1006:
Okay, but right now, as soon as you apply a change, the "Change settings" popup will exit. I'm not sure whether this behavior is always intended.

xet7:
@marc1006 

Yes, currently it does Popup.Close.

Wekan is going to design direction where most things are fullscreen https://github.com/wekan/wekan/issues/1574#issuecomment-609095074 . So if you like, you can combine that one popup menu, or all of those submenus to one big fullscreen settings popup, that has one Apply. If you make fullscreen popup, then you need to move also all those helpers etc also.

xet7:
For mobile view of settings, there is isMiniScreen checks.

```
~/repos/wekan $ ./find.sh isMiniScreen
```

xet7:
Sure starting first with only combining that one popupmenu to use one Apply is smaller first step, and a good place to start.

PR #3038, opened 2020-04-22
Fix getStartDayOfWeek function

marc1006:
In case profile.startDayOfWeek is 0 it's evaluated to false and 1 is returned.
Let's fix this by differentiating between undefined and an actual value.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3038)

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -530,7 +530,8 @@ Users.helpers({
 
   getStartDayOfWeek() {
     const profile = this.profile || {};
-    return profile.startDayOfWeek || 1;
+    // default is 'Monday' (1)
+    return profile.startDayOfWeek ?? 1;
   },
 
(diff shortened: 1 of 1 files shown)


models/users.js
```diff
@@ -530,7 +530,11 @@ Users.helpers({
 
   getStartDayOfWeek() {
     const profile = this.profile || {};
-    return profile.startDayOfWeek || 1;
```
boeserwolf, line 533: You can also use the new `??` operator since yesterday, see: https://github.com/tc39/proposal-nullish-coalescing, like so: `return profile.startDayOfWeek ?? 1;`

marc1006: Changed - thanks for the feedback!

Issue #3047, opened 2020-04-25
Feature Request: Total votes / Total of possible voters


helioguardabaxo:
Would be interesting add total votes per total possible voters on card(total members on board).

Something like this:

![Screenshot_2020-04-25 Kanban -(1)](https://user-images.githubusercontent.com/24301085/80285560-4e31f980-86fc-11ea-81e2-c1c03a27aac1.png)

helioguardabaxo:
@NicoP-S 
What do you think?

NicoP-S:
I added this Feature.
There is also another change which allows non Board members of public Boards to vote. If this is enabled on a public Board the Total voters are hidden.

helioguardabaxo:
Great! :D
Vote in public board is very useful.
What do you think about to add the "finish voting" feature? Maybe could have a date/time to finish. This is necessary when we have a due date to resolve a task/card.
Nowadays we just have "delete voting and all votes". If we deleted a voting, we lose all the votes.

In this scenario, in "card actions" we could have "voting in card" item. 
In the pop-over we could have "voting question:text", "public vote:check", "due date:date/hour", "save:button" and "delete:link".

And pop-over delete confirmation with this: 
- Title: Delete voting and all votes?
- Message: All votes will be removed from the voting and you won't be able to re-open the voting. There is no undo.
- "delete:button"

helioguardabaxo:
@NicoP-S 
I don't know if you were notified for my last message suggesting some improvements. I'm citing you here. :D

NicoP-S:
@helioguardabaxo yes i have noticed it. I started on this too. I am struggeling with the date picker. Try to commit it later on

NicoP-S:
@helioguardabaxo this Feature does now exist. can you close the issue?

PR #3048, opened 2020-04-26
additional vote features

NicoP-S:
- adds the total vote thing #3047 
- adds option "Allow non board members to vote" ( for public boards )
- display vote result on minicard

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3048)


xet7:
Hi,
please change these in all code variables and English translation:
- Public vote => Show who voted what
- Allow non board members => Allow anonymous vote on public board

Also, currently there only can be one anonymous vote on public board. Even when accessing board from different IP address to set second anonymous vote, it modifies that original first anynymous vote. This is because for Meteor, there is only one anonymous user, it does not depend on IP address. For anonymous vote, there is no user listed in for or against. There could be, in theory, IP address of the anonymous voter.

NicoP-S:
@xet7 ah damm. I missed to query of the User is logged in on public Board votes...
I will fix this asap. 
Naming of variables and text will be change too.
Thanks

xet7:
If someone does anonymous vote on public board, it could be useful that they could input their name to that vote. For example:
(anonymous vote font awesome icon here) Alex (tooltip: ip address)

xet7:
actually, ip address would better be visible, not in a tooltip

xet7:
Hmm, I don't know. Maybe visually better still in a tooltip.

xet7:
Oops sorry I merged it early. Anyway, please add additional changes in new pull request.

xet7:
I think I will not revert this, it's just missing some additions.

xet7:
Hmm, actually, I need to revert this, because it makes changes to database, and having this in-progress version would bring migrations issues.

NicoP-S:
@xet7  my idea behind "allow non board members" was to allow logged in users to vote on public boards. I don't know how to handle not logged in User (anonym).

xet7:
@NicoP-S 

Then you should check that user is logged in, so anonymous can not vote:
```
if (Meteor.user()) {
```

PR #3061, opened 2020-04-28
Fix getStartDayOfWeek once again :)

marc1006:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3061)

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -530,8 +530,11 @@ Users.helpers({
 
   getStartDayOfWeek() {
     const profile = this.profile || {};
-    // default is 'Monday' (1)
-    return profile.startDayOfWeek || 1;
+    if (typeof profile.startDayOfWeek === 'undefined') {
+      // default is 'Monday' (1)
+      return 1;
+    }
+    return profile.startDayOfWeek;
   },
 
   getTemplatesBoardId() {


sandstorm.js
```diff
@@ -22,7 +22,7 @@ const sandstormBoard = {
 
 if (isSandstorm && Meteor.isServer) {
   const fs = require('fs');
-  const Capnp = Npm.require(`capnp`);
+  const Capnp = Npm.require('capnp');
```
xet7, line 25: How this is related to getStartDayOfWeek ?

marc1006: That's another commit and unrelated.

xet7: Please remove this change.

marc1006: Shall I create a separate pull request for the separate commit "Prettifier fixes"? Thought this should be good enough to keep the changes in two separate commits :) Will change.

xet7: Do not create any prettifier fixes. Very likely it breaks generating OpenAPI docs.

marc1006: Done.

xet7: Also, please revert all changes related to new Javascript syntax https://github.com/wekan/wekan/pull/3034 like ?? because if that is used in any Javascript file, not all OpenAPI docs are created https://github.com/wekan/wekan/commit/9ae20a3f51e63c29f536e2f5b3e66a2c7d88c691

marc1006: This time I did not use the operator `??` for exactly this reason.

marc1006: > Do not create any prettifier fixes. Very likely it breaks generating OpenAPI docs.

Shouldn't we add an automatically check for that? (e.g. by using Travis-CI for that)

xet7: Maybe there could be a check that OpenAPI docs include all parts. Wekan does use Travis-CI but it only checks some syntax.

PR #3064, opened 2020-04-28
Missing query for logged in users

NicoP-S:
Here are the missing parts from #3048 / #3047
On card with "allow non board members to vote" is now a query if the user is logged in.

There is also the option to set an end date
image

@xet i have a small problem with the datepicker in an popup. 
If I edit a new vote on a fresh card the cardStartVotingPopup opens. There is the option to set an end date ( see screenshot ). This opens the Datepicker popup. I need to get the selected date back to the cardStartVotingPopup. 

https://github.com/NicoP-S/wekan/blob/master/client/components/cards/cardDetails.js#L1044-L1047

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3064)


xet7:
@NicoP-S 

Some info about popups here, does it help?
https://github.com/wekan/wekan/blob/master/client/lib/popup.js#L6-L14

xet7:
Could there be a way to do it without subpopup?

xet7:
cardDetails does have due date popups working. Maybe if something similar date popup would work for voting?

PR #3068, opened 2020-04-30
[WIP] Some fixes

marc1006:
@xet7 Can you please write the code for the popups (see `TODO popup` comments), because I don't know the code well enough. Another point I'm not sure about is whether the check for the linker cards should take place on the server or not.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3068)


xet7:
@marc1006 

Yes, I will write that popup code and test. Thanks!

PR #3090, opened 2020-05-14
Change attachment library from Meteor-CollectionFS to Meteor-Files

urakagi:
I have tested on `Sandstorm` and standalone version for some existed boards and new boards, but I don't have a systematical test. Just played around.

- Client security is not set yet (`allowClientCode: true` now)
- The new export feature and import from `wekan`/`Trello` are not changed yet.

@xet7 What do you think about branching this change?

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3090)


xet7:
@urakagi 

When I tested this:
- showing attachments at card, and switching cover image does not work
- export does not work, in your code there seems to be removing those exporter lines that do work
- I did not test import

When you have added additional commits to fix these, I will test again.

I do only keep one master branch in Wekan, because I don't have time to merge between branches.

Thanks!

xet7:
You seem to have added this directory and test file to git repo, you can remove it from git, I will leave it attached to here for testing:
```
atachments/attachments-gAjLYeSrtAneFBdzt-提議者電子郵件(第一波+第二波).xlsx
```
[attachments-gAjLYeSrtAneFBdzt-提議者電子郵件(第一波+第二波).xlsx](https://github.com/wekan/wekan/files/4627940/attachments-gAjLYeSrtAneFBdzt-.%2B.xlsx)

urakagi:
@xet7 

For showing and switching attachments, it works on my machine so I think there are some environmental bugs. The storage path will be environment variable `ATTACHMENTS_STORE_PATH` or as described in the [library FAQ #1](https://github.com/VeliovGroup/Meteor-Files#faq).

Also, in my case, I use the same repo for standalone and Sandstorm version, so they will store files in different paths but shares the same database, and it will cause that attachments are not found. But for course, it only happens in the development stage.

Can you provide more information about this issue? Thank you.

urakagi:
@xet7 
Also, can you give us a Sandstorm spk on the pre-release market? Since we don't really care about the import/export and meteor security problem. (Since Sandstorm version only have one board)

urakagi:
The cover issue is solved.

xet7:
@urakagi

Sandstorm version has All Boards page and many boards on one grain since [Wekan v2.28 2019-02-27](https://github.com/wekan/wekan/blob/master/CHANGELOG.md#v228-2019-02-27-wekan-release)

urakagi:
@xet7 OK, I'll solve the import, export, and security problem, but can you give me an experimental sandstorm spk so we can run and check on my current machine? Thank you!

urakagi:
Exporting attachments is fixed, but I don't find where the export feature is so I can't test. Where is it?

marc1006:
> The cover issue is solved.

Are you sure? Because I still had the same issue. Here is a fix for it: https://github.com/wekan/wekan/pull/3111/commits/66d35a15280795b76a81c3e59cebbd2a29e4dff8

urakagi:
@marc1006 I didn't mean this but OK I'll add that.

server/migrations.js
```diff
@@ -80,7 +80,7 @@ Migrations.add('lowercase-board-permission', () => {
 Migrations.add('change-attachments-type-for-non-images', () => {
   const newTypeForNonImage = 'application/octet-stream';
   Attachments.find().forEach(file => {
-    if (!file.isImage()) {
+    if (!file.isImage) {
```
marc1006, line 83: Is this change intended?

urakagi: Yes, `isImage` is an attribute in `Meteor-Files` rather than a method in `Collection-FS`.

marc1006: Is this going to work if, for example, someone is trying to update from a really old version of Wekan to a newer one?

server/migrations.js
```diff
@@ -1041,3 +1041,46 @@ Migrations.add('add-sort-field-to-boards', () => {
     }
   });
 });
+
+import { MongoInternals } from 'meteor/mongo';
```
marc1006, line 1045: Should be probably moved to the top of the file.

server/migrations.js
```diff
@@ -1041,3 +1041,46 @@ Migrations.add('add-sort-field-to-boards', () => {
...
+
+Migrations.add('change-attachment-library', () => {
+	const fs = require('fs');
+	CFSAttachments.find().forEach(file => {
+    const bucket = new MongoInternals.NpmModule.GridFSBucket(MongoInternals.defaultRemoteCollectionDriver().mongo.db, {bucketName: 'cfs_gridfs.attachments'});
+    const gfsId = new MongoInternals.NpmModule.ObjectID(file.copies.attachments.key);
+ 	  const reader = bucket.openDownloadStream(gfsId);
+    let store = Attachments.storagePath();
+    if (store.charAt(store.length - 1) === '/') {
```
marc1006, line 1054: store.endsWith('/') reads easier IMHO.

server/migrations.js
urakagi: In that case, the `server/migration.js` should have updated all records when starting up so it should work.

I only tested with a not-old version of Wekan (current running version), though.

urakagi: Oh, no, sorry, I misunderstood this. I think this modification is an error when I fixing around `isImage()` to `isImage`. I'll recover this.

urakagi:
I put the next PR here: https://github.com/wekan/wekan/pull/3116

PR #3093, opened 2020-05-15
Added clearblue theme

CidKramer:
I wanted a theme which changed the color of the board's background aswell as the header. 

The theme uses transparent elements for the swim lanes and cards. The color option "clearblue" is added as an example but any desired color be created using the `setBoardClear(color1,color2)` function in **boardColors.styl**.

wekan_clearblue_20200515

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3093)

### DIFF
--- a/client/components/boards/boardColors.styl
+++ b/client/components/boards/boardColors.styl
@@ -293,3 +293,236 @@ setBoardColor(color)
 
   //.header-quick-access
   //  backgroud-color: #568ba2
+
+
+/*
+  Alternate "Clear" Styling
+*/
+setBoardClear(color1,color2)
+  //color1: The quick access color
+  //color2: The main bar color
+
+  &.sk-spinner div,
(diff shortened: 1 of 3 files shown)


xet7:
@CidKramer 

There is some bug, your changes disables the ability to change Swimlane Color and Card Color in all themes.

CidKramer:
Hmm, I was not able to replicate that. I did notice one problem with my theme though is the text doesn't change to white when you select an alternate card color.

xet7:
@CidKramer 

For me, there those change color popups did not have any colors listed at all.

CidKramer:
This is the behavior I am seeing:
card_color_1

card_color_2

I'm clicking the three bars, and selecting "Set color" on the swimlane and card respectively. In both cases I get a selection of colors to choose from:
card_color_3

xet7:
Ok, I will test again.

Issue #3107, opened 2020-05-23
Rules: If I have a card get moved to a list it is moved to the first swimlane rather than the one it was created in


hickorysb:
## Issue

Add these issues to elsewhere:
- Snap: https://github.com/wekan/wekan-snap/issues
- LDAP: https://github.com/wekan/wekan-ldap/issues
- UCS: https://github.com/wekan/univention/issues

Other Wekan issues can be added here.

**Server Setup Information**:

* Note: Please anonymize info, and do not add to this public issue any of your Wekan board URLs, passwords, API tokens etc, do you understand?: Yes
* Did you test in newest Wekan?: Yes
* For new Wekan install, did you configure root-url correctly so Wekan cards open correctly https://github.com/wekan/wekan/wiki/Settings ?
* Wekan version: 4.03.0
* Operating System: Ubuntu 19.10
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): snap
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): Nginx Reverse Proxy
* Node Version: v12.16.3
* MongoDB Version: 3.6.8
* Wekan only works on newest desktop Firefox/Chromium/Chrome/Edge/Chromium Edge and mobile Chrome. What webbrowser version are you using? Chrome Version 81.0.4044.138

**Problem description**:
![Problem](https://user-images.githubusercontent.com/44131234/82725876-9d803100-9c8c-11ea-8bf8-66b4d6434988.gif)
It moves the card to the first swimlane. I would like it to remain in whatever Swimlane it was originally generated it. Here is a screenshot of the rule I have set.
![image](https://user-images.githubusercontent.com/44131234/82725893-c56f9480-9c8c-11ea-808e-8e90e643f1f4.png)
Is there a way to do this currently without adding a rule for each swimlane, or is it possible for this to be introduced in a future update?

hickorysb:
I found that Wekan isn't automatically setting the swimlane to be "*" when I leave it blank, even though it says it is in the details. It's actually setting it to undefined. If I manually set the swimlane to "*" when I create the rule it works as intended.

marc1006:
@hickorysb Would be great if you could verify my fix by using my branch (see https://github.com/wekan/wekan/pull/3109) or wait until my pull request got merged.

PR #3109, opened 2020-05-23
Set '*' as default value for swimlane and list name in card move action

marc1006:
This is the same default as `Utils.getTriggerActionDesc` has. This commit fixes
https://github.com/wekan/wekan/issues/3107.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3109)


marc1006:
This should be probably fixed for many other actions as well. And probably there are some JS utils to implement this more nicely.

Issue #3099, opened 2020-05-18
strange problem : user swaped or overloaded


salleman33:
Hi, 

**Server Setup Information**:
- Wekan 4.01.0
- deploy on Openshift
- Meteor : 1.10.2
- Nodejs : 12.16.3
- Mongo : 3.6.3

**Problem description**:
I have problems with some users being swaped or overloaded. I explain : 
A few week ago, I saw in a board that an user (A) has been replaced by user (B). On the mongoDB, user A did not exist anymore and user B had the same ID of previous User A. It's like User B had overloaded user A and kept the same ID, so he has same access to boards, etc.
Today, I have an other problem : 
User C has their informations suddendly cleared : no username, no mail, no first name, etc.
On mongodb, I can find user C. But on the board member, there is not User C but an other ID members (123456). If I search this Id user on mongodb, there is nothing... 

 Do you encoutered this error ? How can I debug this ? Is there a debug mode ? 

On the wekan pod in openshift, there are sometime some error like that : 
```
Exception while invoking method 'login' RangeError: Maximum call stack size exceeded:
at Object.hasOwnProperty (<anonymous>)
    at hasOwn (packages/ejson/utils.js:9:70)
    at isArguments (packages/ejson/utils.js:17:50)
    at Object.EJSON.clone (packages/ejson/ejson.js:583:7)
    at packages/ejson/ejson.js:600:22
    at Array.forEach (<anonymous>)
    at Object.EJSON.clone (packages/ejson/ejson.js:599:13)
    at Array.forEach (<anonymous>)
    at Object.EJSON.clone (packages/ejson/ejson.js:599:13)
    at packages/ejson/ejson.js:600:22
    at Array.forEach (<anonymous>)
    at Object.EJSON.clone (packages/ejson/ejson.js:599:13)
    at packages/ejson/ejson.js:600:22
```

Thanks for your help

xet7:
@salleman33 

Can you try using newest MongoDB 4.x by using just docker tag `mongo` without any version number? I don't know is that MongoDB bug that is already fixed in newest MongoDB version.

Does your Wekan have enough RAM? If it does not have, it starts swapping, and slows down.

Also, if database gets out of disk space, it will corrupt.

You could Google search what that error message means.

salleman33:
Hi, 
Yes, RAM and disk space are ok. I will try mongo v4.
Is there a way to enable a debug mode ?

xet7:
@salleman33 

[...]

PR #3180, opened 2020-06-17
Update users.js

salleman33:
Don't know if it could be usefull : change method to find existing user for #3099

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3180)

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -933,7 +933,7 @@ if (Meteor.isServer) {
       user.authenticationMethod = 'oauth2';
 
       // see if any existing user has this email address or username, otherwise create new
-      const existingUser = Meteor.users.findOne({
+      const existingUser = Users.findOne({
         $or: [{ 'emails.address': email }, { username: user.username }],
       });
       if (!existingUser) return user;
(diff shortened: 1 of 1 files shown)


salleman33:
Yes, I think it was the wrong user who was deleted

models/users.js
```diff
@@ -933,7 +933,7 @@ if (Meteor.isServer) {
       user.authenticationMethod = 'oauth2';
 
       // see if any existing user has this email address or username, otherwise create new
-      const existingUser = Meteor.users.findOne({
+      const existingUser = Users.findOne({
```
marc1006, line 936: @salleman33 Can you please explain what's the benefit of this change? If I understand the code correctly, `Users` is simply a reference to `Meteor.users` (see https://github.com/wekan/wekan/blob/master/models/users.js#L7).

Do you prefer `Users` since the string length is shorter?

xet7: @marc1006 Currently I'm in progress of reverting these changes.

xet7: @marc1006 Because comments are not visible anymore in mobile view #3182

salleman33: Yes, this commit is useless. I did'nt know that Users equals Meteor.users. I thought it was different.
You can revert it .

Issue #3241, opened 2020-08-26
Drop custom scrollbar


tborychowski:
## Issue

**Server Setup Information**:

* Did you test in newest Wekan?: Yes
* Wekan version: [latest](https://hub.docker.com/layers/wekanteam/wekan/latest/images/sha256-16b859b86cfa0960c5e02328f7b586227e9ba3615034ce7d6c925b14d97106ce?context=explore)
* Operating System: Macos
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): docker
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): none
* Node Version: docker
* MongoDB Version: docker
* Wekan only works on newest desktop Firefox/Chromium/Chrome/Edge/Chromium Edge and mobile Chrome. What webbrowser version are you using?
Firefox

**Problem description**:
When a column has more cards than can fit in a screen, so that the scrollbar appears - scrolling on a mac in Firefox adds a weird fading delay and it scrolls more than needed. If I scroll even more - the cards will go all the way up and disappear eventually. (It seems to work fine in chrome).
I'm guessing this is because of a custom scrollbar implementation?

![Screen Recording 2020-08-26 at 08 22 28](https://user-images.githubusercontent.com/392513/91273663-819d9700-e775-11ea-9fd2-11de0661c4b2.gif)

xet7:
What version of Firefox you use?

xet7:
What version of macOS you have?

xet7:
I can not figure out how to duplicate this on macOS Catalina using Firefox 80. More information required.

tborychowski:
Hi @xet7 , thanks for your response.
I've tested this and was able to reproduce on the following platforms/browsers:
- macos catalina + firefox 80
- macos catalina + firefox 81 beta
- macos big sur + firefox 80
- macos big sur + firefox 81 beta
- windows 10 + firefox 80
- windows 10 + firefox 81 beta

Which would suggest that this is not a problem with a particular os/browser version but rather the setup.
Have you been testing this with a regular mouse? 
(I'm using apple's magic mouse or trackpad)
This looks like a problem with how firefox handles the inertia-scrolling and basically the "spin-down" from where the scrolling stopped and the inertia-scrolling starts is when it happens (as in the gif).

xet7:
There are settings for scrollInertia: https://github.com/wekan/wekan/issues/2949#issuecomment-600021905

[...]

PR #3246, opened 2020-08-28
fix(scrollbar): remove custom scrollbar

tborychowski:
fix for #3241

I've removed the custom scrollbar styling.
However, as this is a complex project and I'm not familiar with all the caveats, please test thoroughly and make sure it works as expected.
Especially on Windows (as it handles scrollbars differently than MacOS).

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3246)


.eslintrc.json
```diff
@@ -45,7 +45,7 @@
     "no-spaced-func": 2,
     "no-trailing-spaces": 2,
     "operator-linebreak": 2,
-    "quotes": [2, "single"],
+    "quotes": [2, "single", { "avoidEscape": true }],
```
tborychowski, line 48: This is because on commit - the eslint was complaining:
Screenshot 2020-08-28 at 11 48 14

PR #3370, opened 2020-11-25
sticky swimlane

progressify:
Please merge this pull request for my mental sanity :)
Now the swimlane header is sticky and is fixed if the page scroll vertically

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3370)

### DIFF
--- a/client/components/swimlanes/swimlanes.styl
+++ b/client/components/swimlanes/swimlanes.styl
@@ -44,6 +44,8 @@
   flex-direction: row
   overflow: 0;
   max-height: calc(100% - 26px)
+  position: sticky
+  left: 0;
 
   &.placeholder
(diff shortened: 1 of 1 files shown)


xet7:
Unfortunately, this does not work. After adding this, when there is many lists, it's not possible to scroll right to see more lists.

progressify:
I will try to fix

progressify:
ok, I found the trouble
I want to apply 
```
  position: sticky
  left: 0;
```
only at the divs that has only the class "swimlane".
is that the right syntax?
```
[class=swimlane]
  position: sticky
  left: 0;
```

progressify:
yes! Now works great

PR #3385, opened 2020-12-08
Minicard, reduce space after assignees label

mfilser:
I saw that there is a lot of space after the assignees label on the minicard if an assignee is set.

Before:
grafik

After:
grafik

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3385)

### DIFF
--- a/client/components/cards/minicard.jade
+++ b/client/components/cards/minicard.jade
@@ -85,7 +85,6 @@ template(name="minicard")
       .minicard-assignees.js-minicard-assignees
         each getAssignees
           +userAvatar(userId=this)
-        hr
 
     if getMembers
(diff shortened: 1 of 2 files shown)


xet7:
With your changes, problem is, when I add assignee and member to same card, assignee and member avatars overlap each other.

Wekan original layout is made so that assignees are above that line, because assignees are responsible for getting that task done.

mfilser:
I tried your note, for me, it's not overlap each other, or do you mean another thing?

Before:
grafik

After:
grafik

This is the card:
grafik

xet7:
Well OK, maybe this is better.

PR #3431, opened 2021-01-13
Search

jrsupplee:
Global search for searching cards from all boards.
Note:
* I have created some routines on the `cards` and `Boards` models so that I can use the same code for retrieving cards on the server and client side.  I am not sure if this is the best technique.  Let me know if there is a better way.
* Created a method `Boards.userSearch(...)` that I use to only return boards the user is permitted to see.  Take a look and make sure the logic is what is expected.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3431)


xet7:
@jrsupplee 

Thanks a lot !

Currently when I tried this, by registering first admin user and second normal user, searching cards as first user retuns also info from cards of second user, and vice versa. So there is some bug because it's not limited to only own cards.

I don't know is creating a publication required at all, I think you could remove publication.

Instead, at clientside do something like `Meteor.call('search',...)` similarly like there is other code you see with:
```
cd wekan
./find.sh Meteor.call
```
And then at `wekan/models/*.js` in that method you can search database with limiting to own cards etc only, and return result to client. When there are multiple results, you could maybe return Javascript array, and then at clientside loop though that array displaying all results.

There are some untranslated strings, like this beginning with:
```
      h1 Search Operators
        +viewer
          = 'Searches can include the operators to refine the search.  Operators are specified by writing the operator'
```          
For that, you could add translatable for example:
- `{{_ 'search-operators'}}`
- `{{_ 'search-can-include'}}`

etc, for all parts that can be translated. For other parts that can not be translated, like formatting of text and english-only search operators, do not add those as translatable strings at all.

I don't know would it be possible for search operators be translatable, or to have web UI where by clicking you select translatable search operators. For example, because Rules is web UI and translatable, it can be understood also by those that do not understand English at all.

Thanks again for your amazing contributions!

xet7:
You could try to add checkbox:
```
[_] fuzzy
```
And try to use this https://atmospherejs.com/perak/fuzzy-search for optionally having fuzzy search.

jrsupplee:
> Currently when I tried this, by registering first admin user and second normal user, searching cards as first user retuns also info from cards of second user, and vice versa. So there is some bug because it's not limited to only own cards.

It should only return cards from boards where the user is a member or public boards.  Was the search showing cards from boards of which the current user is not a member?

```
Boards.userSearch = (
  userId,
  selector = {},
  projection = {},
  includeArchived = false,
) => {
  if (!includeArchived) {
    selector.archived = false;
  }
  selector.$or = [
    { permission: 'public' },
    { members: { $elemMatch: { userId, isActive: true } } },
  ];

  return Boards.find(selector, projection);
};
```

jrsupplee:
> There are some untranslated strings, like this beginning with:
> 
> ```
>       h1 Search Operators
>         +viewer
>           = 'Searches can include the operators to refine the search.  Operators are specified by writing the operator'
> ```
> 
> For that, you could add translatable for example:
> 
>     * `{{_ 'search-operators'}}`
> 
>     * `{{_ 'search-can-include'}}`
> 
> 
> etc, for all parts that can be translated. For other parts that can not be translated, like formatting of text and english-only search operators, do not add those as translatable strings at all.
> 
> I don't know would it be possible for search operators be translatable, or to have web UI where by clicking you select translatable search operators. For example, because Rules is web UI and translatable, it can be understood also by those that do not understand English at all.

I wanted to make sure the description in English is good before having people translate it.  It still needs some editing, I think.

Making the search operators translatable would not be difficult.

```
          const operatorMap = {
            board: 'boards',
            b: 'boards',
            label: 'labels',
            lable: 'labels',
            user: 'users',
            u: 'users',
            swimlane: 'swimlanes',
            swim: 'swimlanes',
            s: 'swimlanes',
            list: 'lists',
            l: 'lists',
            is: 'is',
          };
```

As you can see here it is easy to map new search operators to a desired function.

xet7:
@jrsupplee 

Yes, currently it searches from all users private boards too.

jrsupplee:
@xet7 

This is part of the selector for all the searches.

```
  const selector = {
    archived: false,
    type: 'cardType-card',
    boardId: { $in: Boards.userBoardIds(userId) },
    swimlaneId: { $nin: Swimlanes.archivedSwimlaneIds() },
    listId: { $nin: Lists.archivedListIds() },
  };
```

As you can see the `boardId` must be in the set of boards returned by `userBoardsId(userId)`

I'll look at it and do some more testing, but it does not make sense to me.

jrsupplee:
@xet7 

uncomment the `console.log(...)` on line 1888 of `models/cards.js`

That will show the selector for the `Cards.find()`

jrsupplee:
@xet7 

Ok.  I was able to reproduce what you are seeing.

xet7:
@jrsupplee 

Could you try to make something this simplified web interface, with all text as translatable? That [X] And [_] Or is option buttons, so when one is selected, other is unselected. Here is minimal amount of text to translate. It should only require adding search keywords, and not anything special characters like `@` that could be hard to type on some language keyboards. Label colors description below should be replaced with actual visible colors of palette, and color names that are used in Wekan.

```
Examples:
- If not spaces: Searchtext1 Searchtext2
- If spaces: "Some text to search" "Another text to search"

_________________________________________________

Search text: ______________________________

Fuzzy: [_]

Limit:

Board title: __________________ [X] And [_] Or

Swimlane title: __________________ [X] And [_] Or

List title: __________________ [X] And [_] Or

Username of card member or assignee: __________________ [X] And [_] Or

Label color: 
[X] (first color showed here, as color, and as color name text)
[_] (second color showed here, as color, and as color name text)
[X] And [_] Or

Label text: _____________________ [X] And [_] Or
```

jrsupplee:
@xet7 

I found the error.  `Meteor.UserId` instead of `Meteor.userId()`.

jrsupplee:
As for making a form, this would be helpful, but it needs to be more complex than what you have shown.  Need to be able to enter multiple values for each operator (e.g. `list:Todo list:Done`).

The basic model I followed was what is used on another famous kanban application.

To be honest, I don't have the time to implement a form like that now.  There are other parts of the search that need to be finished first.  Like pagination.  It currently just returns a max of 50 counts.

jrsupplee:
I'll submit a new pull request with the patched code.

xet7:
@jrsupplee 

Ok, then I think making a form is not a good idea, because it can not cover all use cases. Your original idea is better.

For those operators, they would be better made translatable, with also some examples shown as translated. For example:
```
- list:Todo list:Done
```
Would be in Finnish:
```
- lista:Tehtävää lista:Valmis
```
The translatable strings in those examples could be: List, Todo, Done

jrsupplee:
> @jrsupplee
> 
> Ok, then I think making a form is not a good idea, because it can not cover all use cases. Your original idea is better.
> 
> For those operators, they would be better made translatable, with also some examples shown as translated. For example:
> 
> ```
> - list:Todo list:Done
> ```
> 
> Would be in Finnish:
> 
> ```
> - lista:Tehtävää lista:Valmis
> ```
> 
> The translatable strings in those examples could be: List, Todo, Done

How do you get translations inside of the Javascript code in Meteor.  What is the function name?

xet7:
@jrsupplee 

Search for examples like this:
```
cd wekan
./find.sh TAPi18n
```
For example:
```
TAPi18n.__('title'),
```

xet7:
@jrsupplee 

Detecting current language code is here, where text at language menu is changed for those languages that are not correctly translated in meteor tap:i18n package already:
https://github.com/wekan/wekan/blob/master/client/components/main/layouts.js#L72

jrsupplee:
I tried implementing the search using `Meteor.call

> @jrsupplee
> 
> Detecting current language code is here, where text at language menu is changed for those languages that are not correctly translated in meteor tap:i18n package already:
> https://github.com/wekan/wekan/blob/master/client/components/main/layouts.js#L72

Thanks.  `TAPi18n.__(...)` is what I need.

jrsupplee:
I tried implementing the search using `Meteor.call(...)`.  The problem is that it seems it is not possible to return a cursor.  You must first do a `.fetch()` on the cursor to convert it to an array.  `minicard` does not work with data as an array.  It is expecting to be passed a `Card` object.

It seems leaving it as it is might be best.  What do you think?

jrsupplee:
@xet7 
> @jrsupplee
> 
> Ok, then I think making a form is not a good idea, because it can not cover all use cases. Your original idea is better.
> 
> For those operators, they would be better made translatable, with also some examples shown as translated. For example:
> 
> ```
> - list:Todo list:Done
> ```
> 
> Would be in Finnish:
> 
> ```
> - lista:Tehtävää lista:Valmis
> ```
> 
> The translatable strings in those examples could be: List, Todo, Done

Actually, only *list* would be translatable.  *Todo* and *Done* are just text that is searched for in all the list titles.

xet7:
@jrsupplee

> It seems leaving it as it is might be best. What do you think?

Yes, leave it as it is currently. If there are any problems with it later, it can then be fixed later.

> Actually, only *list* would be translatable.  *Todo* and *Done* are just text that is searched for in all the list titles.

Ok, then only have *list* as translatable.

PR #3433, opened 2021-01-14
Search

jrsupplee:
# Global Search

* Search instructions page is still not translatable.  Thought we could wait to see if they are clear before taking this step.
* Currently only presents up to the first 50 cards.  Need to add pagination.
* Does not allow for negative predicates.  This should not be difficult to add.
* Left some commented out debug code that can be deleted when the code is stable.
* Added a new field to the User document to communicate some session data between server publications and the client.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3433)


xet7:
Please make search intructions translatable as full sentences or some amount of words. Translations are made for people that can not understand any English,

jrsupplee:
I will look at the translation interface to see the best way to get the instructions ready for translation.  It seems you can have HTML in translatable text if you use three curly braces.  Also will change some of the translation string to used named placeholders since the order may change in a translated language.

xet7:
@jrsupplee 

Please do not use 3 curly braces in translations, it is insecure unescaped HTML.

xet7:
@jrsupplee 

I am looking at adding translations myself.

xet7:
@jrsupplee

Actually, it's better for you to add translations, if possible. I'm not familiar enough with content of this PR.

xet7:
@jrsupplee 

I wish you could make every visible text (or most text) translatable.

In my example:
```
- list:Todo list:Done
```
Would be in Finnish:
```
- lista:Tehtävää lista:Valmis
```
This would be done this way:
- Translated texts: list, Todo, Done
- This means, that also those example texts like Todo is translated, to be a example in target language.

It would work this way:
1. In Search examples, show example like: `{{list}}:{{todo}} {{list}}:{{done}}
2. In that case, when someone has added lists in their own language, they could search with those list names
3. When parsing search text, you could for example:
```
// Split with space
var keywords = str.split(" ");
// Split with : 
var keywords = str.split(":");
// Convert text to English keywords
if keyword === TAPi18n.__('list') {
   // search for list:searchword
}  
```
I have not yet looked at your code how you implemented it, but, what do you think? Would it be possible?

jrsupplee:
> @jrsupplee
> 
> Please do not use 3 curly braces in translations, it is insecure unescaped HTML.

Yes, it creates the possibility of an injection attack.  However, there are instances in the code that use them.  I thought maybe you had some way of monitoring the translations.

xet7:
> > @jrsupplee
> > Please do not use 3 curly braces in translations, it is insecure unescaped HTML.
> 
> Yes, it creates the possibility of an injection attack. However, there are instances in the code that use them. I thought maybe you had some way of monitoring the translations.

Huh? Thanks for info! Just a moment I check.

jrsupplee:
> @jrsupplee
> 
> I wish you could make every visible text (or most text) translatable.
> 
> In my example:
> 
> ```
> - list:Todo list:Done
> ```
> 
> Would be in Finnish:
> 
> ```
> - lista:Tehtävää lista:Valmis
> ```
> 
> This would be done this way:
> 
> * Translated texts: list, todo, done
> * This means, that also those example texts like Todo is translated, to be a example in target language.
> 
> It would work this way:
> 
> 1. In Search examples, show example like: `{{_ 'list'}}:{{_ 'todo'}} {{_ 'list'}}:{{_ 'done'}}
> 2. In that case, when someone has added lists in their own language, they could search with those list names
> 3. When parsing search text, you could for example:
> 
> ```
> // Split with space
> var keywords = str.split(" ");
> // Split with : 
> var keywords = str.split(":");
> // Convert text to English keywords
> if (keyword === TAPi18n.__('list')) {
>    // search for list:searchword
> }  
> ```
> 
> I have not yet looked at your code how you implemented it, but, what do you think? Would it be possible?

I am traveling right now, but will try to do something tomorrow.

```
<code>{{_ ‘operator-list’ }}:{{_ ‘title’ }}</code> {{_ ‘operator-list-description’ }}
<code>{{_ ‘operator-list’ }}:{{_ ‘example-list-1’ }}</code> {{_ ‘description-example-1’ }}
```

I do not remember the exact format of the text at the moment, but something like this should work.  Actually, it will be more complex.

xet7:
@jrsupplee 

Thanks! Have a nice travel ! Anyway, your code seems to work mostly correctly, so I'll try to add it to new Wekan release and get some feedback.

xet7:
@jrsupplee 

It looks like you publish something from Users collection/table. Have you tested, does it publish only userId, or does it also publish other data like hashed passwords? I don't see any selector limiting returned data to for example userId only at Users find code.

xet7:
@jrsupplee 

I mean, previously there was this kind of bug: https://wekan.github.io/hall-of-fame/userbleed/

jrsupplee:
> @jrsupplee
> 
> It looks like you publish something from Users collection/table. Have you tested, does it publish only userId, or does it also publish other data like hashed passwords? I don't see any selector limiting returned data to for example userId only at Users find code.

I do not limit the fields on the Users collection returned.  You will want to add that.  I must say I have gotten the sense when displaying data on the client side that the fields are not limited.  For instance, displaying cards with `minicard` it displays field data that was not in the fields projection.

jrsupplee:
@xet7 
I saw your changes to the `Users.find(...)` in `globalSearch` and they seem reasonable.  However, when I run `Users.find().fetch()` in the console it displays many more fields than the few that were specified in the fields projection.

Are you sure that *Meteor* only passes the fields in the projection to the client?

xet7:
@jrsupplee 

When I tried `Users.find().fetch()`, it did only return fields for my own logged in user, and not any other users. Those all fields that a visible for me only are used in various Wekan features like showing avatar URL etc. I did not see my password hash there.

I think you are asking why I did not add limitations here:
https://github.com/wekan/wekan/blob/master/server/publications/cards.js#L247
while I did add limitations here:
https://github.com/wekan/wekan/blob/master/server/publications/cards.js#L255
It is because when I added those limitations to Global Search, then Global Search did not work at all, it did not return any cards.

I don't know are there any problems in these changes. Feedback welcome.

jrsupplee:
@xet7 
I just saw that you had not entered the fields projection for `globalSearch`.  I added that in with the addition of the new `sessionData` field.  This does leak some user information to the client, but there is currently nothing private stored in it.

PR #3469, opened 2021-01-22
Add horizontal scroll and link to open board to My Cards

helioguardabaxo:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3469)

### DIFF
--- a/client/components/main/myCards.jade
+++ b/client/components/main/myCards.jade
@@ -30,6 +30,7 @@ template(name="myCards")
           each board in myCardsList
             .my-cards-board-wrapper
               .my-cards-board-title
+                a.fa.fa-window-maximize.my-cards-board-link(href=board.absoluteUrl) 
                 +viewer
                   = board.title
               each swimlane in board.mySwimlanes
--- a/client/components/main/myCards.jade
+++ b/client/components/main/myCards.jade
@@ -36,15 +36,16 @@ template(name="myCards")
                 .my-cards-swimlane-title(class="{{#if swimlane.colorClass}}{{ swimlane.colorClass }}{{else}}swimlane-default-color{{/if}}")
                   +viewer
                     = swimlane.title
-                each list in swimlane.myLists
-                  .my-cards-list-wrapper
-                    .my-cards-list-title(class=list.colorClass)
-                      +viewer
-                        = list.title
-                    each card in list.myCards
-                      .my-cards-card-wrapper
-                        a.minicard-wrapper(href=card.absoluteUrl)
-                          +minicard(card)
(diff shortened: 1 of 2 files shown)


jrsupplee:
How does the scrolling work on mobiles?

Also what if you want the current behavior of wrapping instead of scrolling?  I personally do not like horizontal scrolling.  IMO would be better to have something you click on to enable the scrolling behavior.

helioguardabaxo:
@jrsupplee 
I have boards with multiples lists. In My Cards this lists and its cards shows outside the boundaries.
Your suggestion is good, but I don't know how to implements.Could you help?

jrsupplee:
I see the problem now.  There was a change made to the list style to use `table-cell` instead of `inline-block`.

```
.my-cards-list-wrapper
  margin: 1rem
  border-radius: 5px
  padding: 1.5rem
  padding-top: 0.75rem
  display: table-cell
  min-width: 250px
  max-width: 350px
```

I think it should be changed back to `inline-block`

jrsupplee:
> @jrsupplee
> I have boards with multiples lists. In My Cards this lists and its cards shows outside the boundaries.
> Your suggestion is good, but I don't know how to implements.Could you help?

Your solution is not  a bad one for `display: table-cell` for lists.

The boards URL is a good idea, but not sure about the window to the right.  Is that symbol used elsewhere in Wekan for a link?

helioguardabaxo:
I thought that a icon/link could create a destac, but brokes Wekan design harmony. 
It's better a text link?

Additionally, I tried to set board color, but doesn't works. I tried this: 
.my-cards-board-title(class=list.colorClass)
Any suggestion?

jrsupplee:
> I thought that a icon/link could create a destac, but brokes Wekan design harmony.
> It's better a text link?
> 
> Additionally, I tried to set board color, but doesn't works. I tried this:
> .my-cards-board-title(class=board.colorClass)
> Any suggestion?

I think that background colors are being set in `myCards.styl`.  Inspect the webpage and see what styles are being applied.

helioguardabaxo:
I inspected element in two differents boards in My Cards and found:
class="my-cards-board-title board-color-belize"
class="my-cards-board-title board-color-pomegranate"

If I disable background-color in myCards.styl, the element becomes transparent and keeps the class above.

helioguardabaxo:
I found here:
.my-cards-board-title(class=board.colorClass, id="header")

It's works!

helioguardabaxo:
I will close this PR and adjust the title link.

@jrsupplee 
Thanks for you help and suggestions

PR #3471, opened 2021-01-23
Add board title link and background color to My Cards

helioguardabaxo:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3471)


jrsupplee:
I think the grey border around the boards should be removed.  I was trying to visually tie the swimlanes and lists to the board, but it never looked great and it displays poorly on mobiles.

PR #3475, opened 2021-01-23
Use simple border at My Cards

helioguardabaxo:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3475)

### DIFF
--- a/client/components/main/myCards.styl
+++ b/client/components/main/myCards.styl
@@ -1,14 +1,15 @@
 .my-cards-board-wrapper
-  border-radius: 4px
+  border-radius: 0 0 4px 4px;
   //padding: 0.5rem
   min-width: 400px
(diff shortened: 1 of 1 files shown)


helioguardabaxo:
@jrsupplee 
I tried to modify board style.
What do you think?

xet7:
I have not looked it yet, but shadows are very hard for those that have low vision, so they are not used in Wekan. Wekan needs to have clear color differences, not shadows.

xet7:
So please remove shadow.

helioguardabaxo:
Could I use gradient or simple border is the better way?

xet7:
Simple border is the better way.

xet7:
Gradients are only in some themes. Sometime I will add global theme, so it's like board theme but for all non-single board pages.

PR #3489, opened 2021-01-27
changed method to create initials same as others for new user of oidc

sato-64bit:
In Japanese name, current method cause an error, "undefined.join()".
I copied initial creation method from User's method.  This worked on my env for English and Japanese.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3489)


sato-64bit:
I'm beginner of Wekan. If I need to open an issue at first, please let me know.

xet7:
@sato-64bit 

Thanks! It's not requred to open issue.

PR #3491, opened 2021-01-27
Mobile and Desktop have now the same Quick Access view + scrollable

mfilser:
**Improvements mobile view:**
- Quick Access scrollable
- Notification and Avatar doesn't overlap anymore
- Show all boards at Quick Access

**Improvements desktop view**
- Quick Access scrollable if necessary

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3491)

### DIFF
--- a/client/components/main/header.jade
+++ b/client/components/main/header.jade
@@ -17,6 +17,11 @@ template(name="header")
               li(class="{{#if $.Session.equals 'currentList' _id}}current{{/if}}")
                 a.js-select-list
                   = title
+          else
+            each currentUser.starredBoards
+              li(class="{{#if $.Session.equals 'currentBoard' _id}}current{{/if}}")
+                a(href="{{pathFor 'board' id=_id slug=slug}}")
+                  = title
         #header-new-board-icon
       else
         //-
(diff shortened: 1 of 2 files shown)


tod31:
@mfilser could you please take another look at this? Spacing is a bit off, in my eyes. Additionaly I cannot imagine, that the text-decoration of the "all boards" link is intended. I also do not know, whetther it is necessary that desktop and mobile view have the same header height.

old: 
grafik

current:
grafik

If I revert commits 282505cde2265b71f9da5c174f623d94e78e2686, bad4cc12b4037a67555cef04627812495d153f49 and d2d4c19196ee61a36369021585ac5ed9d783de2a I get the old look on desktop, but mobile is messed...

mfilser:
@tod31
If i remember right the desktop view was messed up if you have a lot of boards.

This is before the scrollbar occures, with some extra space:
grafik

And if i reduce the view or add boards, the scrollbar is there:
grafik

That's why there is currently extra space without a scrollbar. I am not an expert at html / css. If there is something easy to increase the high if the scrollbar occures it would be also fine. But for now without the extra space.
If i reduce the height it looks now like this:
grafik

In general i prefer to have not many extra code for anything. The more extra code there is the faster it breaks things when something changes at one view.

Issue #2074, opened 2018-12-27
Feature Request: Search in all boards


rlach:
**Problem description**:
- Wekan can currently search for cards only in the current board. In Trello you can also choose to search in all boards you have access to.

This way you could:
* find subtasks related to stories you currently have in your board
* find cards in other boards matching selected name

It's very useful for workflows where users move stories between boards.

Screenshot of the search example on Trello:
![zrzut ekranu z 2018-12-27 13-30-08](https://user-images.githubusercontent.com/5655000/50480329-8c6dba80-09db-11e9-88c7-66cfb58aa723.png)

superlou:
It would also be really helpful if this could search for all cards that someone is a member of. This could satisfy some use cases of #2055.

bhueck:
I also think this is a "must have" feature to get wekan in production (for a company).

gsommavilla:
Is there any update about this issue?
The "search-in-all-boards" feature is really something that I would love to have.

xet7:
@gsommavilla 

No one has started coding search feature yet. It would require someone to have time to read Wekan source code existing search feature, and add to it searching for other boards of user too. Code contributions as pull requests welcome. See Wekan GitHub wiki for Developer Docs and Directory Structure.

xet7:
If it is really important to have this soon, someone can add bounty, see:
https://www.bountysource.com/issues/68005834-feature-request-search-in-all-boards
https://wekan.team/bounties/index.html

r0bbie:
While I can't support the 500 € at this point unfortunately, would definitely be really keen to see this feature at some point, and if there was a BountySource (or similar, since I see you're moving away from that!) setup I'd happily chip in with others.

xet7:
@r0bbie 

Current payment methods are here:
https://wekan.team/commercial-support/

It works similarly, so if someone sends some part of that 500 € then it is added to [Roadmap](https://boards.wekan.team/b/D2SzJKZDS4Z48yeQH/wekan-open-source-kanban-board-with-mit-license).

xet7:
Moved to here from https://github.com/wekan/wekan/issues/552#issuecomment-692482399

From @gregruthenbeck 

> Search doesn't search comments. It would be great if it did.

Yes, I would think this `Search in all boards` will include searching comments. I did get funding for this feature at this week.

> Can this behavior be added/enabled relatively easily?

[...]

PR #3492, opened 2021-01-27
Global Search Update

jrsupplee:
* Add paging of results
* add new operators for searching(`comment`, `status`, `due`, `sort`, `created`, `modified`)
* text searches now search comments associated with cards

#2074 

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3492)


xet7:
@jrsupplee 

As a note for any future pull requests: I do only push pull request translations changes of original base language `wekan/i18n/en.i18n.json` to Transifex, all other translations are done at https://transifex.com/wekan/wekan . Sending any other translations as pull request is not useful, because it will get overwritten by translations I download from Transifex.

xet7:
@jrsupplee 

Thanks a lot for your amazing contributions!

jrsupplee:
> @jrsupplee
> 
> As a note for any future pull requests: I do only push pull request translations changes of original base language `wekan/i18n/en.i18n.json` to Transifex, all other translations are done at https://transifex.com/wekan/wekan . Sending any other translations as pull request is not useful, because it will get overwritten by translations I download from Transifex.

It’s ok if they get overwritten.  Most of it was for testing

xet7:
@jrsupplee 

It seems building Snap fails because of https://github.com/wekan/wekan/blob/master/client/components/main/globalSearch.js#L484 , see https://forum.snapcraft.io/t/userwarning-attempted-to-set-unknown-attribute-location-on-instance-of-container/22421 .

Could that code be changed to not use text `location` ?

xet7:
Issues at:
- https://forum.snapcraft.io/t/userwarning-attempted-to-set-unknown-attribute-location-on-instance-of-container/22421/4
- https://github.com/meteor/meteor-feature-requests/issues/389 

I don't know why this happens.

xet7:
Ok I fixed it, it was just missing backtick quotes in code. Strings values like regex text should be surrounded with quotes.

jrsupplee:
> Ok I fixed it, it was just missing backtick quotes in code. Strings values like regex text should be surrounded with quotes.

This has broken the search.  The backticks are not needed.

jrsupplee:
> @jrsupplee
> 
> It seems building Snap fails because of https://github.com/wekan/wekan/blob/master/client/components/main/globalSearch.js#L484 , see https://forum.snapcraft.io/t/userwarning-attempted-to-set-unknown-attribute-location-on-instance-of-container/22421 .
> 
> Could that code be changed to not use text `location` ?

The location code is not new.  The snap built with it before.  It, of course, can be rewritten to not use `window.location`

jrsupplee:
@xet7 

Could it be a Javascript version problem with the snap build?  I am using Unicode classes in the RegExp (which is new).

jrsupplee:
These two commits, broke the RegExps:
* bf7b1789ec16e3c52397318c799ec5a0fc2de3a5
* 0ff215f78f03c81d153dfc0ffa08fac94b542ec2

PR #3501, opened 2021-01-28
Rename Wekan import JSON menu title

jrsupplee:
The current title text does not clearly state that it is for Wekan JSON data

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3501)


xet7:
Yes, because it's whitelabeled. Text Wekan has been removed from everywhere, except Admin Panel, because of requests from my customers.

jrsupplee:
Ok.  Maybe we could insert `productName`?

PR #3505, opened 2021-01-29
Use `new RegExp(...)` to define a regex

jrsupplee:
Snap was not building with regexs defined using slashes.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3505)


jrsupplee:
If snap still does not build, I will try removing the Unicode flag and classes from regex

xet7:
@jrsupplee 

With your changes, Snap build did work. I will create a new Wekan release soon.

jrsupplee:
Great!

PR #3518, opened 2021-02-02
Sorted archives

bronger:
Added sorting to archives. Please review carefully as I don’t read JavaScript, Jade, or Meteor, so I had to rely on try-and-error and pattern reuse.

I particular, I was unable to add the `archivedAt` field to the list of archived boards.  Whatever I tried, I could only display the `title` of the board in the list of archived boards. Another problem is that I could not capitalise the first letter of “archived at”. Of course, one could use “Archived at” in `en.i18n.json` – but this is not as re-usable.

The new field `archivedAt` is necessary because `modifiedAt` of an archived object may change.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3518)

### DIFF
--- a/client/components/boards/boardArchive.js
+++ b/client/components/boards/boardArchive.js
@@ -11,7 +11,7 @@ BlazeComponent.extendComponent({
     return Boards.find(
       { archived: true },
       {
-        sort: { sort: 1 /* boards default sorting */ },
+        sort: { archivedAt: -1, modifiedAt: -1 },
       },
     );
   },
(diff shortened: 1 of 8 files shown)


jrsupplee:
@xet7 

Would it be better to set the archived at date (or remove it) via the schema when the archived attribute changes?

jrsupplee:
@bronger 

You need to change the `archiveBoards` publication in `server/publications/boards.js` to return your new `archivedAt` field.

PR #3519, opened 2021-02-02
Hopeful fix for i18n not working in `onRendered()`

jrsupplee:
* Remove the i18n initialization code from an `autorun()` block
* Add some console statements to help with debugging production.
* Add functions to `Boards` for label colors and color mapping

Issue #2074

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3519)

### DIFF
--- a/client/components/main/globalSearch.js
+++ b/client/components/main/globalSearch.js
@@ -78,12 +78,10 @@ BlazeComponent.extendComponent({
   onRendered() {
     Meteor.subscribe('setting');
 
-    this.colorMap = {};
-    for (const color of Boards.simpleSchema()._schema['labels.$.color']
-      .allowedValues) {
-      this.colorMap[TAPi18n.__(`color-${color}`)] = color;
-    }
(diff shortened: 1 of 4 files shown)


xet7:
@jrsupplee 

In newest Wekan I just released, when I select some other language, it does not change immediately. For Snap version, it only changes after reloading webpage. For Sandstorm, it does not change at all, it stays English. I will look sometime later should I revert this.

jrsupplee:
@xet7 

I noticed this too. I will try to look at it today

yangjy666:
> @jrsupplee
> 
> In newest Wekan I just released, when I select some other language, it does not change immediately. For Snap version, it only changes after reloading webpage. For Sandstorm, it does not change at all, it stays English. I will look sometime later should I revert this.

I found it

yangjy666:
I found that the v6.46 version still has this problem.

jrsupplee:
Tried French and Arabic on 6.46 and it seems to be working fine.

PR #3559, opened 2021-02-11
added padding

### LINKED ISSUE #3558: No padding between logo and Text in header
In the latest version there is no space between the logo and the text.

image

lindhork:
Fixes #3558 

image

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3559)


mfilser:
@lindhork 
Did you also check the mobile view how it looks there?
In the end it breaks the mobile view, especially if there are many favorite boards at the quick access.

lindhork:
Hey @mfilser,
I checked the mobile view, on an ipad pro it looks ok, on smaller devices the logo and text is not shown anyway.

grafik
grafik

I don't understand your argument why it breaks the mobile view? I have only edited the case when no board is favorited, the behavior when you have favorited boards has not changed.
You still get a long list in which you can scroll left and right.

grafik

mfilser:
@lindhork 
I was just thinking about it (because i did there some changes a few weeks ago) and didn't check it on my dev system.
Thank you for checking this and the screenshot's :)

lindhork:
@mfilser 
What do you think about moving the house a bit to the right for small devices, so that it doesn't stick directly to the edge?

grafik

mfilser:
@lindhork
I think this will be a good solution 👍
Thank you for mention it.

Issue #3424, opened 2021-01-09
Change the order of labels


ItamarSoffer:
## Issue

**Placing a new label higher is not possible.**

The first labels that was created is the first for good.
I would like to add a new labels and place them high, and move old labels lower.

basketball00011:
This would be awesome. I too find myself trying to re-order the labels.

helioguardabaxo:
@xet7 
Out of curiosity: would an alternative be to use a function similar to initSorting in [wekan/client/components/cards/checklists.js](https://github.com/wekan/wekan/blob/b64cd358ed0af4395357423ad172b8dac9dc3178/client/components/cards/checklists.js#L9)?

xet7:
@helioguardabaxo 

Yes, if you have time for that. Wekan uses JQuery draggable and sortable:
https://jqueryui.com/draggable/

For Wekan, search code like:
```
cd wekan
./find.sh sortable
```
Mobile web uses drag handles:
https://github.com/wekan/wekan/pulls?q=is%3Apr+is%3Aclosed+drag

helioguardabaxo:
@xet7 
Good! I never used JQuery, but I will learn it.
Thanks!

xet7:
@helioguardabaxo 

I don't mean you should learn JQuery. You just copy Checklists drag-drop code to be like Labels drag-drop code.

PR #3573, opened 2021-02-14
Adds pluralize to show cards count in list headers

helioguardabaxo:
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3573)


helioguardabaxo:
Hi, @xet7 !
I am trying to implements sort labels #3424.
In popover labels drag and drop is working, but partially. In labels.js, I copy onRendered() from subtasks sortable.
I don't know how to implements stop event. const 'label'  is returns undefined.
Other problem: drag is working just in popover and doesn't updates labels in minicard.
Any suggestions?
Thanks!

helioguardabaxo:
I will commit these features in separated commits

PR #3577, opened 2021-02-17
Setting to deny password change

lindhork:
Maybe one of you can help me, I wanted to add a setting to hide the change password field. The value in the DB is correctly adjusted when changing the setting ... but Metor.call always returns false for this setting.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3577)

### DIFF
--- a/client/components/settings/settingBody.jade
+++ b/client/components/settings/settingBody.jade
@@ -143,6 +143,14 @@ template(name='accountSettings')
         input.wekan-form-control#accounts-allowUserDelete(type="radio" name="allowUserDelete" value="false" checked="{{#unless allowUserDelete}}checked{{/unless}}")
         span {{_ 'no'}}
     li
+    li.accounts-form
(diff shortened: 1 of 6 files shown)


lindhork:
I noticed that there is the same problem with the already existing queries (allowEmailChange and allowUserDelete).

lindhork:
The calls work now, my experience with jade is limited, try to find out the days why it makes no difference what the function returns.

jrsupplee:
@lindhork 

This code should not work unless the timing is just right:

```
  allowPasswordChange(){
    Meteor.call('accountSettings.allowPasswordChange', (_, result) => {
      // console.log(result);
      if (result) {
        console.log(true);
        return true;
      } else {
        console.log(false);
        return false;
      }
    });
  },
```

`Meteor.call()` is asynchronous.  The function finishes with no return before the the result function is called.

Do you get the `true` logged to the conosle?

PR #3593, opened 2021-02-20
Add custom field editing to the REST API

dudeofawesome:
I'm not sure what else I need to add in order for the REST API docs to be updated correctly, etc.

This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3593)

### DIFF
--- a/models/customFields.js
+++ b/models/customFields.js
@@ -477,6 +477,43 @@ if (Meteor.isServer) {
   );
 
   /**
+   * @operation edit_custom_field_dropdown_item
+   * @summary Update a Custom Field's dropdown item
+   *
+   * @param {string} name names of the custom field
+   * @return_type {_id: string}
+   */
+  JsonRoutes.add(
+    'PUT',
+    '/api/boards/:boardId/custom-fields/:customFieldId/dropdown-items/:dropdownItemId',
+    (req, res) => {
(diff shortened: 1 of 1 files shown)


jrsupplee:
I am not sure how the API is supposed to be used, there are some practices in the APIs that should be reviewed.

1. The use of `collection.direct...` to avoid the firing of schema hooks.  This appears to be standard practice in the APIs.  Why?
2. In this pull request the ability is granted to change the type of a custom field.  This could easily leave the custom field data in a corrupted state.  It’s as if the APIs are like being in a 3rd party database browser and you can change the data however you want without regard to data integrity.

xet7:
@jrsupplee 

I think, that direct is used, because if all those hooks would be fired, and some code uses many REST API calls, many Wekan users would get a lot of email notifications as in levels of spam.

xet7:
Hmm, I think this is only additional API. It existing does not affect anything else. Those that use this API can try it, check that it does changes correctly, fix database if required, and also send additional fixes to this API.

xet7:
But, because this pull request is still marked as work in progress, I'm not able to merge it yet. Only when @dudeofawesome changes it to normal pull request I can merge it.

xet7:
Seems I was able to merge this.

PR #3631, opened 2021-03-04
Added real autolinking and simpler adding of new url schemes prepared setting in admin backend

xet7:
In Progress, does not work yet.


chrisi51:
didnt saw u tried it on your own ... i made a new request ...

i dunno why there are missing files ... seems like i cant git commit directly from my development server without having trouble like this. so i cleaned everything up and made a request now

PR #3635, opened 2021-03-05
Fixed Bug: inconsistent use of relative/absolute URLs

Majed6:
Some pages use relative links such as boards link at the home page.
 Others use absolute url such as cards in boards' lists.
 This commits goal is to allow for consistent use of relative
  urls.

Origin relative URLs also helps decoupling Wekan from the infrastructure
 it's deployed on. i.e if it's being served, it should work.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3635)

### DIFF
--- a/client/components/activities/activities.jade
+++ b/client/components/activities/activities.jade
@@ -82,7 +82,7 @@ template(name="activity")
             +viewer
               = activity.checklist.title
         else
-          a.activity-checklist(href="{{ activity.card.absoluteUrl }}")
+          a.activity-checklist(href="{{ activity.card.originRelativeUrl }}")
             +viewer
               = activity.checklist.title
 
@@ -103,7 +103,7 @@ template(name="activity")
 
       if($eq activity.activityType 'addChecklistItem')
         | {{{_ 'activity-checklist-item-added' (sanitize activity.checklist.title) cardLink}}}.
-        .activity-checklist(href="{{ activity.card.absoluteUrl }}")
+        .activity-checklist(href="{{ activity.card.originRelativeUrl }}")
           +viewer
             = activity.checklistItem.title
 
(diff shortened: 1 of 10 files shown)


models/activities.js
```diff
@@ -136,7 +136,7 @@ if (Meteor.isServer) {
         params.board = '';
       }
       title = 'act-withBoardTitle';
-      params.url = board.absoluteUrl();
+      params.url = board.originRelativeUrl();
```
Majed6, line 139: Should we use `absoluteUrl()` here instead ?

models/activities.js
```diff
@@ -179,7 +179,7 @@ if (Meteor.isServer) {
       watchers = _.union(watchers, card.watchers || []);
       params.card = card.title;
       title = 'act-withCardTitle';
-      params.url = card.absoluteUrl();
+      params.url = card.originRelativeUrl();
```
Majed6, line 182: Should we use `absoluteUrl()` here instead ?

models/activities.js
xet7: @Majed6 

Please try to get it working everywhere with relative URLs.

models/activities.js
xet7: @Majed6 

Please try to get it working everywhere with relative URLs.

Majed6:
The reason behind this PR is an issue I faced with my deployment of Wekan. I have it deployed on an android phone with root url being the default localhost. When I try to access this Wekan instance from a computer on the same wifi network some of the UI elements stop working because they're pointing to localhost. For example I can open a board but not a card within the board. Setting root url will fix it but that's too burdensome for such a portable device.

xet7:
@Majed6 

There is workaround to use 2 node.js wekan servers with different ROOT_URLs and same database, but of course relative URLs for everything would be better.

models/activities.js
Majed6: I've reverted this. A relative notification (email) doesn't make a loot of sense. Relative to whom ? The activity maker or the activity notification receiver ?

Issue #3647, opened 2021-03-12
Can't remoove filter "Sort Cards"


Kairiefr:
## Issue

**Server Setup Information**:

* Did you test in newest Wekan?: Yes
* Wekan version: 5.05
* Operating System: CentOS 7
* Deployment Method(snap/docker/sandstorm/mongodb bundle/source): docker
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): apache
* Wekan only works on newest desktop Firefox/Chromium/Chrome/Edge/Chromium Edge and mobile Chrome. What webbrowser version are you using? Chrome 89.0.4389.82, Firefox 78.8.0esr, Opera

**Problem description**:

After sorting the cards in a board using the "Sort Cards" option, I can not more rearrange cards as I want, I just can change them from one column to another.
It becomes impossible to move the cards up or down. It's like I can't remove the "Sort Cards" filter
In addition, the option "Sort cards" applies to all board. So we have the same problem of inability to move the cards with all the boards

The only solution I have found is to delete my cookies from the browser

I don't know if this is the desired use of this new feature (added in version 4.98) or a "bug".

Is it possible to remove this filter "Sort Cards" to have the "real" board (not sorted) and to be able to move the cards again without having to delete the cookies of the browser?
Not like delete the option,but juste choosing activate it or not.

Hoping to have been as clear as possible.

xet7:
@ednamaeG 

Do you have time to fix this?

xet7:
In general, is something a bug or feature, depends does it work well enough, and is someone so interested in some feature that someone keeps fixing it so it mostly works. Those features that do not work usually get removed.

ClaudiaK21:
I noted this issue here too: https://github.com/wekan/wekan/discussions/3607#discussioncomment-409409

When using a sorting in the board area, you can not move a card in this view - only in the standard/default view (which is correct).
But to get back to the standard default, **you need to refresh** the board page, switching to another board and coming back does not reset the set sorting, so it might be helpfull to add a "Standard" or Default option to the sorting options and maybe a highlighting so you can see which sorting is active.

xet7:
Fixed at https://github.com/wekan/wekan/pull/3662

ClaudiaK21:
@ednamaeG
thanks for the adjustment. The  "Added a button to clear/reset sort" is nice and highlighted
[...]

PR #3662, opened 2021-03-19
Fixed sort cards feature

ednamaeG:
- Fixed Sort Cards Feature #3647 
- Added a button to clear/reset sort
- Sorting is reset to default when switching to other boards

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3662)


ClaudiaK21:
I noted an issue here https://github.com/wekan/wekan/issues/3647#issuecomment-803942354
"Sorting is reset to default when switching to other boards" does not work correctly.

PR #3123, opened 2020-05-25
Add the 'Currency' Custom Field type

habenamare:
This PR implements #3082.

wekan-currency-custom-field

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3123)

PR #3701, opened 2021-04-08
Add custom field "stringtemplate"

tod31:
This PR implements a new custom field of type "stringtemplate".

It allows the user to add a string template containing a placeholder "%{value}" which will then be replaced by the card's field value.

I want to use this in combination with abas ERP, where you can directly open windows from the browser by using the custom URL scheme abasurl. For showing the data of the part with the part number 123456 the url would be `abasurl:<(Part)>123456<(View)>?DDE=001`. As the command is rather complex, it is quite tedious to write this link and I cannot imagine, that my users are able to write such a link by hand. So I will change the current custom field of type text to a new custom field of type stringtemplate, where the format is `[%{value}](abasurl:<(Part)>%{value}<(View)>?DDE=001)`, so that the link is auto generated.

The raw value is split by line separators and each non-empty line is formatted separately.

The following GIF shows an example, how you could use this field, to add a custom field for a phone number which is linked with the tel scheme.
stringtemplate

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3701)

### DIFF
--- a/client/components/cards/cardCustomFields.jade
+++ b/client/components/cards/cardCustomFields.jade
@@ -119,3 +119,23 @@ template(name="cardCustomField-dropdown")
         if value
             +viewer
                 = selectedItem
+
+template(name="cardCustomField-stringtemplate")
+    if canModifyCard
+        +inlinedForm(classNames="js-card-customfield-stringtemplate")
+            +editor(autofocus=true)
+                = data.value
+            .edit-controls.clearfix
+                button.primary(type="submit") {{_ 'save'}}
+                a.fa.fa-times-thin.js-close-inlined-form
+        else
+            a.js-open-inlined-form
+                if value
+                    +viewer
+                        = formattedValue
+                else
+                    | {{_ 'edit'}}
+    else
+        if value
+            +viewer
+                = formattedValue
--- a/client/components/cards/cardCustomFields.jade
+++ b/client/components/cards/cardCustomFields.jade
@@ -125,7 +125,7 @@ template(name="cardCustomField-stringtemplate")
         +inlinedForm(classNames="js-card-customfield-stringtemplate")
             each item in stringtemplateItems.get
                 input.js-card-customfield-stringtemplate-item(type="text" value=item placeholder="")
-            input.js-card-customfield-stringtemplate-item.last(type="text" value="" placeholder="{{_ 'custom-field-dropdown-options-placeholder'}}")
+            input.js-card-customfield-stringtemplate-item.last(type="text" value="" placeholder="{{_ 'custom-field-stringtemplate-item-placeholder'}}")
             .edit-controls.clearfix
                 button.primary(type="submit") {{_ 'save'}}
                 a.fa.fa-times-thin.js-close-inlined-form
--- a/client/components/cards/cardCustomFields.jade
+++ b/client/components/cards/cardCustomFields.jade
@@ -123,8 +123,9 @@ template(name="cardCustomField-dropdown")
 template(name="cardCustomField-stringtemplate")
     if canModifyCard
         +inlinedForm(classNames="js-card-customfield-stringtemplate")
-            +editor(autofocus=true)
-                = data.value
+            each item in stringtemplateItems.get
+                input.js-card-customfield-stringtemplate-item(type="text" value=item placeholder="")
+            input.js-card-customfield-stringtemplate-item.last(type="text" value="" placeholder="{{_ 'custom-field-dropdown-options-placeholder'}}")
             .edit-controls.clearfix
                 button.primary(type="submit") {{_ 'save'}}
                 a.fa.fa-times-thin.js-close-inlined-form
--- a/client/components/cards/cardCustomFields.js
+++ b/client/components/cards/cardCustomFields.js
@@ -272,7 +272,9 @@ CardCustomField.register('cardCustomField');
           if (event.keyCode === 13) {
             event.preventDefault();
 
-            if (event.target.value.trim()) {
+            if (event.metaKey || event.ctrlKey) {
+              this.find('button[type=submit]').click();
+            } else if (event.target.value.trim()) {
               const inputLast = this.find('input.last');
 
               let items = this.getItems();
--- a/client/components/cards/cardCustomFields.js
+++ b/client/components/cards/cardCustomFields.js
@@ -234,3 +234,33 @@ CardCustomField.register('cardCustomField');
     ];
   }
 }.register('cardCustomField-dropdown'));
+
+// cardCustomField-stringtemplate
+(class extends CardCustomField {
+  onCreated() {
+    super.onCreated();
+
+    this.stringtemplateFormat = this.data().definition.settings.stringtemplateFormat;
+  }
+
+  formattedValue() {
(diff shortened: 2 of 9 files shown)


mfilser:
Very well @tod31 
I had this already in my mind too.
Exactly the same idea you implemented :+1:
For me i wanted to create Links to our ticket system.

I think this is a very useful feature for nearly everyone who uses Wekan :clap: 

By the way, we have Abas ERP also :slightly_smiling_face:

mfilser:
I just see at your screencast the second line isn't "translated" to a `<br>`.
Is this the behavior you want ?

tod31:
Thank you, @mfilser 

About the second line: yes, that is indeed the intended behaviour. The join is performed with a (hardcoded) space as separator. Do you think it would be better to add a second configuration field for the separator?

mfilser:
@tod31 
Yes, for me i would expect this that every line is after the replacement on an extra line. In the screencast it's also shown on the minicard, and also the custom field "Text" also translates each line with a line break.


<br /><br />
Another, but not important for me for now, is, to change the value.
e.g.:
`${value:^0(.+)/$1/}`
value: 012345
replace: 12345

and/or another replace option:
`${value:^0(.+)/0049$1/}`
value: 012345
replace: 004912345

mfilser:
@tod31 
I see the replace code is at several places, maybe it would be easier to have one function for the replacement. What do you think?

tod31:
@mfilser I understand your points. Just let me clarify my thoughts and point of view while implementing this feature.

**line break:** Originally I started with a single line input field: one raw value -> one formatted string. But as I also want to be able to link multiple tickets, parts, etc. I needed a way to split the input, to apply the formatting multiple times. As I had especially URL parameters (and abas commands) in mind, I decided to use a line break as separator for the split as every other typical split character (for example `,/;|-`) might be part of the raw value.

So while inputting the raw values and doing a line break, I do not have a line break in the output in mind, but a way to tell the field where to perform the split of the raw values to format them. Imho you cannot compare the behaviour to the custom field "text" since this field simply passes its content. About the screencast: It is multiline on the minicard, because the column width is narrower there. There is no line break in html.

**duplicate replacement code:** you are totally right, I frowned about this too. I was inspired by PR #3123 of the custom field currency, where you can find the same issue. But I must admit, that I do not know, where to put the extracted method, so that it can be called from both functions/classes. I am actually just a mechanical engineer, who happens to be halfway capable of programming and did his first steps with meteor.

**modifying the raw value while formatting:** looks really interesting, but exceeds my skills at first glance :sweat_smile:

mfilser:
@tod31 
Thanks for your good explanation.
If you like i could have a look at it too.

You are doing well at coding, maybe you should change your job :slightly_smiling_face:

tod31:
@mfilser feel free to add changes where you see fit

I did some further digging about the rendering of the custom field on minicard and card detail. On the minicard there is a surrounding container with class .minicard-custom-field-item which sets a max-width of 100px whereas the surrounding container of the card details is of class .card-details-item-customfield with a max-width of 36%.

I then also toyed around with the idea of an additional setting field for a join seperator. I first noticed that all input gets trimmed and entering a simple space and saving does not put anything into the database. So you would have to enter `&#32;` or `&nbsp;` (which might lead to other undesirable effects like overflowing lines, because it is a non-breaking space) to use a space as separator. During this I also noticed, that I misunderstood the join function. I thought joining with an empty string would also lead to commas as separators, but this is only the case when the argument is omitted or null/undefined. So I am currently implementing a second configuration variable stringtemplateSeparator as I am no fan of a hardcoded line break. (... any idea how I can speed up the insanely slow rebuild times (around 10 to 15 minutes) of meteor?)

I also had further thought about entering the raw values. Currently I am abusing a multiline field for entering a list, where every line represents a single raw value. This already led to confusion in this pull request. It might be better to adopt the behaviour of the configuration of the dropdown custom field, where you are presented a single line input field for every option which then gets stored as an array in the database. So it would be clear to the user, that he is currently entering multiple values, which then get treated each on their own.

tod31:
Had to force-push, as I accidentally published my private email in the commits :sob: Hope that does not lead to further problems...

Nevertheless, 3778a9ebcf17f9b9b81740b98280881e46aeee31 implements the separator setting. (label content needs fixing...)

grafik

mfilser:
@tod31 
This is a very good solution with the Separator field :+1: 

About building Meteor:
I develop Wekan in a docker container:
https://github.com/wekan/wekan-dev.git

But also with this docker container it takes about 5 minutes for me, maybe less, maybe more. I will have a look at it in future.

xet7:
@tod31 

> (... any idea how I can speed up the insanely slow rebuild times (around 10 to 15 minutes) of meteor?)

It's all about disk speed. HDD write speed is about 60 MB/sec, new SSD speed is about 900 MB/sec. SSD is over 10X faster. This also means, that when running Wekan, with HDD Wekan can load one board in 5 minutes, and with SSD in 2 seconds.

Disk write speed can be seen this way:
```
dd if=/dev/zero of=test.img bs=1M count=1000 status=progress
```

Also, using real bare metal install of Linux (Ubuntu 20.10, Pop OS Linux, etc .deb based distro), because Linux filesystem is much faster than Windows filesystem. Building at Win10 WSL2 Ubuntu with SSD disk is fast too, but if Defender slows down build, Defender should be stopped.

Build steps are:
```
cd wekan
# select 1 to install deps, only once:
./build-wekan.sh
# select 2 to build Wekan, only once:
./build-wekan.sh
# select 3 to to run wekan at http://localhost:4000 , it will detect file changes and rebuild automatically:
./build-wekan.sh
# only if some changes are not detected, a) ctrl-c stop that 3 and start 3 again or b) build with select 2 again, then 3
```
New in Meteor is [hot module replacement](https://docs.meteor.com/changelog.html#v2020210120), it shoud rebuild even faster, but I have not tried yet how it works.

mfilser:
@xet7 
If someone (like me) wants to develop in a docker container, is this still the best way to do it?
https://github.com/wekan/wekan-dev.git

xet7:
@mfilser 

I have not used that wekan-dev. I don't use docker container, it's slow. I build directly on ubuntu x64 or debian x64.

xet7:
@mfilser

I build Wekan this way https://github.com/wekan/wekan/wiki/Emoji

tod31:
I now use a string array for storing the values in the database and multiple input fields for entering the raw item values in the card details. I want to be able, to inject items between existing ones (user presses enter and gets a new empty field below the current one), but I am currently struggling with focusing the new (empty) field.

2021-04-11 14_29_28-Window

tod31:
focus
setting the focus after rerendering working now

mfilser:
@tod31 
I tried your latest commit c845d25, it works very well for me, thank you a lot for this.
I noticed Ctrl+Enter wasn't working (save all input fields to database and close the formular). 
I added the extra code to my local git. Should i send you a screenshot or push the commit above your commit?

tod31:
@mfilser I don't mind, as you prefer.

mfilser:
@tod31 
I forgot to ask if it's ok.
If it's ok i would push it above your commit

mfilser:
I can't push here, maybe you can cherry pick my commit 4d6e208 and push here if you like.

mfilser:
@tod31 
Great, thank you :slightly_smiling_face:

PR #3731, opened 2021-04-19
Support filtering by due date

mcrute:
This pull request adds the ability to filter a board by due date. Supported by default are:

* No due date
* Overdue (any time before today)
* Due Today
* Due Tomorrow
* Due This Week

The code could theoretically support other use-cases but there isn't an established UI design pattern for using the date picker in the filter bar and I don't really need open ended date support for my use-case.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3731)


marcungeschikts:
Hi @mcrute  ,
Thanks a lot for adding this new functionality.
In my use case, I miss the default value "Due Date", to filter all the card with a Due Date, no matter the date.

mcrute:
> In my use case, I miss the default value "Due Date", to filter all the cards with a Due Date, no matter the date.

I agree, that would be nice to have. It should be quite easy to add to this code. If you have the time, please add the feature. Otherwise I’ll plan to submit another PR in a few days.

Issue #802, opened 2017-02-08
Add Feature: Teams/Organizations similar to Trello


xet7:
_From @aegrey on December 18, 2016 1:53_

### Feature Description   
Similar to Trello, an administrator would be able to create a Organization/Team/User. Any boards created under this team would be accessible to all members of the team. Very high level of changes would be:

- [ ] Organizations: Add/Modify/Remove in Admin Panel.
- [ ] Teams: Add/Modify/Remove in Admin Panel. Board overview page would be available for the team, similar to the users, but with additional controls to manage the team members.   
- [ ] Teams: Privacy: Private/Public. With Private option, making board public is not allowed.
- [x]  Users: Modify in Admin Panel, change password.
- [x]  Users: Add in Admin Panel.
- [ ]  Users: When adding, also save fullname.
- [ ]  Users: Remove in Admin Panel. There is Error after deleting a user #1289
- [ ] LDAP: Subgroups https://github.com/wekan/wekan-ldap/issues/58
- [ ] LDAP: map unlimited Wekan profile items to LDAP fields. If profile is changed in Wekan, should Wekan have ability to save data back to LDAP? Or should Wekan be able to override LDAP settings? https://github.com/wekan/wekan-ldap/issues/58
- [ ] LDAP: Map Wekan Roles to LDAP Roles https://github.com/wekan/wekan-ldap/issues/58
- [ ] LDAP: Check that only Admin LDAP subgroup gets Wekan Admin permissions https://github.com/wekan/wekan-ldap/issues/58#issuecomment-495890729
- [ ] LDAP: Check that LDAP filter works correctly. Is it possible to simplify filter settings? https://github.com/wekan/univention/issues/5

### Discussion Requests      
https://discourse.wekan.io/t/teams-and-permissions/77 (somewhat similar)     
https://discourse.wekan.io/t/any-progress-on-users-being-able-to-see-and-interact-all-boards-immediately-upon-registration-suggestion-groups/206      
     
    
I currently don't know of any code bases that would assist in implementation of this feature, but this is the main reason why I'm still using Trello for my team, so I'd like to contribute. I did not see any pull request for it in the main repo. I know this would be a fairly heavy lift with database restructuring necessary.   
    
If this isn't something someone currently has on their radar, I have forked this repo and will plan on documenting the technical requirements to implement this feature, and then communicate this and possibly start implementation of the feature as time allows. 

_Copied from original issue: wefork/wekan#52_

xet7:
[...]

PR #3842, opened 2021-06-07
Addressed one of the bugs cited in issue #802 'bug when creating and …

Emile840:
Addressed one of the bugs cited in issue #802 'bug when creating and modifying organization and team'.

@xet7  I'm going to work on assignment of authenticated users to organization/team. I think i'll use the first alternative you gave me last Friday.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3842)

### DIFF
--- a/client/components/settings/peopleBody.jade
+++ b/client/components/settings/peopleBody.jade
@@ -132,28 +132,28 @@ template(name="newUserRow")
 
 template(name="orgRow")
   tr
-    if orgData.loginDisabled
+    if orgData.orgIsActive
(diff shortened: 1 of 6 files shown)


xet7:
@Emile840 

Thanks! I think with `authenticated users` you mean those users that do not have `loginDisabled true`:
- https://github.com/wekan/wekan/blob/master/models/users.js#L302
- https://github.com/wekan/wekan/blob/master/client/components/settings/peopleBody.jade#L203

PR #3920, opened 2021-07-26
Fixing importing of json file without labels

BartoszBroda:
Simple fix, adding one if statement.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3920)


BartoszBroda:
Ah, you were faster to fix than me creating the pull request :)

Issue #1473, opened 2018-02-09
Feature request: Consecutive card numbering?


therampagerado:
When do you plan to introduce this feature?

xet7:
@therampagerado 

Do you mean counting cards in list? Wekan already has that feature, see screenshot below.

Or do you mean other feature? Please explain, with screenshot mockup if possible.

![test counting cards](https://user-images.githubusercontent.com/15545/36022620-606a2aae-0d92-11e8-807e-d9cacd3d901d.png)

therampagerado:
Hello xet7, thank you for your quick reply!

No I mean each new card to receive new unique number so this system can be used as internal tickets system or orders system.

Our case is:
We are shop and we need to manufacture our goods and we need software that can have kanban with different stages of one order and each order must have unique number so we can track them by that.

therampagerado:
Also I installed snap with Ubuntu on my virtualbox for testing.

Is this secure enough for intranet use? I see it does not send out emails which is nice.

But I see couple of issues:
1. There is no settings on normal users - what they can and can not do. So now they can delete the whole kanban, do whatever they want.
2. The attachments (in our case pdfs) are not displaying in the preview. They download normally.

The first one is a stopper for us. I would like to have granulated control so not everybody can stop the company's daily work routine at any time (by mistake or on purpose).

Also I can you give me info on how to update the snap release to the latest version?

Sorry for this offtopic!

xet7:
@therampagerado 

Hi,
this is not offtopic, Wekan is general purpose kanban tool that is in use from single users to large companies where there is thousands of users of Wekan, and also in high security places like healthcare etc. Wekan does not by itself send any data to Internet. Wekan is in use in intranets and also works without direct internet access, if you don't need to download updates, or move updates with some media like usb stick or cdrom to intranet.

Wekan does not send email if email settings are not configured correctly.

In board settings you can set from other user icons the permissions to comment only, they are not able to delete then.

Previewing PDF files is not implemented, please ask about it at https://github.com/wekan/wekan/issues/1318 could somebody implent it, or pay some developer to implement it, or add bounty to that issue - instructions at https://wekan.team

Example of implemented workflow with rest api is at https://github.com/wekan/wekan-gogs , other way to implement workflows is with Huginn or Flogo https://github.com/wekan/wekan/issues/1160

You ca update all snaps to newest version with command: sudo snap refresh
After updating you need to setup root-url again https://github.com/wekan/wekan-snap/issues/29

[...]

PR #3935, opened 2021-08-02
Feature: Consecutive boardwise card numbering

syndimann:
Hey @xet7,

first of all thanks for maintaining wekan!
We have been using wekan some time now in our company as a Trello alternative and we came across this [feature-request](https://github.com/wekan/wekan/issues/1473) as we also need to keep track of the cards.

This Pull Request adds this feature to wekan in the following way:

- it assigns a boardwise consecutive `cardNumber` to each card via a migration-script to ensure backward-compatibility
- it further assigns the nextCardNumber whenever a new card is created either through the rest-endpoint or websocket
- Through the card-settings it is possible to enable or disable the appearance of the cardNumber per board. The migration script sets this config to `false` initially
- If enabled the card-number is shown on the mini-card as well as the detail-card.

Please find below screenshots of the feature.

Feel free to reach out anytime.

Thanks again for your effort and time!

Best,
Kai

card-settings
card-details-cardnumber
minicard-cardnumber

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3935)

### DIFF
--- a/client/components/cards/cardDetails.jade
+++ b/client/components/cards/cardDetails.jade
@@ -25,6 +25,9 @@ template(name="cardDetails")
         h2.card-details-title.js-card-title(
           class="{{#if canModifyCard}}js-open-inlined-form is-editable{{/if}}")
             +viewer
+              if currentBoard.allowsCardNumber
+                span.card-number
+                  | ##{getCardNumber}
               = getTitle
             if isWatching
               i.card-details-watch.fa.fa-eye
--- a/client/components/cards/cardDetails.styl
+++ b/client/components/cards/cardDetails.styl
@@ -114,6 +114,12 @@ avatar-radius = 50%
     background: darken(white, 7%)
     border-bottom: 1px solid darken(white, 14%)
 
+    .card-number {
+      color: darken(white, 30%);
+      display: inline-block;
+      margin-right: 5px;
+    }
+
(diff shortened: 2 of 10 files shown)


mfilser:
This is a good idea, thanks for your work.

What's just came into my mind: What happens if cards are moved to and from the board? Will this interrupt the numbering?

syndimann:
@xet7 thanks for merging !

@mfilser that's a good question. i personally would say the card numbering needs to be boardwise unique but not necessarily gap-free. So when a card is moved to another board the moved card should be assigned the nextCardNumber from the new board. This is not implemented yet. I will test this and provide another PR for this

mfilser:
@syndimann 
Thanks for your reply.
I didn't test it too and also didn't have a deep look into the change, it's just what i thought after a "light" look at the code, especially of this function:
https://github.com/wekan/wekan/blob/def2ec46e026e8345096abb7929f974f8fd3b24e/models/boards.js#L1068-L1069

Maybe it's also worth to test after 1 card is moved to another board and a new card is added if the created number exists twice.

Best regards

syndimann:
@mfilser good points, will definitely have a look at that !

aklabsus:
Is there a way to filter or search by card number?  As those numbers almost useless when we can't search by a number.

PR #3936, opened 2021-08-03
Fix: Consecutive Card numbering when a card is moved to another board or copied

syndimann:
Hey @xet7 and @mfilser ,

because of @mfilser very good remarks on my previous [PR](https://github.com/wekan/wekan/pull/3935) regarding the consecutive boardwise card numbering i submit another PR that includes the following fixes:

- If a card is moved to **another** board it, the card will receive the next card number from the target board. 
- If a card is copied to the same or another board, it will always receive the next card number from the target board

This PR also includes enhancements for `Boards.getNextCardNumber()`. It basically works now the following way:
1. return `1`, If no card is assigned to the board, 
2. return `1`, If the board has cards where **none** of the cards have the attribute `cardNumber` (this should never happen due to the migration-script `assign-boardwise-card-numbers` that is running prior to launch)
3. return `highestCardNumber + 1`, if the board has cards with the attribute `cardNumber`

If there is anything else i should need to take care of please do not hesitate to add further comments to this PR.

As mentioned in the other PR, i think the cardNumbers should continuously increase but can contain caps within a board. this happens if for instance a card that does not have the highest cardNumber in a board is moved to another board.

Thank you both for taking the time

Best
Kai

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3936)

### DIFF
--- a/models/boards.js
+++ b/models/boards.js
@@ -1065,8 +1065,23 @@ Boards.helpers({
   },
 
   getNextCardNumber() {
-    const boardCards = Cards.find({ boardId: this._id }).fetch();
-    return boardCards.length + 1;
+    const boardCards = Cards.find(
+      {
+        boardId: this._id
+      },
+      {
+        sort: { cardNumber: -1 },
+        limit: 1
+      }
+    ).fetch();
+
+    // If no card is assigned to the board, return 1
+    if (!boardCards || boardCards.length === 0) {
+      return 1;
+    }
+
+    const maxCardNr = !!boardCards[0].cardNumber ? boardCards[0].cardNumber : 0;
+    return maxCardNr + 1;
   },
 
   cardsDueInBetween(start, end) {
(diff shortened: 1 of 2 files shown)


mfilser:
@syndimann 
Great, thank you too for this :slightly_smiling_face: 

Just another idea / thinking:
I don't know if it's faster, but maybe it's more performant to store the current card number at the board data, take this card number and increase it instead to make a "big query". A query could cost more time on boards with many cards. Just an "inspiration" of possibilities :wink:

syndimann:
@mfilser you are absolutely right and while i implemented the fix i thought about that too. in terms of performance this is not the best solution. my intention was to provide a quick fix for the move and copy issue and from there we could aim for a solution with a better performance. i will definitely take a look at this and provide another pr as soon as the implementation is finished.

thanks a lot for taking my contributions into account!

mfilser:
@syndimann 
So we thought the same :)
Your fix is absolutely well done, better a quick fix than a bigger taking  longer fix :+1:

PR #3945, opened 2021-08-13
Feature: Adds Card Comment Reactions

syndimann:
Hi @xet7 

This PR adds card comment reactions to wekan. Reactions to comments are a useful feature to acknowledge, verify, approve or deny actions taken by assigned members and to give feedback in an easy way.

the feature was implemented by
- introducing a new mongo collection `card_comment_reactions`
- adds a starting set of unicode codepoints as available reactions (see image attached)
- introduce new templates in the activities template
- whenever someone hovers over a reaction, the tooltip shows all people that reacted in the same way (see image attached)
- added the i18n title to all files for key `addReactionPopup-title` (i tried my best to translate this in as many languages as i was able) thus the great number of changed files (i've split this in to commits)
- all reactions i logged in user has chosen are display in blue

all users that are board members are able to add reactions to comments

attached images:

add-reaction
show-reactions

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3945)


xet7:
@syndimann 

Thanks!

You only need to add new strings to be translated to `wekan/i18n/en.i18n.json` that [I push](https://github.com/wekan/wekan/blob/master/releases/translations/push-english-base-translation.sh) to [Transifex](https://transifex.com/wekan/wekan) and [later pull](https://github.com/wekan/wekan/blob/master/releases/translations/pull-translations.sh) and commit before making new Wekan release. You don't need to do any translations, Wekan is already translated by wordwide translators that have registered to Wekan Transifex project.

Anyone can translate any language in Wekan, and request new languages:
- `en` is source language, same as `en_US`
- `language` and `language_COUNTRY` are same, only one of those is at Transifex, like `fi_FI`

syndimann:
@xet7 thanks for the info ! i didn't know that. that is a great process. i will consider this the next time i introduce new titles :)

xet7:
Usually when some new strings are added to Wekan, within 2-5 days most translators have translated all new strings.

PR #3964, opened 2021-08-25
[All boards page] Possibility of filtering board by team or organization

Emile840:
A list of teams and organizations to which connected user belongs, will be displayed on « All boards page » then a list of displayed boards can be filtered by selecting one or more  teams « using Ctrl + select » (see gif image below)
Filter boards by teams or organizations

### DIFF
--- a/client/components/boards/boardsList.jade
+++ b/client/components/boards/boardsList.jade
@@ -1,11 +1,32 @@
 template(name="boardList")
   .wrapper
+    ul.AllBoardTeamsOrgs
+      li.AllBoardTeams
+        if userHasTeams
+          select.js-AllBoardTeams#jsAllBoardTeams("multiple")
+            option(value="-1") {{_ 'teams'}} :
+            each teamsDatas
+              option(value="{{teamId}}") {{_ teamDisplayName}}
+
+      li.AllBoardOrgs
+        if userHasOrgs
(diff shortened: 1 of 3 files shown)


Seven26:
why everyone can see that, i think it should can be controlled by access right, because as administator, i will assign different permissions, and someone just need to see their team information, and i didnt assign any teams/organizations for them. but now they can see this function and they can't use it (or didn't need to use it) thanks.

Emile840:
> why everyone can see that, i think it should can be controlled by access right, because as administator, i will assign different permissions, and someone just need to see their team information, and i didnt assign any teams/organizations for them. but now they can see this function and they can't use it (or didn't need to use it) thanks.

Is my comment (second point of the pull request #3983) responds to your question ?

xet7:
Yes, it's fixed at https://github.com/wekan/wekan/pull/3983 . Thanks!

Seven26:
> > why everyone can see that, i think it should can be controlled by access right, because as administator, i will assign different permissions, and someone just need to see their team information, and i didnt assign any teams/organizations for them. but now they can see this function and they can't use it (or didn't need to use it) thanks.
> 
> Is my comment (second point of the pull request #3983) responds to your question ?

> will check after upgraded,thanks.

Seven26:
checked after upgraded, it was fixed. thanks a lot.

PR #3976, opened 2021-08-27
[Admin panel] Added a parameter to display or not the visibility of a board in private mode only

Emile840:
I added a setting in admin panel page, so that an admin user can activate or not a possibility of creating board in private mode only (see gif image below)

parameter for board private visibilility

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3976)


xet7:
Actually, I think using Meteor subscription is OK.

Issue #2444, opened 2019-06-04
Bug: First User (Admin) Login/Registration with OAuth2&OIDC fails


DominikPf:
This Issue was detected in #2412 and is also documented here:
https://github.com/wekan/wekan/wiki/Keycloak
or
https://github.com/wekan/wekan/issues/1874#issuecomment-460802250

If a fresh Wekan installation with OAuth2 is used, the first User cannot be generated on Registration. So Wekan cannot be used.

This Bug happens only for the first User (if you create the User via Username/Password and than change to OAuth2 everything works fine).

This Issue also seems to happen with LDAP Login (we haven't checked this again)

ww-daniel-mora:
I am also running issues with first user creation with OAuth2. User creation appears to succeed but no profile data (user name, email etc) are populated, and cannot be be manually entered afterwards. The client just errors when encountering surprising null values like email and will not save updates.

While attempting to debug the issue I noticed that [Accounts.onCreateUser](https://github.com/wekan/wekan/blob/76f4b8316b52a81aadfc1d4759447f41d002e6f9/models/users.js#L1273) did not appear to be called during first user creation.

xet7:
Fixed at https://github.com/wekan/wekan/pull/3994

PR #3994, opened 2021-09-05
Fix first user creation for via OIDC

ww-daniel-mora:
When setting up a new deploy and registering the first user via OIDC no profile information would be populated. This also prevented updating profile information as the client would error when encountering a null email value. I manually tested this change against the OIDC use case, but, not others. This fix should address #2444

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/3994)


ww-daniel-mora:
👏 👏  Thank you @xet7 !

PR #4031, opened 2021-10-05
Allow setting custom kubernetes labels when using the helm chart

ariep:
This adds two helm values to the helm chart, for setting custom kubernetes labels on the `Deployment` object, and on the `Pod`s created by the deployment.

We need this when using the backup system velero, to specify which components need to be restored.

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/4031)

### DIFF
--- a/helm/wekan/templates/deployment.yaml
+++ b/helm/wekan/templates/deployment.yaml
@@ -8,6 +8,9 @@ metadata:
     component: wekan
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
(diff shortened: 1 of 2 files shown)


ariep:
I wasn't sure whether to bump the chart version as part of this PR, or if you have a separate procedure for doing that. Please let me know if I should add that version change here.

xet7:
@ariep 

I bump versions when releasing new Wekan version. PRs do not include version number changes.

Issue #3830, opened 2021-05-31
[Login page] The default wekan logo appears briefly before displaying the custom logo


Emile840:
## Issue
![wekanLogo](https://user-images.githubusercontent.com/83423044/120175029-0edeb200-c206-11eb-8af7-cad50ab48663.gif)

Each time you arrive on the login page, there is a call to the database to retrieve the “setting” object which contains the configuration data (including the link to custom logo), the problem is that there is a delay of a few seconds between the first display of the page and the call to the database. For those few seconds wekan default logo is displayed (see gif image above). I'm working on this issue

xet7:
Moved to #3832

PR #4057, opened 2021-10-18
add full name if exists in email-invite-subject or when tagging someone with '@' while commenting a card

Emile840:
A username is used on e-mail notification or when tagging someone with '@' while commenting on a card. But sometimes user name can be meaningless for Wekan users (IGA5594 for example). Thus, i added a full name if exists in email-invite-subject params when sending e-mail notification or when tagging a user with '@' while commenting on a card. 

I also fixed regression issues for below points:
- point #3830 (I don't know why there were no longer the codes used to correct this point in the latest versions)
- point #3983

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/4057)

### DIFF
--- a/client/components/main/layouts.jade
+++ b/client/components/main/layouts.jade
@@ -34,7 +34,7 @@ template(name="userFormsLayout")
           img(src="{{currentSetting.customLoginLogoImageUrl}}" width="300" height="auto")
           br
       unless currentSetting.customLoginLogoImageUrl
-        if isSettingDatabaseFctCallDone
+        div#isSettingDatabaseCallDone
           img(src="{{pathFor '/wekan-logo.svg'}}" alt="" width="300" height="auto")
           br
       if currentSetting.textBelowCustomLoginLogo
--- a/client/components/main/layouts.js
+++ b/client/components/main/layouts.js
@@ -23,7 +23,7 @@ const validator = {
   },
 };
 
-let isSettingDatabaseFctCallDone = false;
+// let isSettingDatabaseFctCallDone = false;
 
 Template.userFormsLayout.onCreated(function() {
   const templateInstance = this;
@@ -40,7 +40,11 @@ Template.userFormsLayout.onCreated(function() {
         oidcBtnElt.html(htmlvalue);
       }
 
-      isSettingDatabaseFctCallDone = true;
+      // isSettingDatabaseFctCallDone = true;
+      if(currSetting && currSetting !== undefined && currSetting.customLoginLogoImageUrl !== undefined)
+        document.getElementById("isSettingDatabaseCallDone").style.display = 'none';
+      else
+        document.getElementById("isSettingDatabaseCallDone").style.display = 'block';
       return this.stop();
     },
   });
@@ -69,9 +73,9 @@ Template.userFormsLayout.helpers({
     return Template.instance().currentSetting.get();
   },
 
-  isSettingDatabaseCallDone(){
-    return isSettingDatabaseFctCallDone;
-  },
+  // isSettingDatabaseCallDone(){
+  //   return isSettingDatabaseFctCallDone;
+  // },
 
   isLoading() {
     return Template.instance().isLoading.get();
--- a/client/components/main/layouts.styl
+++ b/client/components/main/layouts.styl
@@ -542,3 +542,6 @@ a
 
   100%
     transform: rotate(360deg)
+
+#isSettingDatabaseCallDone
+  display: none;


saschafoerster:
@Emile840 Does tagging with @ still work for you? When I type enter, then the full name is copypasted, which then is not the correct username, so the tagging is broken.

Emile840:
@saschafoerster 
Tagging with @ is working for me (it just add a full name in parentheses).

saschafoerster:
Maybe this visual shows better, what I mean? Tagging feels broken to me with @(NAME)name and not just @name. Also CSS is broken, because it doesn't show like before, that the tagging was successful. Sorry, I am not able to change it myself. 
Bildschirmfoto 2021-12-26 um 20 44 58

Ben0it-T:
@saschafoerster 
Thanks for the screenshot. I didn't quite understand what the problem was... and had not yet tested this feature.
I think this is a _simply_ regex issue.

https://github.com/wekan/wekan/blob/2ac5f0b2d71cd057b9f2672ca16723648fa14b5a/client/components/main/editor.js#L358-L367

https://github.com/wekan/wekan/blob/2ac5f0b2d71cd057b9f2672ca16723648fa14b5a/models/activities.js#L205-L213

Actually only `@username` is working to get `username` and and trigger/fire a notification.
I'm not a regex ninja.
Playing a bit on fiddle I probably got something...
```js
// original regex : /\B@([\w.]*)$/
let mentionRegex = /\B@[\w.\s]+\(([\w.]*)\)|\B@([\w.]*)/gi;
let mentions = "@foo @foo(bar) @foo (bar) @foo.bar @foo.bar(foo.bar) @foo bar(foo.bar) @foo.bar (foo.bar) @foo bar (foo.bar)";
let currentMention;

while ((currentMention = mentionRegex.exec(mentions)) !== null) {
  const [fullMention, quoteduser, simple] = currentMention;
  let str = quoteduser||simple;
  document.write("currentMention: " + currentMention + "<br>");
  document.write("username: " + str + "<br>");
  document.write("<hr>");
}
```

https://jsfiddle.net/Ben0it_T/jztoapfk/

But this can have some sideboard effects
https://jsfiddle.net/Ben0it_T/atnb3cms/

Need to dig more but I think we have something.
Can an expert take a look ?

Ben0it-T:
@xet7 
Me again, sorry...
I can't get my hands on the allowed characters for `username`... or a function/method to check `username` requirements
Could you tell me which characters are allowed _(or forbidden)_ in `username` ?

saschafoerster:
I guess we have a missunderstanding. I don't like the mentioning in this way @ FILL NAME (USERNAME),, I just would like the simple @USERNAME back, like it is here in Github and I guess everywhere else.

Ben0it-T:
@saschafoerster
> But sometimes user name can be meaningless for Wekan users (IGA5594 for example). Thus, i added a full name if exists in email-invite-subject params when sending e-mail notification or when tagging a user with '@' while commenting on a card.

I think that @Emile840 is right
For mentions, maybe a compromise could be `@username (full name)`.
Or `@mention` and a tooltip (_with the full name_)?



@xet7 
If we look at the different regex relating to user mentions
https://github.com/wekan/wekan/blob/2ac5f0b2d71cd057b9f2672ca16723648fa14b5a/client/components/main/editor.js#L12-L14

https://github.com/wekan/wekan/blob/2ac5f0b2d71cd057b9f2672ca16723648fa14b5a/client/components/main/editor.js#L358

https://github.com/wekan/wekan/blob/2ac5f0b2d71cd057b9f2672ca16723648fa14b5a/models/activities.js#L205

https://github.com/wekan/wekan/blob/2ac5f0b2d71cd057b9f2672ca16723648fa14b5a/sandstorm.js#L219-L226

`username` should contain 
- any word character => `[a-zA-Z0-9_]`
-  a dot `.`  
- a whitespace character (_personally, I'm not a fan, but it's like that!_)

It will probably be necessary to add hypen `-` in the future (_not really urgent_)

saschafoerster:
> Or @username and a tooltip (with the full name)?

Please, yes! A tooltip is what I would expect. Just see what Github does, when you hovering over your Username @Ben0it-T. It's what I would expect as a user from a lot of other experiences.

xet7:
Username can also be email address, like when using OAuth2/LDAP2 etc.

xet7:
For example, WeKan login to Google https://github.com/wekan/wekan/wiki/Google-login to work correctly, it is required that username is email address.

Sure there could be another field like `realusername` or alternatively some way to convert email address to username. And also look that there is no duplicate username.

Ben0it-T:
Need to dig deeper for the integration of a tooltip.
Technically, it must be possible. But what bothers me is that we will no longer show the `fullname` in the comment... and it will not appear in an Excel export either...

Ben0it-T:
@saschafoerster
Small feedback
I did some tests but I am not satisfied ... and not sure that it brings anything
username-tooltip



Honestly the cleanest is to let `@username (fullname)`

saschafoerster:
@Ben0it-T Thank you for trying out. I don't really have a better idea. I thought about maybe turning it around: 
you start a mention with @, then you see the @USERNAME (Fullname) list, and when you enter the tag you see the FULLNAME with markup and the @username as overlay. But some users don't have full names, so this might break it again. 

Maybe we leave it as it is. I mostly missed the markup with underline and background (so you know you tagged successfully) and I missed the messages. With features seem restored. :) 

Let's close it until someone else has a better idea how to do it. :)

PR #4073, opened 2021-10-25
Labels wrapped if too long

mfilser:
Long labels on card and minicard are wrapped if too long

---
This change is [Reviewable](https://reviewable.io/reviews/wekan/wekan/4073)

### DIFF
--- a/client/components/cards/minicard.styl
+++ b/client/components/cards/minicard.styl
@@ -80,8 +80,6 @@
 
   .minicard-labels
     float: none
-    display: flex
-    flex-wrap: wrap
 
     .minicard-label
       width: 11px


Ben0it-T (commented):
@mfilser 
Labels in minicard are displayed verticaly when "Hide minicard label text" is checked.
Could you have a look at this? I think it's because of change in client/components/cards/minicard.styl
Thanks

was :
Labels in minicard - 01

now : 
Labels in minicard 02

mfilser:
@Ben0it-T
Thank you for the screenshots. Sure, i will have a look at this

mfilser:
@Ben0it-T
I hope it's fixed now at #4109

Ben0it-T:
I tested on my environment. It's great !
Thanks !

mfilser:
Great. Thank you for testing and your feedback

xet7:
Fixed at https://github.com/wekan/wekan/pull/4109

dsvk:
@mfilser I noticed this change also added a [solid black border to full-size labels](https://github.com/wekan/wekan/pull/4073/files#diff-522745edbadb5ff5d7c84405287d1f671d02dd13241c93aa9c3161f536b30244R5), is that actually required by something or could it be a development/testing leftover?

A solid border, in my opinion, seems out of place (see attached screenshots):
- solid black results in inconsistent look when combined with the wide variety of label colors (almost invisible on dark ones while instead very striking on light ones)
- no other UI elements are solid-bordered (cards themselves, codeblocks, start/due/end date pickers...)
- (Firefox only): the added borders appear to cause a spacing overflow bug in the label creation/editing dialog - which gets fixed if I remove the label border with a CSS override (`.card-label { border: 0px !important; }`)

Current look, also showing the Firefox overflow bug:
current

With the border removed (no overflow bug, IMHO also a much cleaner design better fitting the rest of the UI):
previous

From my basic testing with the CSS override in place, removing the border doesn't appear to break anything and long labels get wrapped across multiple lines with no visible issues.

Thanks.

mfilser:
I can have a look at the "overflow bug", i didn't reproduce it yet.
The border depends in my opinion on the color, here in your screenshot the border has a good contrast at the white color, most color's are more brighter. But as always the design depends on the user...

mfilser:
@dsvk 
I can't reproduce the FireFox button "bug" you sent, i tested with FireFox 95. I only can reproduce the bug if i manually edit the css property `width: 10.5%` to `width: 11.5%`, but as i said, only if manually editing at the developer console.

dsvk:
@mfilser just found out the overflow bug is reproducible on 90% browser zoom (which I use Wekan on to reduce the somewhat oversized fonts compared to pretty much every other website or app). Chromium seems to display the picker grid correctly even on 90% zoom though.

Inconsistency of the contrast might indeed depend heavily on which label colors you use, however I do have a pretty even split between light and dark labels and it gets quite bothersome visually. It also gets even worse when two light labels next to each other have inverted font colors:

labels

The primary issue I think, though, is that the (visually harsh) solid borders don't seem to be in line with the overall UI design choices. Everything else is either borderless or at most a light gray split line or shadow effect. Trello doesn't seem to use harsh black borders (on labels or anywhere else) either.

mfilser:
I tried with 90% and even lower browser zoom, no chance to get the same behaviour with Firefox or Chrome. Strange. I keep testing to reproduce it.

dsvk:
And I just reproduced it on a completely new FF95 profile, no addons. At this point I'm concerned it might be some platform-specific rendering thing (Firefox on Linux).

With a 3-decimal accuracy: `width: 10.494%` is where the grid gets fixed at 90% zoom, `10.398%` for 80%, `10.267%` for 70%...not sure what to make of that. Zoom levels 100%+ are all fine.

Simply removing the border completely fixes all zoom levels all the way to the smallest available 30%.

PR #4124, opened 2021-11-08
change scroll bar design

saurabharch:
Please Merge this scroll native support scroll design 

it supports chrome, firefox, opera browsers


xet7:
That `wekan.html` is always overwritten with this generating script:
https://github.com/wekan/wekan/blob/master/releases/rebuild-docs.sh

You should make new pull request that actually also affects generated `wekan.html` file.

PR #4262, opened 2021-12-25
Fix getLabels exception in template helper

Ben0it-T:
Fix getLabels exception in template helper when board has no labels
Fix getLabels error when boards.labels.name doesn't exists

(refers to https://github.com/wekan/wekan/pull/342 and https://github.com/wekan/wekan/commit/5d77ad4f6ba70038486d734e97844c547e664e88)


Ben0it-T:
I did not know this feature before (@user and #label use when creating new cards).
It's really great !!

PR #4271, opened 2022-01-02
added endpoint to query username for a specific ID

### LINKED ISSUE #4230: API Feature: get user name from ID
## Feature request

**Server Setup Information**:

* Did you test in newest Wekan?: Yup (version 5.83.0)
* For new Wekan install, did you configure root-url correctly so Wekan cards open correctly https://github.com/wekan/wekan/wiki/Settings ? yup
* Operating System: Hosted on a Synology NAS
* Deployment Method(Snap/Docker/Sandstorm/bundle/source): Docker
* Node.js Version: 12.22.7
* MongoDB Version: 4.4.9-rc0

**Feature description**:

I am currently building a client to be able to use WeKan in a very personalized way. I was able to find the REST API documentation and use it without any problems.

However, I don't see any route to get a user's name from their ID. The only similar route seems to be `GET /api/users/...`, but it is reserved for administrators (according to the doc), so it is not suitable for my use.

I suggest adding a `GET /api/users/.../username` route or equivalent to get a user's name from his ID in a simple and secure way. This avoids opening the admin route to everyone, but the username seems to me to be important to get from any user.

If a similar route exists and I missed it, please let me know, I sometimes miss obvious things :)

lindhork:
This should fix #4230


ZRunner:
Could we consider getting a correct "public profile" instead of just the username? It might be a bit more annoying to do so I won't blame you if you don't, I'm just suggesting lol

If I'm right it would give something like
```js
let user = Meteor.users.findOne({
    _id: id,
});
let data = {
    username: user.username,
    isAdmin: user.isAdmin,
    avatarUrl: user.profile.avatarUrl,
    fullname: user.profile.fullname, // might be too sensitive? but the client uses it so...
    initials: user.profile.initials
}
```
And then the endpoint might be `/api/users/:userId/profile` (or `/api/users/:userId/info` as profile may be ambiguous with the user attribute).

lindhork:
I think returning the full name is not good if every user should be able to use this endpoint. I could imagine that if an administrator uses this endpoint, we also return the full name and just omit it for a non-admin user ... or maybe add an extra setting to be able to determine the behavior?

I adjust the pull request immediately, so that besides the username also the other info is returned. Will use `/api/users/:userId/info`.

I leave the full name out for now. Gladly @xet7 can say what he would prefer.

xet7:
@lindhork 

There will be mobile and desktop apps that only use API. 

For that query, there should not be searching from all users.

Instead, there should be searching from boards:
- where both API user and member user exists and are active
- those board members are not part of any team or organization

https://github.com/wekan/wekan/blob/master/models/boards.js#L659
```
member => member.userId === userId && member.isActive,
```
Instead of isAdmin, API could give actual permission of user, similar like WeKan UI.

I would think fullname can be in API.

For teams and organizations, there could be API for listing what organizations and teams are at those boards, where API user is member. But not listing of those organizations and teams members.

Sometime later there will be features, that boards can have teams and organization members that are non-active users #4265

lindhork:
@xet7
Just to make sure I understood you correctly:

- The endpoint should be moved to e.g. `/api/boards/:boardId/members/info`.
- It should be checked if the requesting user is assigned to this board and activated.

I would suggest that the endpoint then simply returns the info for all users on the board and there is another endpoint that can also query just one user?

> Instead of isAdmin, API could give actual permission of user, similar like WeKan UI.

This is a very good idea

What exactly do you mean by that:
> For teams and organizations, there could be API for listing what organizations and teams are at those boards, where API user is member. But not listing of those organizations and teams members.

Another endpoint (e.g. `/api/boards/:boardId/members/organizations`) which then returns only organizations and another one (e.g. `/api/boards/:boardId/members/teams`) which then returns only assigned teams? Considering that the requesting user is himself in the organization or the team.
What if the requesting user is an admin, should all organizations or teams assigned to the board be returned?

I'll wait to implement it until we've discussed it further.

lindhork:
> and there is another endpoint that can also query just one user?

Instead of another endpoint we could also use a parameter in the body, if this is set only the information for this member is returned?

PR #4301, opened 2022-01-19
fix notification drawer always on top

Azorimor:
# Old
Currently the notification drawer can be behind other components, which can be annoying.

pre_fix_notification-drawer

# New

The notification drawer is now always on top. As far as I tested, the changes did not break anything on the web version of wekan.

after_fix_notification-drawer


xet7:
Merged at https://github.com/wekan/wekan/commit/bec25720da947e048a63541b4e421bd9bd846c58

Thanks!

Issue #4322, opened 2022-01-25
List not movable when lists require scrolling


gbmd:
## Issue
Lists (Especially on the right) are not movable when there are too many lists to fit the screen.
They become movable again when the left end of the board is visible again. (Extreme Zoom-out)
This could be related to Issue: https://github.com/wekan/wekan/issues/4219

### Server Setup Information
![image](https://user-images.githubusercontent.com/54003582/150948423-83f14e5a-9eff-4066-a199-7d147de083bf.png)
* Operating System:
* Deployment Method Docker
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first):
* Node.js Version:
* MongoDB Version:
* What webbrowser version are you using: Chrome 97.0.4692.99

### Problem description
note: M is last Board
![Peek Wekan](https://user-images.githubusercontent.com/54003582/150949751-778da53f-df65-403b-af8a-fe189dc3ccdb.gif)
#### Reproduction Steps
Can be reproduced by creating a Board with 13 Lists (A-M), scrolling to M and trying to switch M and L

#### Logs
no console messages

mfilser:
I will have a look at this

Issue #4219, opened 2021-12-07
Dragging Swimlanes still does not work with 28 lanes


doniels:
## Issue

We cannot drag swimlanes on the board with 28 swimlanes. Not new swimlanes and not old ones. 

Basically, https://giters.com/wekan/wekan/issues/4098 still happens by us - Using  Windows with Chrome Version 96.0.4664.45 .

On the smaller board with only 12 swimlanes we can move swimlanes.

We have the latest version of WeKan
![image](https://user-images.githubusercontent.com/57865526/145015812-884b7984-4772-48b8-ab6a-2bb6e7582554.png)

xet7:
@doniels 

Please check is your server swapping, with:
```
free -h
```
If it is, please increase amount of total RAM to 4 GB.

Then please add comment about did it help.

xet7:
@doniels 

Please also check that there is enough free disk space:
```
df -h
```

doniels:
@xet7 - thanks for the reply. Here are the results.

> free -h

We were at 2Gb - we upgraded to 4Gb.

> df -h

```
              total        used        free      shared  buff/cache   available
Mem:           1.9G        907M        202M        1.3M        845M        871M
Swap:            0B          0B          0B
```

Results: Still cannot drag the 28th swimlane to the left.

**But** - I managed to drag swimlane #7 to the extreme right - and then drag  the new swimlane to its correct location (#8) - and them move then 7th swimlane back to its correct location (#7).

I did not try this before we increased the RAM.

So, something is still broken - with 28 swimlanes the dragging is not working properly.

xet7:
Ok, I added this to [Roadmap](https://boards.wekan.team/b/D2SzJKZDS4Z48yeQH/wekan-r-open-source-kanban-board-with-mit-license/FXPmk8peKFPP5s7mR). It will be fixed in some future version of WeKan.

er0gl:
I got the same problem with the List view

PR #4330, opened 2022-02-03
Lists now moveable when swimlane requires scrolling

mfilser:
Fixes: #4322


er0gl:
Thanks for your work @mfilser & @xet7 !!! This also fixes #4219 for me!

PR #4359, opened 2022-02-17
Update de.i18n.json

PietsHost:


xet7:
@PietsHost 

Non-English translations are done at https://transifex.com/wekan/wekan , please join there and add changes there. Thanks!

PR #4377, opened 2022-02-23
Feature/shortcuts for label assignment

Viehlieb:
Adds possibility to assign Labels via shortcuts. Numbers 1-9 will toggle Labels 1-9 as long as they exist.
Shortcut is available while:
- Hovered Card
- Multiselected Cards (numbers 1-9 will add; shift+[1-9] will delete labels])
- Current Card


ClaudiaK21:
Shift + Number on numpad is "jump to bottom of the page"
Shift + Number main keyboard numbers works (only when in multiselect mode: Shift  + click card)

I am missing a shortcut to also remove the the label 1-9 on a single card

Viehlieb:
> 

It is true. The numpad and number keys on main part of the keyboard have been confused. 
To be clear again: Number Keys are used for label assignment/removal.

It should be possible to **hover** over the **card** and toggle labels 1-9 as well as **seĺect** the **card** and use the NUMBER KEYS 1-9 
Thks for feedback

PR #4392, opened 2022-03-04
Feature/empower sso oicd data propagation

Viehlieb:
#### Fix weird behavior, which leads to inconsistent db state:
Creating a Team or Org Active: true resulted in user interface shown as active: No

#### More Power for Oidc providers
As described in packages/wekan-oidc/README.md, now it is possible to configure oidc provider - e.g. authentik - to trigger org or team creation via oidc.

**Scenario:**
0. Group created in Oidc Provider, with supported attributes:
    desc: groupDesc           _// default group.name_
    isAdmin: true           _// default "false", sets admin attribute in Wekan_
    website: groupWebsite     _// default group.name_
    isActive: true          _// default "false"_
    shortName: groupShortname // _default group.name_
    forceCreate: true       // _default "false"_
    isOrganisation: true    // _default "false"_

**Scripts changed and why**

Fix weird behavior of teams/orgs creation:
client/components/settings/peopleBody.jade

Functionality for oidc group data propagation:
packages/wekan-oidc/oidc_server.js
packages/wekan-oidc/loginHandler.js
models/team.js (method for creation through oidc)
models/org.js ((method for creation through oidc)

Explanation Setup with authentik:
README.md

### DIFF
--- a/client/components/settings/peopleBody.jade
+++ b/client/components/settings/peopleBody.jade
@@ -148,30 +148,30 @@ template(name="newUserRow")
 template(name="orgRow")
   tr
     if orgData.orgIsActive
-      td <s>{{ orgData.orgDisplayName }}</s>
-    else
       td {{ orgData.orgDisplayName }}
-    if orgData.orgIsActive
-      td <s>{{ orgData.orgDesc }}</s>
     else
+      td <s>{{ orgData.orgDisplayName }}</s>
+    if orgData.orgIsActive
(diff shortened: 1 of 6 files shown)


Viehlieb:
I'll update this shortly to adress following issues:

- updating teams/orgs through oidc not possible
- loginHandler contains function does not deal with userObjs being 'undefined' in if clause
- if user is not already in mongodb collection 'users' then assignment and teams/orgs creation will apply only on second login

Issue #4395, opened 2022-03-05
"My Cards" Section not showing my cards


luistiktok:
Hello, I'm using Wekan docker version 6.11.0 on Ubuntu 20.04. For some reason I cannot seen any card assigned or created by me under "My Cards" section. Myabe i'm missing something or is a bug. Thank you.

subashsn:
Cards not showing up in My Cards or Due Cards also

Same. Tried assigning it, due date, etc.

subashsn:
**Confirmed** that it **works in v6.09**, but **doesn't work since 6.10**. 

Error log when trying to view My cards:
```
wekan_1    | Exception from sub myCards id aooronqbjJ4rzCNHu Error: Publish function returned an array of non-Cursors
wekan_1    |     at Subscription._publishHandlerResult (packages/ddp-server/livedata_server.js:1101:20)
wekan_1    |     at Subscription._runHandler (packages/ddp-server/livedata_server.js:1063:10)
wekan_1    |     at Subscription.MeteorX.Subscription._runHandler (packages/lamhieu_unblock.js:68:22)
wekan_1    |     at Session.MeteorX.Session._startSubscription (packages/lamhieu_unblock.js:56:7)
wekan_1    |     at Session.sub (packages/ddp-server/livedata_server.js:625:12)
wekan_1    |     at Session.MeteorX.Session.protocol_handlers.sub (packages/lamhieu_unblock.js:27:15)
wekan_1    |     at packages/ddp-server/livedata_server.js:559:43
```

Error log when trying to view My cards:
```
wekan_1    | Exception from sub globalSearch id vPA2b3vY3Bhmc7HpA Error: Publish function returned an array of non-Cursors
wekan_1    |     at Subscription._publishHandlerResult (packages/ddp-server/livedata_server.js:1101:20)
wekan_1    |     at Subscription._runHandler (packages/ddp-server/livedata_server.js:1063:10)
wekan_1    |     at Subscription.MeteorX.Subscription._runHandler (packages/lamhieu_unblock.js:68:22)
wekan_1    |     at Session.MeteorX.Session._startSubscription (packages/lamhieu_unblock.js:56:7)
wekan_1    |     at Session.sub (packages/ddp-server/livedata_server.js:625:12)
wekan_1    |     at Session.MeteorX.Session.protocol_handlers.sub (packages/lamhieu_unblock.js:27:15)
wekan_1    |     at packages/ddp-server/livedata_server.js:559:43
```

[...]

PR #4405, opened 2022-03-10
Attachments fixes after migration to meteor files (image preview, global search)

mfilser:
Attachments fixes after migration to meteor files (image preview, global search)

- global search was broken
- image preview is back again
- wrong mongo index was created
- attachment count on minicard was missing

should also fix: #4395

### DIFF
--- a/client/components/cards/attachments.jade
+++ b/client/components/cards/attachments.jade
@@ -20,7 +20,7 @@ template(name="attachmentsGalery")
     each attachments
       .attachment-item
         a.attachment-thumbnail.swipebox(href="{{link}}" title="{{name}}")
-          if isUploaded
+          if link
             if isImage
               img.attachment-thumbnail-img(src="{{link}}")
             else if($eq extension 'mp3')
@@ -38,7 +38,7 @@ template(name="attachmentsGalery")
             else
               span.attachment-thumbnail-ext= extension
           else
-            span.attachment-thumbnail-ext= extension
+            +spinner
         p.attachment-details
           = name
           span.file-size ({{fileSize size}} KB)
--- a/client/components/cards/minicard.jade
+++ b/client/components/cards/minicard.jade
@@ -127,10 +127,10 @@ template(name="minicard")
           span.badge-icon.fa.fa-check(class="{{#if pokerState}}text-green{{/if}}")
           if expiredPoker
             span.badge-text {{ getPokerEstimation }}
-      if attachments.count
+      if attachments.length
         .badge
           span.badge-icon.fa.fa-paperclip
-          span.badge-text= attachments.count
+          span.badge-text= attachments.length
       if checklists.count
         .badge(class="{{#if checklistFinished}}is-finished{{/if}}")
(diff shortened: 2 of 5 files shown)


BabyFnord:
@mfilser and @xet7 though stated otherwise, this did not fix https://github.com/wekan/wekan/issues/4393 unfortunately. We would be happy, if you could have another look into it. Thank you!

mfilser:
@BabyFnord
I think your issue is completly different from this pull request. I don't have any experience with uberspace deployment, maybe the migration wasn't completed at your installation. As i see uberspace is commercial you should consider to contact the support of uberspace...

BabyFnord:
@mfilser Apparently @marek94901 has [the same issue](https://github.com/wekan/wekan/issues/4393#issuecomment-1065189981) on a presumably non-Uberspace installation. @marek94901 may you can shed light on your install specs here, for further reference? WeKan deployment on Uberspace is on behalf of every user her-/himself, and largely follows [this bundle based installation guide](https://lab.uberspace.de/guide_wekan.html).

My WeKan instance never had this issue before, starting from v5.17 through (almost) every iteration. The changes introduced to v6.11 broke existing and new attachments. How should Uberspace support know what WeKan's issues are in this transition? If you do know what to ask or tell them, please tell me. @marek94901 probably would be glad to get pointers too, for his instance. 

Switching Node main versions is just `sending uberspace tools version use node 14` which renders
```
Selected node version 14
The new configuration is adapted immediately. Patch updates will be applied automatically.
```
Apart from that, nothing else changed on my server when upgrading to v6.11. Hence it remains hard to imagine where we users went wrong in this transition. So far I could maintain my WeKan instance without any programming language knowledge.

mfilser:
Wekan needs a WRITABLE_PATH for migration, this is already mentioned https://github.com/wekan/wekan/blob/master/CHANGELOG.md and in the issue you mentioned here.
You can remove the migration attachement entry from the Migrations collection to force a new migration at next Wekan start. Here's the migration done by Wekan:
https://github.com/wekan/wekan/blob/825bc4cf257c7a15d687f59b9e710f9cbe4652fe/server/migrations.js#L1129

BabyFnord:
Please forgive me asking for more details, I am not a programmer but try my best by trial & error. 

I am unable to find the specific file and set WRITABLE_PATH in a post-install bundle environment. Where do I find the specific file? Once this has been set, how do I
> remove the migration attachement entry from the Migrations collection
?

BabyFnord:
`start-wekan.sh` does not to exist neither in my post-install WeKan v6.11 nor v6.12. I see it in the base folder when downloading a fresh copy, but my running installation folder structure is different.

As `start-wekan.sh` can't be found, what about adding that path to `~/etc/services.d/wekan.ini` ?

BabyFnord:
Access by nosqlbooster now works, but in order to get anywhere, I would need to issue some commands. These answers require knowledge about MongoDB and database handling, which likely most end users do not have unless they are dev ops. No answer has been provided on where to set WRITABLE_PATH in a running instance of WeKan. I'm lost here, sorry.

Issue #4374, opened 2022-02-22
Import Done Button do not redirect


cristese:
## Issue
After a importing a board via JSON, the botton done ist shown.
by clicking on done button, nothing happends.

### Server Setup Information
Wekan-snap
WeKan Version: 6.05.0
Meteor-Version: 2.2
Node-Version: 12.22.10
Mongo-DB-Version: 3.2.22
MongoDB-storage-engine: wiredTiger
OS: Linux x64
OS Version: 4.15.0-143-generic

### Problem description
After a importing a board via JSON, the botton done ist shown.
by clicking on done button, apparently nothing happends. But if you go to the boards overwiev, the imported board ist shown many times. reproducing this problem showed, that the click of done create the imported board. because nothing happens by clicking the done button, user clicking much more times on the done button. per click on done, one importet board is created. so 6 times clicking on done = 6 boards

#### Reproduction Steps

Export Board to JSON. 
Create new Board -> Import
Import from Export (JSON Code).
Button Import.
Button Done (Problem).

Ben0it-T:
@xet7 
I think I got something
https://github.com/wekan/wekan/blob/0712309266b78847524e237d9e4bf46d4db87f6a/client/components/import/import.js#L80-L94

We need to subscribe to the new created board
```js
const subManager = new SubsManager();
subManager.subscribe('board', res, false);
```

However, `Utils.goBoardId(res);` doesn't work... strange.
If I replace by `FlowRouter.go('home');`, it seems to do the job.

![wekan-import](https://user-images.githubusercontent.com/37017213/156935801-f46df053-0e0f-4f64-97af-5741c5cc027c.gif)

Need to dig a little more

xet7:
@Ben0it-T 

Cool !! If you get it working, please send PR. Thanks!

Ben0it-T:
@xet7 
Interesting !
If I wait for the ready status of subscription, it seems to work like a charm.
```js
subManager.subscribe('board', res, false);
Session.set('fromBoard', null);
Tracker.autorun(function() {
if(subManager.ready()) {
  // all the subscriptions are ready to use.
  const board = Boards.findOne(res);
  if (board) {
    console.log("OK ! => " + board._id);
  }
  Utils.goBoardId(res);
}
});
```

<img width="1245" alt="wekan-import" src="https://user-images.githubusercontent.com/37017213/157112484-a81441bd-cc67-45ab-a543-2b6dd1739e2d.png">


I keep digging but I think we have something !

[...]

PR #4413, opened 2022-03-14
Fix Duplicate board and create board from template doesn't open board

Ben0it-T:
"Duplicate Board" and "Create board from template" now open board


Ben0it-T:
(https://github.com/wekan/wekan/issues/4374#issuecomment-1066135106)

Issue #4225, opened 2021-12-09
Bug: Long Subtasks cannot be deleted [with video]


Meeques:
Hi, I discovered a bug while trying to delete a long subtask. I used Firefox. If you need the exact version of wekan I can ask my colleague. 
I think the problem is that the whole subtask text is copied into the message that pops up asking you wether you are sure that you want to delete the subtask. In the Video I refreshed the page at a point where everything was frozen. Then I tried deleting the subtask in the minimized card display and it didn't work either. 
Kind regards!

Meeques:
I will delete the video now because of data sensibility. If you need it I can temporarily upload it again or sent it to you.

sleroq:
It's ugly, but I can delete the subtask on [public instance](https://boards.wekan.team/)
![2022-03-18_16-21](https://user-images.githubusercontent.com/52239427/159010452-65117974-fe0d-4515-b24d-33c71f5427a6.png)

xet7:
Fixed at https://github.com/wekan/wekan/pull/4446

PR #4446, opened 2022-04-03
UI improvements on maximized card header

helioguardabaxo:
Changed maximized card header to use width 100%. Added border to split left and right sides.


helioguardabaxo:
This PR resolves #4225

Issue #4463, opened 2022-04-06
Multi File / Attachment Storage


mfilser:
After moving von CFS to Meteor-Files it's now possible to implement a "multi file storage" at Wekan.
In the first step GridFS and Filesystem are implemented, later any other Storage System like Amazon S3 Bucket, Google Drive and so on are easy to implement.

**I am working on this**, see this preview:

![wekan-multi_file_storage](https://user-images.githubusercontent.com/13166201/162073707-71c1305d-e765-44e3-8269-6ed6a5f97849.gif)

There are some debug informations in this gif, they are removed in the final version.

Ben0it-T:
Thank you for this work
it will be great

BabyFnord:
Cool 👍🏻

mfilser:
In Admin panel there is a new page:
![grafik](https://user-images.githubusercontent.com/13166201/164312128-82e8e58d-bbdd-4a71-87dc-c03083973d97.png)

mfilser:
Attachment renaming is coming soon, initial idea of #2099

PR #4484, opened 2022-04-20
Multi file storage

mfilser:
Preview and "documentation" of this PR is: #4463

Attachments and files in general can be stored in this PR at GridFS and Filesystem. Future storages like Amazon S3, Google Drive and so on should be implemented in Wekan easier because all code logic is developed with the "strategy design pattern".

At Admin Panel there is a new page to move the attachments between the storages.


xet7:
Yippee, fileStoreStrategyFactory !! Javascript = Java !!

mfilser:
> Yippee, fileStoreStrategyFactory !! Javascript = Java !!

Indeed, this comes from my daily experience in Java development 😀 🕺

PR #4557, opened 2022-06-07
Move and copy card set sorting

mfilser:
Moving and copy cards to lists didn't set the sorting, therefore the card was anywhere in the new list.
Now, set the card sorting to the lowest number, therefore the card is always visible


xet7:
@mfilser 

Thanks!!

mfilser:
@xet7
Your welcome 🙂

PR #4598, opened 2022-07-18
Feature/save description on card exit

Viehlieb:
Wekan users reported repeatedly unintentional deletion of card descriptions when they were pressing outside the card details pane.
To prevent this, a dialogue will now show up and display the question whether changes should be saved to database or not (currently translations are for English and German)
There is an option in the settings to toggle the behavior via checkbox:
Screenshot from 2022-07-18 15-52-50
Screenshot from 2022-07-18 15-52-13


Viehlieb:
Hey @xet7, is there anything more needed to finish this pull request?
best - viehlieb

xet7:
@Viehlieb 

Please close this pull request, and make new pull request with below changes:

Please remove:
- All changes to users.js
- Meteor.call to users.js
- All changes to "Change Settings menu", removing all menu options
- German translation

Only have code for:
- By default, always rescuing unsaved description (with no options to turn off feature)
- Translation to English

xet7:
Actually, why there is popup and translation? Can you remove those also?

Just directly save from localstorage to card.

xet7:
Actually, I think this is mostly OK.

webenefits:
This is a great feature, by the way. It has already protected us from data loss several times. Thank you!

However, it would be very practical if the popup also appeared when you close a card with the “X” and not just when you click outside the card to close it. Or what do you mean @Viehlieb?

PR #4614, opened 2022-07-31
Update UI Interface

saurabharch:
Glassmorphism card and group design


xet7:
@saurabharch

What does this look like? Any screenshot?

WeKan does not use Stylus .styl files anymore. Stylus was removed at [WeKan v6.23](https://github.com/wekan/wekan/blob/master/CHANGELOG.md#v623-2022-05-17-wekan--release) and converted to CSS, that fixed many bugs. Your files are from some old version, that does not have newest CSS changes.

Current CSS code looks like this:
https://github.com/wekan/wekan/blob/master/client/components/boards/boardColors.css

Where are those fonts, icons etc from? What is their license?

xet7:
Sorry, it's not possible to merge Stylus .styl files.

Issue #4569, opened 2022-06-22
Feature Request: Android tablet, Enable drag handles automatically (was: At All Boards page, can not click existing board icon to open board)


xet7:
- Samsung Galaxy Tab A8 4G, 3 GB RAM
- Chrome
- Android 11

At All Boards page, can not click existing board icon to open board. This is when using desktop mode of WeKan on Android tablet. Desktop mode is visible when display is bigger that smartphone display.

xet7:
Solution was to enable drag handles. But, if possible, for Android etc those should be enabled automatically.

repmovs:
should be repaired after I've changed some calls of isMiniScreen to isTouchScreen

xet7:
Fixed at https://github.com/wekan/wekan/pull/4633

PR #4633, opened 2022-08-12
Set drag handle on touch screen not miniscreen

repmovs:
Changed calls to isMiniScreen to isTouchScreen where drag handles are initialized, should fix the issues with large screen tablets like in [ #4569](https://github.com/wekan/wekan/issues/4569)

### DIFF
--- a/client/components/boards/boardBody.js
+++ b/client/components/boards/boardBody.js
@@ -193,7 +193,7 @@ BlazeComponent.extendComponent({
     });
 
     this.autorun(() => {
-      if (Utils.isMiniScreenOrShowDesktopDragHandles()) {
+      if (Utils.isTouchScreenOrShowDesktopDragHandles()) {
         $swimlanesDom.sortable({
           handle: '.js-swimlane-header-handle',
         });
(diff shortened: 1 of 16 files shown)


mfilser:
@repmovs 
Thank you for your PR, for me, it work's great :slightly_smiling_face: 

I just had 1 "issue", i got a lot of merge conflicts while rebasing. I think it's an editor setting from you which changed the code to a new formating and style, or? Maybe you can disable it so the next time, me and anyone else get's less merge conflicts.

Thank you in advance,
BR

repmovs:
@mfilser will try to fix that in future pull requests, thanks for your observation

xet7:
Usually it's not needed to use any code formatters. Only change lines that need actual code changes.

Some years ago WeKan had some code formatters, like Prettier and ESLint, but because code formatters did disagree with each other sbout style, and caused too many changes and wasted time, they were removed.

Also, some years ago WeKan repo there were many branches, but because merging between branches did take huge amount of time, all was merged to master branch and extra branches deleted.

PR #4646, opened 2022-08-18
"Copy card" copies now attachments too

mfilser:


ChrisMagnuson:
@mfilser Copying cards that were complex (many checklists and check list items, 100+) has had a lot of problems since we started using Wekan as it was managed client side. We do this often and so we had to compensate for the flakiness by doing things like opening a separate browser window where we would do the copy and just leave that window alone while still working in Wekan in another window.

We would have to check the count of the tasks after the copy as if someone forgot to pay close attention you would end up with not all the tasks copied if for instance you navigated to another board before the copy operation finished.

If I am understanding some of this code correctly it looks like the whole copy card process is now server side and in our practical testing I have started a card copy, closed wekan, opened it again, navigated to other boards, etc. and while it may take 20+ seconds for the copy of the card to appear (less time than before) it now works reliably without being affected by what we do in the browser.

Thank you so much for making this change, this completely eliminates one of the pain points I had to make sure anyone using wekan was aware of and now it just works as expected, albeit still slowly for cards with lots of check list items, but it at least works reliably which is so nice! 🙂🙏

xet7:
@ChrisMagnuson

> Thank you so much for making this change, this completely eliminates one of the pain points I had to make sure anyone using wekan was aware of

What other pain points you have, that you need to make sure anyone using wekan is aware of ?

ChrisMagnuson:
One of the others that jumps to mind is telling people to be very careful when dragging and dropping cards, here is a animated gif showing the drag and drop behavior getting a bit mixed up when dealing with lists that have items that are lazy loaded combined with multiple swimlanes:

Recording 2022-08-22 at 11 29 10

xet7:
@ChrisMagnuson 

In what way it could be improved? Or would it be easier to move cards with keyboard keys, or clicking buttons?

ChrisMagnuson:
Another one that can cause heart burn is being very careful when selecting a board not to accidentally click the copy board icon:
image

Accidentally clicking this button on a board with several hundred unarchived cards plus a few thousand more archived cards with many checklists per card caused us a production outage for a few hours as the operation will run for a while, eventually fail, and restart over and over until the cache of the browser that accidentally clicked the copy button is cleared.

Closing that browser stops the cycle but as soon as that window opens up wekan again it starts all over again until the cached is cleared as I assume the retry logic for this operation was client side.

This was almost a year ago now so I am not sure if we would have quite the same level of pain but it would be very nice to have a pop up dialog or something else helping prevent someone from just trying to select a board and then having wekan go down because they accidentally clicked the invisible until hovered over button to copy the board.

Making those buttons visible all the time would also probably be a good idea, at least they would be able to see it and avoid it instead of quickly moving and clicking and never realizing they actually clicked a button that was appearing under their mouse position.

I have had users archive board unintentionally for the same reason that they didn't see that button and though they were just moving their mouse and then clicking the board and then it disappeared.

ChrisMagnuson:
@xet7 If you watch the animation you will see that the drop shadow of where the card is supposed to be placed eventually becomes disconnected from where the card is being dragged, you have to keep watching for a while to see it start to happen but it happens which then makes the card go to a place the user doesn't expect as it doesn't go where their mouse is.

mfilser:
@ChrisMagnuson
Thank you very much for your kind words, it's good to hear if the features are helpful and well done 🙂

Yes, the card copy is now done on server side because only the server can access the files at filesystem or mongodb.
It sounds like the "second feature" is a lot better than the main feature.

PR #4663, opened 2022-08-31
Remove hard-coded port 8080 binding in Dockerfile

willman42:
Removed hardcoding port 8080 in image, allowing more flexibility for complex docker deployments.

Forcing use of port 8080 invalidates various pieces of documentation claiming you can connect to Wekan on other ports like 3001 (see [here](https://github.com/wekan/wekan/wiki/Settings#webserver-config)and [here](https://github.com/wekan/wekan/wiki/Caddy-Webserver-Config#caddy-2)), but it also creates unnecessary rigidity in more complex Docker deployments.

For example in my environment (and the reason for this PR), I don't want to expose containers directly to the underlying host network stack, so I run Caddy as my single host-connected container. I want my Wekan container to connect directly to the Caddy bridge network I've created, but when I set `ports: 8081/tcp` in my wekan `docker-compose.yml`, `docker ps` still reveals that the Wekan container is also bound to the host net on port 8080 in addition to 8081 on the Caddy net.

Setting the environment variable `PORT=8081` in Wekan `docker-compose.yml` doesn't seem to work either. My only guess is because the container image itself was built to bind to port 8080 only.

The best solution I can think of is to simply remove the port declaration and binding in the Dockerfile (f26f86dd3f5eb47152d4cd3e27594780c13a3cbb), letting users set their own port declarations. To this effect I also changed the default `docker-compose.yml` to include a default `PORT=8080` environment variable as well as user guidance on the topic (ca35970f3f1e23d774b1ef15fb1686b58daff282).

This is only my 2nd PR to an oss repo so please let me know if I need to change anything or do anything differently to make everyone's lives easier!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -346,8 +346,6 @@ RUN \
     #rm -R /home/wekan/python
     #rm /home/wekan/install_meteor.sh
 
-ENV PORT=8080
-EXPOSE $PORT
 USER wekan
 
 #---------------------------------------------------------------------
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -171,6 +171,12 @@ services:
       #  - http://192.168.1.100    <=== using at local LAN
       - ROOT_URL=http://localhost  #   <=== using only at same laptop/desktop where Wekan is installed
       #---------------------------------------------------------------
+      # ==== DEFAULT PORT SETTING ====
+      # By default, Wekan will bind to TCP port 8080. If you need to change this to work with your
+      # reverse proxy settings or other Docker network topology requirements, change this value.
+      #  - Remember to change the above port binding if you change this value!
+      - PORT=8080
+      #---------------------------------------------------------------
       # ==== EMAIL SETTINGS ====
       # Email settings are only at MAIL_URL and MAIL_FROM.
       # Admin Panel has test button, but it's not used for settings.


xet7:
Hmm? I don't know. You just change exposed port from 80 to some other here:

https://github.com/wekan/wekan/blob/master/docker-compose.yml#L152

xet7:
Usually, when you have Caddy2 in ports 80 and 443, you proxy to local ports like for multiple wekan, like 3000, 3001 etc.

And then with ufw firewall only allow what's necessary:

```
sudo apt install ufw
sudu ufw allow http
sudo ufw allow https
sudo ufw allow ssh
```
And setup ssh public keys, and disable password login at `/etc/ssh/sshd_config`, and `sudo systemctl restart ssh`

xet7:
Oh, it seems I did remember to add Caddy2 examples to wiki:

https://github.com/wekan/wekan/wiki/Caddy-Webserver-Config

Currently WeKan Snap has only Caddy 1, that does not work well anymore. I should sometime add Caddy 2.

willman42:
> Hmm? I don't know. You just change exposed port from 80 to some other here:
> 
> https://github.com/wekan/wekan/blob/master/docker-compose.yml#L152

Well the whole point of my setup is that I don't expose any ports other than Caddy's 80 and 443. 

If I completely remove any port bindings in the Wekan service in `docker-compse.yml` the container still binds to `8080/tcp` on the `wekan-app_default` network.

I want to change this so that I can set which port wekan binds to instead of having to work around the hard-coded default of 8080.

xet7:
Hmm, well, I did not think about that before. Ok.

bronger:
Actually, EXPOSE is meant only for documentation and should not trigger any action when running the container.  Odd that a Docker tool like Compose itself violates that principle.

xet7:
@bronger 

I don't know. If there is some problem with Dockerfile or docker-compose.yml, someone can send PR.

mfilser:
@bronger
That's what i thought too, EXPOSE is only for documentation as i remember and for clearness it should be in the Dockerfile.

bronger:
FWIW, I need the `ENV PORT=8080` in the Dockerfile. Otherwise, my sidecar container (an nginx reverse proxy) in the Kubernetes pod cannot forward to localhost:8080 any more. Is the default port different from 8080?

willman42:
Ok it looks like I was mistaken about the EXPOSE command in Dockerfile. It does say it does not actually bind a port, but just to tell the person running the container what port is intended to be used. My bad.

However, when using the image made from Dockerfile in Wekan v6.41, I was unable to change which port the Wekan container bound to. Even if I specified a different port in `docker-compose.yml`, either with the `ports:` argument or by setting the PORTS environment variable. No matter what I put here, the container always bound to port 8080, in addition to what I specified. 

Can someone explain that behavior to me and how to properly override the default port binding for the Wekan container?

My apologies for the mistake in my PR 😞 I'll submit a PR reversing my changes.

bronger:
Don’t worry, I am an experienced Docker/Kubernetes guy and had to test a lot yesterday to understand the behaviour on _my_ side. However, I’ve always used Kubernetes instead of Compose. You tell in your docker-compose.yml in the ports section a value of `3025:8080`, correct?

willman42:
No, I do not want to "expose" any Wekan app port to the host network stack. I'm running caddy too which is the only container bound to the host net on 80/443. I want Wekan app to bind only to the defined "caddy-net" but on port 8081 since I already have another service bound to 8080.

I could get the wekan app service bound on 8081 but only _in addition to_ it binding itself to 8080 as well (which I never actually configured in compose). And it would not respond on 8081, though I could not find a specific error or log that told me exactly why.

That's why when I saw in Dockerfile it was "exposing" port 8080 I figured that was the culprit since whatever is configured in there cannot be changed by compose since the image has already been built by it. Hence the the "hard-coded" I used in the PR.

bronger:
Isn’t there a difference between publishing a port to the host and listening on that port on the host’s interface(s)?

willman42:
I don't believe so.

If docker-compose.yml has `ports: 3025:8082/tcp` the port bindings as shown in `docker ps` are: `8080/tcp, 0.0.0.0:3025->8082/tcp, :::3025->8082/tcp` which indicates wekan is listening on 3025 on the host directly.

Also, if docker-compose.yml has `ports: 3025:8080/tcp` then `docker ps` doesn't show the _additional_ binding of 8080: `0.0.0.0:3025->8080/tcp, :::3025->8080/tcp`.

When navigating to `http://<host>:3025` I could only get to wekan UI via the second method. The first returns connection refused.

xet7:
@bronger 

If port is published to localhost, like MongoDB to port 27017 or 27019, it is available only to Node.js etc on localhost, it is not exposed outside of server. Outside visible ports are defined with firewall like ufw https://github.com/wekan/wekan/pull/4663#issuecomment-1232319371

xet7:
But if port is open only on Docker internal network, it is not even opened to localhost etc.

willman42:
Yes exactly @xet7. For example, my pihole container running on the same host **only** has the `ports: 8080/tcp` declaration in docker-compose.yml for its web UI, which is connected only to the defined `caddy-net`. So traffic to `http://<docker-host-ip>:80` gets picked up by caddy and directed to the pihole container on port 8080. 

This is what I want to be able to do for my wekan container, and I can't seem to do this. My proposed change to Dockerfile was not the right solution, but I cannot figure out what the right solution is just yet.

bronger:
I see why you want to do that. My own topology is the same: An HAProxy listening on the host interface and forwarding traffic to Wekan, which has no connection to the host directly (not even to localhost). But again, I am not a docker-compose user.

The underlying docker runtime seems to use the EXPOSE declaration only for the `docker run -P` parameter. However, docker-compose could read the EXPOSE metadata from the image and do some magic with it. Maybe https://stackoverflow.com/a/40801773/188108 is helpful?

willman42:
Yeah I actually saw that post before. Expose (within docker compose) is a no-op and doesn't actually do anything that the `ports` argument can't do.

However, if the v6.42 Dockerfile (which removed the allegedly-useless EXPOSE env var) broke your Wekan container because it changed port bindings, then perhaps EXPOSE does actually affect the resultant container. Or at least it changes how compose operates.

@bronger when you used the v6.42 Dockerfile, can you elaborate more on how it broke your environment? It'll be interesting to see because you're using the container built by it in k8s and not compose like I am. Maybe there's a conclusion we can draw based on that?

bronger:
The EXPOSE did not affect me at all, whether it was there or not.

But the missing $PORT setting did. Apparently, PORT is used somewhere in the container, and the default port Wekan uses within the container is _not_ 8080. @xet7 Can you confirm that?

xet7:
@bronger 

Default port is 8080 inside container on Docker network. Outside port can be changed.

https://github.com/wekan/wekan/blob/master/docker-compose.yml#L152

willman42:
@xet7 Can this default inside port be changed somehow?

xet7:
@willman42 

Using Docker is optional. PORT is environment variable, to be set before running WeKan. You could use [Source Bundle](https://github.com/wekan/wekan/wiki/Raspberry-Pi) for your specific CPU, and many different [start-wekan.sh](https://github.com/wekan/wekan/blob/master/start-wekan.sh) or [start-wekan.bat](https://github.com/wekan/wekan/blob/master/start-wekan.sh) scripts to start many WeKan processes with different PORT, WRITABLE_PATH etc options, using the same wekan directory cloned with `git clone https://github.com/wekan/wekan`. Or make your own Dockerfile like that with multi wekan.

xet7:
@willman42

Or alternatively, you could use nested Docker container, like Docker in Docker. Exposing multi wekan sub-containers ports to container network where is your caddy etc.

willman42:
`start-wekan.sh` looks like it sets the default port to 2000. So where is port 8080 configured as default?

xet7:
@willman42 

`start-wekan.sh` is what configures it with bundle, when not using Docker.

On Docker, at Dockerfile https://github.com/wekan/wekan/blob/master/Dockerfile#L349

bronger:
> @bronger
> 
> Default port is 8080 inside container on Docker network. Outside port can be changed.
> 
> https://github.com/wekan/wekan/blob/master/docker-compose.yml#L152

This is the default expectation of docker-compose, not the default of the server inside the container in absence of the `$PORT` environment variable.

I checked by calling

    lsof -i -P -n | grep LISTEN
in the container the port that node/meteor uses. It is the same as `$PORT` of course, but if `$PORT` is not set, it is random (> 30000). This explains why I needed the `ENV PORT=8080` in the Dockerfile.

bronger:
Is this summary so far correct:

1. We need `ENV PORT=8080` in the Dockerfile, and it has no downsides.
2. `EXPOSE $PORT` has detrimental effect on docker-compose for which we still have no workaround.

mfilser:
@bronger
I found this documentation of environment variables:
https://docs.meteor.com/environment-variables.html#PORT

azielke:
> 2. `EXPOSE $PORT` has detrimental effect on docker-compose for which we still have no workaround.

AFAIK docker-compose doesn't do any magic with exposed ports. Even from reading through this, I don't really understand what problem compose and expose together cause.

My setup is with docker-compose and I'm using traefik in front of it. Traefik auto-detects services in docker but does need the EXPOSE to know, on what port the container is providing its service.

Removing the expose breaks this, and I need to manually add it to the docker-compose.yml (or add a label to the container for traefik to explicitly specify the port).

willman42:
> @bronger I found this documentation of environment variables: https://docs.meteor.com/environment-variables.html#PORT

@mfilser I'm not very familiar with Meteor, but is there a way we can make this variable configurable by the admin, perhaps via config file mounted with docker compose that the Wekan server reads upon every boot?

Because @bronger was right- the issue is not within Dockerfile EXPOSE or docker compose expose, it is within the Meteor app itself having a hard-coded binding to port 8080. And because of this, any and all docker configuration is required to use this port for Wekan, which is very limiting for any containerized deployment, Docker or otherwise.

bronger:
> Because @bronger was right- the issue is not within Dockerfile EXPOSE or docker compose expose, it is within the Meteor app itself having a hard-coded binding to port 8080. And because of this, any and all docker configuration is required to use this port for Wekan, which is very limiting for any containerized deployment, Docker or otherwise.

We can leave the `ENV PORT=8080` in the Dockerfile and you can override the value of `$PORT` in your docker-compose configuration.

azielke:
> And because of this, any and all docker configuration is required to use this port for Wekan, which is very limiting for any containerized deployment, Docker or otherwise.

This shouldn't be limiting at all. That the app uses port 8080 inside the container should not interfere with your deployment.

If you want to run wekan on any other port, you don't modifiy the port wekan is running on inside the container, but just publish it on any port as you want.

For example, to make wekan reachable on the host itself on port 8081:

```
# inside docker-compose.yml:
[...]
    ports:
      - "127.0.0.1:8081:8080"
[...]
# run docker compose up -d
# test:
root@docker:/opt/wekan# curl -s localhost:8081
<!DOCTYPE html>
<html>
[...]
```

If you omit the `127.0.0.1:`, docker will assume 0.0.0.0, making it reachable from "anywhere" that can reach the host (still on port 8081).

willman42:
> If you want to run wekan on any other port, you don't modifiy the port wekan is running on inside the container, but just publish it on any port as you want.
> 
> For example, to make wekan reachable on the host itself on port 8081:
> 
> ```
> # inside docker-compose.yml:
> [...]
>     ports:
>       - "127.0.0.1:8081:8080"
> [...]
> ```
> If you omit the `127.0.0.1:`, docker will assume 0.0.0.0, making it reachable from "anywhere" that can reach the host (still on port 8081).

See, I don't want Wekan reachable via the host network stack **at all, only** via the docker network my proxy container creates. This should be possible, as isolation from the host network stack is a prime reason for containerization at all. 

So if I set `ports: 8081` in docker compose or if I build an image with `EXPOSE=8081` in Dockerfile, this will tell the resultant container to bind to a port Wekan server itself is not listening on, so it is unreachable.

Hypothetically, how would one run 2 Wekan containers on the same docker network? Or what if I was running a different service entirely whose application was hard-coded to listen on port 8080, and had to be on my proxy container's network as well? There would be a port conflict of 8080 on the proxy network.

See [this post](https://forums.docker.com/t/how-to-use-multiple-containers-with-same-port-and-without-making-those-ports-available-to-the-host-machine/120969/8) from the Docker forums about this exact issue. The answer provided here is exactly what I'm proposing for Wekan:

> Check bar's documentation if you can configure the port used by the service to :4711 (so you don’t need to publish it to the outside world) - maybe with some environment-variables passed to the container or with a conf-file mounted into the container.

mfilser:
@willman42
What you are searching for is called "reverse proxy".

Here are 2 example projects:
- https://hub.docker.com/r/jwilder/nginx-proxy
- https://hub.docker.com/_/traefik

Please have look there and use Google to get more familiar with it. It's a configuration issue of Docker or something else and it's absolutely not a Wekan issue (as far as i read here).

willman42:
@mfilser I am using a reverse proxy- Caddy in my case. Caddy is running as a container beside Wekan and other services on the Docker host. I understand how the reverse proxy works. 

My issue is that the only port I can proxy to for Wekan is 8080, _**unless I expose Wekan to the host network**_ which I am not willing to do, nor should have to do given the virtual networking options containerization offers. Because the Wekan application itself, not the container created for it, **requires** it.

mfilser:
@willman42 
I dont know Caddy, but for now it sounds very limited to me. Maybe you should consider to change your reverse proxy from Caddy to something else.

willman42:
@mfilser i can have Caddy proxy from whatever URL to whatever URL & port I want, so it's not a limitation of my proxy.

How would you configure running 2 Wekan containers on the same host w/o exposing either of them to the host network stack? I.e. you cannot configure the containers to bind to 0.0.0.0 or even 127.0.0.1.

mfilser:
@willman42
I am using https://hub.docker.com/r/jwilder/nginx-proxy
and had running 4 Wekan instances without the need to change anything at Wekan.
As far as i know you don't have to map the ports, the reverse proxy uses the ports internally. My ports confuguration of docker-compose is empty, just configured my reverse proxy.

azielke:
> See, I don't want Wekan reachable via the host network stack **at all, only** via the docker network my proxy container creates. This should be possible, as isolation from the host network stack is a prime reason for containerization at all.

Then you don't need to configure anything. wekan is reachable on the docker you choose (either the default bridge or the bridge it creates...

Remember that containers get their own IP on startup, so you can have multiple wekan running on port 8080 as they run in their own network instance.

Like:

```
root@docker:/opt/wekan# docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' wekan-app
172.19.0.3
root@docker:/opt/wekan# curl -s 172.19.0.3:8080
<!DOCTYPE html>
<html>
```

Remeber as well, that different docker networks are isolated by default. You could run your caddy with host network, so it can reach all docker containers. BUT: You still need to figure out the IP of the wekan container every time and update your caddy configuration. caddy does not natively support listening for docker services.

that's why i use traefik - it just listens for docker events and automatically runs a reverse proxy when a container with certain labels is started.

mfilser:
Traefik is absolutely the best choice in my eyes, i am switching in the future from nginx-proxy to Traefik, it's already on my Todo list since some time

willman42:
So my hypothetical question didn't help convey my real question and just caused more confusion, my apologies. I know that multiple Wekan containers (even if on the same bridge net) have their own IP's so there'd be no port assignment conflict. That's my bad.

So here's my **real** question...

_**How can I configure Wekan to be reachable on a port other than 8080, without binding it to a port on the host network?**_

Obviously a containerized proxy needs to be used to access a container-only bridge network, but the **_choice_** of proxy is **_irrelevant_**. Basically I want this traffic path: external client -> docker host -> proxy container:80 -> wekan container:8081 -> wekan application. As has been demonstrated above, simply configuring the wekan container to listen on port 8081 does not allow me to access the underlying Wekan container which is by default listening on port 8080.

mfilser:
@willman42
It's still a Caddy issue for me as Wekan runs behind nginx-proxy and Traefik proxy already (the 2 links i sent above)
https://github.com/wekan/wekan/pull/4663#issuecomment-1235520542

mfilser:
@willman42 
Or in other words, i doubt there's a change needed in `Dockerfile`, and `docker-compose` is very specific to the whole environment and server setup.

xet7:
@willman42 

I would think, that wekan-app container is not exposed to outside of docker network, it container ports setting is like this, and you setup wekan node.js PORT with environment variable like this, when there is multiple wekan containers, that use different databases at same mongodb database server container.

In this case, wekan2 container has different ROOT_URL for those that use server [via VPN at office network](https://github.com/wekan/wekan/blob/master/docker-compose.yml#L723). In this case, these use different databases, so they have different logins etc.
```
wekan:
    ports:
      - 3001
   environment:
      - PORT=3001
      - MONGO_URL=mongodb://wekandb:27017/wekan
      - ROOT_URL=https://wekan.example.com
wekan2:
    ports:
      - 3002
   environment:
      - PORT=3002
      - MONGO_URL=mongodb://wekandb:27017/wekan2
      - ROOT_URL=http://10.10.10.10
```
Alternatively, you could use same database in MONGO_URL, so then HTTPS and VPN users have same logins, boards they are member of, etc:
```
wekan:
    ports:
      - 3001
   environment:
      - PORT=3001
      - MONGO_URL=mongodb://wekandb:27017/wekan
      - ROOT_URL=https://wekan.example.com
wekan2:
    ports:
      - 3002
   environment:
      - PORT=3002
      - MONGO_URL=mongodb://wekandb:27017/wekan
      - ROOT_URL=http://10.10.10.10
```
You could also have there other containers, like PostgreSQL, ToroDB etc, like this:

https://github.com/wekan/wekan/blob/master/torodb-postgresql/docker-compose.yml

https://github.com/wekan/wekan/tree/master/torodb-postgresql

azielke:
> How can I configure Wekan to be reachable on a port other than 8080, without binding it to a port on the host network?

You have to override _environment_ variable PORT, either with `docker run -e PORT=8081` or with docker-compose:

```
services:
  wekan:
    [...]
    environment:
      - PORT=8081
```

```
root@docker:/opt/wekan# docker compose up -d
root@docker:/opt/wekan# docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' wekan-app
172.19.0.3
root@docker:/opt/wekan# curl -s -v 172.19.0.3:8080
*   Trying 172.19.0.3:8080...
* TCP_NODELAY set
* connect to 172.19.0.3 port 8080 failed: Connection refused
##### -> not working, because port changed to 8081

root@docker:/opt/wekan# curl -s -v 172.19.0.3:8081
*   Trying 172.19.0.3:8081...
* TCP_NODELAY set
* Connected to 172.19.0.3 (172.19.0.3) port 8081 (#0)
> GET / HTTP/1.1
> Host: 172.19.0.3:8081
> User-Agent: curl/7.68.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: text/html; charset=utf-8
< Vary: Accept-Encoding
< Date: Fri, 02 Sep 2022 18:42:39 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< Transfer-Encoding: chunked
<
<!DOCTYPE html>
<html>
##### ->  working on port 8081
```

To get back to the EXPOSE - running `docker compose ps` still shows the container on port 8080:

```
root@docker:/opt/wekan# docker compose ps
NAME                COMMAND                  SERVICE             STATUS              PORTS
wekan-app           "bash -c 'ulimit -s …"   wekan               running             8080/tcp
```

... but as this is a documentation only feature, it does nothing in this setup. BUT autodiscovery between services (like treafik in my example) will break (because wekan on 8081).

This can be overridden with `expose` in docker-compose:

```
services:
  wekan:
    [...]
    environment:
      - PORT=8081
    expose:
      - 8081

root@docker:/opt/wekan# docker compose ps
NAME                COMMAND                  SERVICE             STATUS              PORTS
wekan-app           "bash -c 'ulimit -s …"   wekan               running             8081/tcp
```

willman42:
@azielke I could have sworn this didn't work for me, but maybe I am mistaken. I will retry this in my environment.

If this is the case, then this is the underlying cause of my entire line of questioning the last 2 days- thinking that `ports` arguments don't affect the underlying Wekan app port. 

Will let you know what I find, maybe we can finally close this thread 😊

xet7:
I did not test that above setting https://github.com/wekan/wekan/pull/4663#issuecomment-1235745794 but anyway with this it environment variable it should change port:
```
services:
  wekan:
    [...]
    environment:
      - PORT=8081
```

azielke:
> thinking that `ports` arguments don't affect the underlying Wekan app port

remember that the `ports` argument does indeed NOT affect the wekan app itself - but rather only tells docker what to do with traffic. If the ports you specify are not reachable, you just get a connection reset.

This discussion touched 3 different configs and aspects:
 * [1][2] _environment_ variable `PORT`: what port wekan will run on
 * [3] `EXPOSE`: what the admin/creator _documented_ what port the app will run on
 * [4] `ports:` in compose or `docker run -p`: what docker should do with traffic (bind on host, do port forwarding etc...)


[1] https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file
[2] https://docs.docker.com/compose/environment-variables/#set-environment-variables-in-containers
[3] https://docs.docker.com/engine/reference/builder/#expose
[4] https://docs.docker.com/config/containers/container-networking/#published-ports

willman42:
Ok it works. I must have missed something in my original testing that skewed the results. My sincere apologies for leading everyone down this wild goose chase...

For the record, here are my configs that allow Wekan to be accessible only via Caddy proxy and not host network:

**Caddyfile:**
```
kan.lab {
  tls internal
  reverse_proxy wekan-app:3001
}
```

**Wekan docker-compose.yml:**
```
services:
  wekan-app:
    environment:
      - "ROOT_URL=https://kan.lab"
      - "PORT=3001"
    networks:
      - caddy-net  # for web ui access via caddy
      - wekan-net  # for access to wekan-db service
    ports:
      - "3001/tcp"
networks:
  caddy-net:
    external: true
  wekan-net:
    name: wekan-net
```

Can confirm from `docker inspect wekan-app` that the `PORT` environment variable is set to 3001, so it is being properly passed to the Wekan app by docker compose.

We can (see what I did there?) consider this issue resolved. I may submit a PR for improving clarity in docs though.

PR #4700, opened 2022-09-21
Added a possibiliy of getting some wekan metrics datas

Emile840:
Route "/metrics" will be used to get those datas and "WEKAN_METRICS_ACCEPTED_IP_ADRESS" environment variable must be settled with a comma separated list of trusted ip address

### DIFF
--- a/models/server/metrics.js
+++ b/models/server/metrics.js
@@ -1,0 +1,145 @@
+import { Meteor } from 'meteor/meteor';
+import Users from '../users';
+
+function acceptedIpAdress(ipAdress) {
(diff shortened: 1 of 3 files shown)


mfilser:
@Emile840
Just for information: It seems you have an auto-formatter in your IDE which changed a lot of code, in my point of view it could result in merge conflicts, especially for every other developer who wants to merge their changes to the current master branch.
Here's the file with a lot of unnecessary code changes in this commit:
**models/users.js**

Emile840:
@mfilser 
You're right, i'll desactive it while working on wekan.

PR #4783, opened 2022-12-21
Add link card feature to rules

jos-webservices:
I have added a new feature to the board rules: linking of cards to another board

There is a new action that will allow you to create a linked card on baord as part of a rule. You can specify the target board, swimlane and list.

Remarks:
- if given list is empty or not existing, the rule will fail (similar behaviour to createCard rule)
- I could also provide code that would add the link to the first list of the target board

Let me know if I missed out something. This is my first pull request to a public project.

### DIFF
--- a/client/components/rules/actions/boardActions.jade
+++ b/client/components/rules/actions/boardActions.jade
@@ -77,6 +77,29 @@ template(name="boardActions")
     div.trigger-button.js-create-card-action.js-goto-rules
       i.fa.fa-plus
 
+  div.trigger-item
+    div.trigger-content
+      div.trigger-text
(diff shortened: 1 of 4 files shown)


xet7:
@jos-webservices

Thanks a lot !!!

> I could also provide code that would add the link to the first list of the target board

Please send PR. Thanks!

jos-webservices:
> @jos-webservices
> 
> Thanks a lot !!!
> 
> > I could also provide code that would add the link to the first list of the target board
> 
> Please send PR. Thanks!

Shall I fix the same for createCard rule as well?

xet7:
@jos-webservices 

Yes. Thanks!

PR #4928, opened 2023-05-16
Try to support old Meteor 3.0 syntax of _ensureIndex (not new createI…

SeveriTyo:
…ndex) for Sandstorm.

edited default list width

Thanks to xet7.

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -39,7 +39,7 @@ wekan-accounts-oidc
 
 # Utilities
 check@1.3.2
-jquery@1.11.10
+jquery@2.2.4!
 random@1.2.1
 reactive-dict@1.3.1
 session@1.2.1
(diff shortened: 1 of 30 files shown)


mfilser:
Current release is Meteor 2.12, so why supporting "old" Meteor 3.0 ? 3.0 is the upcoming release, not a old one. 

https://github.com/wekan/wekan/blob/master/.meteor/release

If i remember right the change from `_ensureIndex` to `createIndex` was necessary for Meteor 2.7. I guess this PR will break things again (not tested yet)

ocdtrekkie:
@mfilser Probably meant to say Mongo 3.0, not Meteor.

mfilser:
Hm, maybe.
Mongo 3.0 is so pretty old. And current Meteor needs Mongo 5.

xet7:
@mfilser @ocdtrekkie 

Please ignore, these were jus tests. Actually this was just about list width, but old commits were included also.

Issue #4558, opened 2022-06-07
Feature Request: Change List Width. And easier install on CentOS 8


NLBlackEagle:
## Suggestion
A way to adjust the wekan board css so I can widen the list width.

### Server Setup Information
CentOs 8 and practically all methods of installing among which build from source and using a pre-packaged bundle setup both failed.

### Problem description
I have been attempting to install a version of wekan which would allow me to adjust or append CSS files but have not found a solution as of yet. A separate issue would be that my dedicated server is not able to build from source given it has crashed as I attempted to do this (Most likely due to a lack of performance) I realize that setting up a workspace vm on my own computer would solve this issue but this goes against my reason to use wekan which is to improve my productivity.

NLBlackEagle:
Thanks! Is there a way I could customize the CSS without using a browser extension? I'm using WeKan as a roadmap with a public front and I can't exactly ask every visitor to install a browser extension or am I misunderstanding how it works?

NLBlackEagle:
I have update to the latest branch [v6.30] and see that you have added a settings controlling .list {} width
list._id.width(), 

But I take it these settings have not found their way onto a button yet or am I missing something?

xet7:
Currenly there is only changing any width, not saving yet. Yes, button is missing.

And someone would like to get fixed width back https://github.com/wekan/wekan/issues/4597

mark-i-m:
@xet7 As I understand #4597, they only want to freeze the layout on their board, not freeze it for everyone using Wekan.

If I can figure it out, would you welcome a PR that adds back the changes in https://github.com/wekan/wekan/commit/01d0dd3b1dd0a3c9764e7c9d31eab739db2d3ad8, but makes it configurable by the admin only?

xet7:
@mark-i-m 

This and #4904 should be per-user per-list and per-user per-swimlane profile setting. Simplest way would be to allow input integer number starting from from for example 100, and only allow input integer number (not text), so that list would not become invisible, and setting it in pixels.

There could be:
- List width setting at List ☰ menu, saved to users.js profile: boardId, listId, listWidth
- Swimlane height setting at Swimlane ☰ menu, saved to users.js profile: boardId, swimlaneId, swimlaneHeight

There is still about 780 issues open, all help is very welcome!

xet7:
Making something more fancy, like adjusting width and heigh with drag-drop or slider would maybe be more complex to implement, that just input integer number.

xet7:
So it is about:
[...]

PR #4964, opened 2023-06-13
[WIP] Make swimlane height and list width adjustable

mark-i-m:
cc #4558

@xet7 This is a minimum working version. I still need to fix a couple of things, as indicated as the end of this message... Please let me know what you think.

This PR makes list width and swimlane height a per-user per-list/swimlane adjustable value. I added forms to accept the value as an integer from the action menu.

Here are some screenshots:

image

image

image

The same for swimlanes.

I don't think I have the skill or time to make these drag-and-drop adjustments, though.

Some open questions:
- Currently, there are a bunch of hard-coded constants (e.g., min and default width/height). Where should these go?
- I think I am not doing the internationalization stuff right, as my test version still shows the placeholder text (e.g., `set-list-width-value`)... What am I missing here?


mfilser:
@xet7 
Maybe this PR is better for merging after V7. 00 (Performance Update).
Until then the PR can be tested and maybe extended

xet7:
@mark-i-m 

For Popup titles, you can add translation like `SomethingPopup-title`:
https://github.com/wekan/wekan/blob/master/imports/i18n/data/en.i18n.json#L94

PR #4989, opened 2023-06-30
Add support for viewing pdfs

### LINKED ISSUE #1318: Feature request: Attachment swipebox handling of non-images like viewing PDF etc
## Issue
**Impacted version**: 0.52

When viewing images in swipebox, if attachment is not image, spinner is shown, and nothing happens. Maybe non-images like .pdf of .zip could be excluded from swipebox? Or could .pdf files be shown somehow?

@brooksbecton 

Do you have ideas how to fix this?

VidVidex:
Fixes #1318 and maybe some related issues.

I added support for previewing pdfs in the browser (and also fixed some typos).

PDF viewer works by using the native HTML5 `<object>` tag, which can display pdfs without adding a new dependency.

image

This works fine in desktop Firefox and Chrome and in Firefox on Android.
It does **not** work in Chrome on Android.
To support that we'd probably need to use a 3rd party library.
The most popular one seems to be [PDF.js](https://mozilla.github.io/pdf.js/) but it uses Apache licence, which is [not ok](https://github.com/wekan/wekan/issues/1318#issuecomment-696023110).

### DIFF
--- a/client/components/cards/attachments.jade
+++ b/client/components/cards/attachments.jade
@@ -30,14 +30,14 @@ template(name="attachmentDeletePopup")
   p {{_ "attachment-delete-pop"}}
   button.js-confirm.negate.full(type="submit") {{_ 'delete'}}
 
-template(name="attachmentsGalery")
+template(name="attachmentsGallery")
   #viewer-overlay.hidden
     #viewer-container
       object#pdf-viewer(type="application/pdf")
         span.pdf-preview-error Your device does not support previewing pdfs. Try downloading instead.
     a#viewer-close.fa.fa-times-thin
 
-  .attachments-galery
+  .attachments-gallery
     if currentUser.isBoardMember
       unless currentUser.isCommentOnly
         unless currentUser.isWorker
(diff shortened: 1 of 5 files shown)


client/components/cards/attachments.jade
```diff
@@ -30,8 +30,14 @@ template(name="attachmentDeletePopup")
   p {{_ "attachment-delete-pop"}}
   button.js-confirm.negate.full(type="submit") {{_ 'delete'}}
 
-template(name="attachmentsGalery")
-  .attachments-galery
+template(name="attachmentsGallery")
+  #viewer-overlay.hidden
+    #viewer-container
+      object#pdf-viewer(type="application/pdf")
+        span.pdf-preview-error Your device does not support previewing pdfs. Try downloading instead.
```
VidVidex, line 37: I have not yet figured out how to add a new translation key for this. Any help is appreciated.

xet7: Added at https://github.com/wekan/wekan/commit/83ae1d4a19925363d8f176553fb10a98a77420be

xet7:
Hmm, it seems you figured out viewing attachment without swipebox. I'll try could I get bitmap images also viewing without swipebox, so that I could maybe remove swipebox, that is not compatible with newer jquery.

PR #5030, opened 2023-07-28
Yevhenii

ipyramiddev:
Added the ability to customize wekan's strings.


xet7:
Hmm? I don't understand for what purpose this is.

xet7:
@mfilser @helioguardabaxo 

Do you understand what is this for?

mfilser:
It's a feature to add custom language translation by the server admin.
But i would not merge it because changes in Dockerfile and docker-compose.yml arent't needed, and also not the file in custom_data, it's just an example which should be at a wiki.

xet7:
Translations for non-English should be done at https://explore.transifex.com/wekan/wekan/
so they are understanable in those languages.

PR #5036, opened 2023-07-31
Make available to add customized strings

ipyramiddev:


xet7:
Please no.

ipyramiddev:
What is the problem with this Pull Request?

ipyramiddev:
This is for users to be available to customize some strings when they deploy wekan on their own. Now this pull request doesn't any modifications of Docker file.

xet7:
@ipyramiddev 

Translations should be made at https://app.transifex.com/wekan/wekan/dashboard/ . Only 7 of 128 translations have everything translated.

ipyramiddev:
@xet7 
This function is intended to customize some of strings only.
Of course translations should be made at your url. But if a developer wants to change one or two sentences for only his desire, he should not touch i18n/data files, as they can be updated regularly.
In this case, the developer can create a file in i18n/custom_data folder, and make his own changes for a few strings. Then it will not be updated even though the developer updates from wekan's latest codes.

xet7:
@ipyramiddev 

Adding checking non-existing 128 custom translation files is a major slowdown.

xet7:
@ipyramiddev 

Also that means 128 error messages to browser console of every user of WeKan.

Developer can already change those strings in one translation file he is interested about. No need to slow down every WeKan user.

ipyramiddev:
@xet7 
There is no need to be exist for all files. The developer can make even only one file which he requires to modify only, and in the file he also insert only strings to modify, not all.
And when developer changed some strings in one translation file and the same file is updated by Wekan, it would be conflicts while merging.

ipyramiddev:
This function was requested from Alexander Bun, so I made it.

xet7:
@ipyramiddev 

Where it was requested? Is there some issue about it?

xet7:
@ipyramiddev 

If there is real need to change some some string in WeKan for some feature, for example changing:

- Add Swimlane

To different, examples:

- Add Car Lane
- Add Production Lane
- Add Team Lane

Then that translated string would be editable by BoardAdmin, and be saved to database to swimlane of that board.

WeKan database structure is at wekan/models/

Swimlanes etc board parts are at wekan/client/components/

xet7:
That way those alternative translations could be added to be included in import/export of board JSON.

ipyramiddev:
This is what I am requested.

"On the following page you can find a documentation about the outgoing webhooks of Wekan: https://github.com/wekan/wekan/wiki/Webhook-data.

As you can see, the data and the template are documented there. Our requirement now would be that these templates you see there can be customized. Of course, this should not be done via a core hack, but there should be a way within Wekan to override these JSON templates. For us, it would be fine if there was an option to, for example.

1. configure them in the snap settings (like here: https://github.com/wekan/wekan/wiki/Apache#7-snap-settings ) or
2. if there is a "child template" directory where you can put the templates and which of course is not overwritten by updates or
3. the nicest variant would be if you could store them via text fields in the backend admin area of wekan."

xet7:
@ipyramiddev 

For customizing webhooks, add environment variables to all WeKan platforms like this:

https://github.com/wekan/wekan/commit/ba05f383ca29211c5474e06c5ba6673e712afe7a

And then add those environment variables to webhooks:

https://github.com/wekan/wekan/blob/master/server/notifications/outgoing.js#L56

Then send PR.

ipyramiddev:
@xet7 
The webhook description is based on i18n encodings. So in case I am going to customize these descriptions, I should change language file in current source codes, right?

xet7:
@ipyramiddev 

I think you do not need to change any existing translations files.

You just replace this kind of code https://github.com/wekan/wekan/blob/master/server/notifications/outgoing.js#L138-L142 with something like this:
```
if (process.env.WEBHOOK_CUSTOM_USERNAME !== '') {
  // Change below to have your custom username
  const text = `${params.user} ${TAPi18n.__(
    description,
    quoteParams,
    user.getLanguage(),
    )}\n${params.url}`;
} else {
  const text = `${params.user} ${TAPi18n.__(
    description,
    quoteParams,
    user.getLanguage(),
  )}\n${params.url}`;
}
```
And after you have added that environment variable WEBHOOK_CUSTOM_USERNAME to all WeKan platforms like this:

https://github.com/wekan/wekan/commit/ba05f383ca29211c5474e06c5ba6673e712afe7a

Then it will be possible in future WeKan version be possible to use that setting like this:

```
sudo snap set wekan webhook-custom-username='Mr. Someone Else'
```

xet7:
@ipyramiddev

If you like to change something that is not translated yet, then you use that variable without `${TAPi18n.__` , making changes to string in original English language, using Javascript replace text.

xet7:
@ipyramiddev 

If you would like to make changes to webhook only if it some specific language, you could have environment variable for language, for example WEBHOOK_CUSTOM_LANGUAGE . Then at outgoing.js, you could check for it:
```
if (user.getLanguage() === process.env.WEBHOOK_CUSTOM_LANGUAGE) {
   // only change webhooks when there is this language
}
```

ipyramiddev:
@xet7 
Thank you for your kind description.
But what I was going to do was not for only me. I was going to make the user be able to customize their strings conveniently, without putting ENV variables and inserting codes in outgoing.js.

xet7:
@ipyramiddev 

Conveniently? Maybe in Admin Panel?

In that case:

## 1. Create new table to database for translations

- https://github.com/wekan/wekan/tree/master/models
- for example, table/collection `wekan/models/translation.js`
- There add fields like this:
```
    language': {
      /**
       * The uniq ID of the member
       */
      type: String,
    },
    'text': {
      /**
       * The uniq ID of the member
       */
      type: String,
    },
    'translationText': {
      /**
       * The uniq ID of the member
       */
      type: String,
    },
```
Then add editing those values code, similar like in other files at `wekan/models`.

## 2. Add Admin Panel left menu option

For example below webhook settings:

https://github.com/wekan/wekan/blob/master/client/components/settings/settingBody.jade#L37-L40

For example this translation setting menu:

```
            li
              a.js-setting-menu(data-id="translation-setting")
                i.fa.fa-globe
                | {{_ 'translation'}}
```

## 3. Add editable translation table

Add file `wekan/client/components/settings/translationBody.jade`

There add Form for editing transations similar like editable people table here:

- https://github.com/wekan/wekan/blob/master/client/components/settings/peopleBody.jade#L216-L294

For example:
```
template(name="translation")
  table
    tbody
      tr
        th {{_ 'language'}}
        th {{_ 'text'}}
        th {{_ 'translationText'}}
        th
          +newTranslationRow
      each translation in translationList
        +translationRow(translationId=translation._id)
```

Add CSS file `wekan/client/components/settings/translationBody.css` similar like this:

- https://github.com/wekan/wekan/blob/master/client/components/settings/peopleBody.css

Add Javascript file for editing translations like this:

- https://github.com/wekan/wekan/blob/master/client/components/settings/peopleBody.js

## 4. Publish translations

Add file `wekan/server/publications/translation.js`
```
Meteor.publish('translation', function() {
  const ret = Translation.find();
  return ret;
});
```

## 5. Read published translations and use them

Subscribe to translation texts similar like this:

https://github.com/wekan/wekan/blob/master/client/components/boards/boardBody.js#L39

Subscribe at your translations change code file, and if that language is selected and in use, change that text to translationText:

https://github.com/wekan/wekan/pull/5036/files#diff-280798bf3228358cf15345e23bad3394c03fcbc47eebff8ecca3652a15df8f90

ipyramiddev:
@xet7 I understand your intends, but I think it is limited on customizing webhook strings on a specific language only, not for two or more lanauges. Could you make this as an issue out so as to discuss more with Alexander, who requested me this specification?

xet7:
@ipyramiddev 

Moved to https://github.com/wekan/wekan/issues/5065

ipyramiddev:
Thanks.

PR #5049, opened 2023-08-02
Move authentication UI handling in correct place

### LINKED ISSUE #5048: No login mask after snap refresh to 7.02
### Server Setup Information
Installed wekan snap package tracking latest/candidate.

### Problem description
After a snap refresh from 7.01 to 7.02 the login mask is no longer shown and it is not possible to login to wekan. A snap revert to 7.01 fixes the problem.

image

Expected:

image

nebulade:
The connectionMethod component only deals with the authentication method select input.

Fixes #5048

### DIFF
--- a/client/components/main/layouts.js
+++ b/client/components/main/layouts.js
@@ -26,36 +26,64 @@ Template.userFormsLayout.onCreated(function () {
   templateInstance.isLoading = new ReactiveVar(false);
 
   if (!ReactiveCache.getCurrentUser()?.profile) {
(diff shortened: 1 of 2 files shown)


nebulade:
If anyone knows how to attach a listener for when the OAuth plugin injects the `.at-oauth` objects for https://github.com/wekan/wekan/pull/5049/files#diff-2d72427ec71cd7044ed889fc4cbf6f5dd5773bb2d438a632b09a17b968d3987dR74 then we should change that.

PR #5059, opened 2023-08-04
Make default swimlane auto-height

### LINKED ISSUE #5057: Bug: Fixed swimlane height
## Issue
With 26d2efdedb the swimlanes have a default height of 270px. The settings do not allow setting a nullish value to make it expand the full height, which was the behavior before. The only workaround right now seems to be to set this to a very high number. Am I missing something obvious to make it by default use the whole available height?

mark-i-m:
Fix #5057

I haven't tested this yet, but I think this could be a simple solution...


mark-i-m:
Setting a height of `-1` should restore the "auto" height behavior. It's not great UX, but it at least works.

Issue #4972, opened 2023-06-20
Feature Request: Open card near where clicked open card. Not always at left.


xet7:
From Brian King at Vanila Chat:

> The position of the card doesn't make it less editable.
> Its just user experience. When working on a card on the far right of the screen, I then have to mouse move to the far left of the screen to update the card. This involves a lot of left/right travel across the screen. I thought it felt 'nicer' when the card opened near where I had clicked to select the card.

![card-open-near](https://github.com/wekan/wekan/assets/15545/9669eff4-fd50-4200-a0dc-a4f4f9b4cf84)

techworked:
Same issue here. Previously it is not.

xet7:
That card is at left usually. It has maximize button to make card fullscreen.

Card is not currently freely moveable.

Previously there was bug, that when opening card, card opened off screen, it was not at visible area at all. It did take some fixes to have card always visible. That bug was in some cases, not always. So that bug was then fixed.

mfilser:
I changed it to open in the middle of the screen. Needs some testing, if everything is fine, PR comes after v7.00

chrisi51:
last working version was 6.81 what we are using right now ... im also checking this topic for a fix, which recovers the behavior for normal cases to show the opened card next to the minicard but also make maximized and scrolled view being displayed properly

xet7:
@chrisi51 

Maybe some of these?

- https://github.com/wekan/wekan/issues/4572
- https://github.com/wekan/wekan/issues/4822
- https://github.com/wekan/wekan/issues/5052
- https://github.com/wekan/wekan/issues/5064

You could search open and closed issues:
https://github.com/wekan/wekan/issues?q=is%3Aissue+is%3Aopen+card+open

All help very welcome! There are many bugs in new browser versions.

chrisi51:
> 

sorry @xet7 i tried for 2 hours to setup dev environment ... even put out my 2 year old from an snapshot ... i allways get dependency hell for some npm packets 
```
Please enter your choice: 1
Linux
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
wget is already the newest version (1.21.2-2ubuntu1).
curl is already the newest version (7.81.0-1ubuntu1.13).
git is already the newest version (1:2.34.1-1ubuntu1.9).
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
[...]

PR #5066, opened 2023-08-08
Fixing positioning of opened cards

chrisi51:
try to fix https://github.com/wekan/wekan/issues/4972 

- open card opens beside the minicard (as in earlier versions)
- re-enabled the autoscrolling feature to get the opened card in perfect view
- maximized opened card opens with position:fixed 
- new class "overlayed" added to board element when a card is opened to set `overflow:hidden` for the moment to prevent user from accidently scrolling in a point of death where you cant get back into a nice view
- scrolling on maximizing is commented out cause there is no need to scroll in any way

### DIFF
--- a/client/components/boards/boardBody.css
+++ b/client/components/boards/boardBody.css
@@ -19,6 +19,7 @@
 .board-wrapper .board-canvas.is-sibling-sidebar-open {
   margin-right: 248px;
 }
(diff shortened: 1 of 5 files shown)


chrisi51:
whoops ... had to use a new branch i guess ... the second commit tries to fix https://github.com/wekan/wekan/issues/5068

Issue #4227, opened 2021-12-09
Can't download attachments that have # in filename


CaptainPackers:
I am running a docker version of Wekan on a headless Unbuntu virtual machine, and it is working flawlessly for the most part. The only thing I can't get to work is the downloading of attachments. When I try I get a 403 error on my client machine.

I also have a snap instance installed on my client, and I have no trouble downloading attachments from this local wekan instance.

I don't think this is a bug. I'm just looking for community support.

Any suggestions would be appreciated.

Thanks

xet7:
@CaptainPackers 

Do you have Nginx in front of Docker Wekan?
https://github.com/wekan/wekan/wiki/Nginx-Webserver-Config
Attachment file size: https://github.com/wekan/wekan/issues/2103#issuecomment-455014303

Or maybe just backup docker, and restore it to snap?
https://github.com/wekan/wekan/wiki/Backup

CaptainPackers:
Hi Lauri,
Thanks for the quick reply.
It looks like switching to the snap version is going to work for me.
But I don't get how to export the docker database and import it to the new snap instance.
Any pointers there?

Regards,
Donald Lund

Sent with [ProtonMail](https://protonmail.com/) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

xet7:
@CaptainPackers 

https://github.com/wekan/wekan/wiki/Backup#backup-docker

https://github.com/wekan/wekan/wiki/Backup#restore-wekan-snap

CaptainPackers:
I did a docker dump and zipped up the dump directory.

Then I'm trying
wekan.database-restore FULL_PATTH_TO_DUMP_ZIP_FILE/dump.zip

All the files in the zip were chmod to 0777 before the zip
The zip file has 777 permissions, as does the directory that contains it
It looks like the zip file is being parsed, but the last line of output is
Failed: open /home/myhome/dump/dump.zip: permission denied

Any thoughts?

Regards,
Donald Lund

Sent with [ProtonMail](https://protonmail.com/) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

xet7:
@CaptainPackers 

Unzip that .zip file, use mongorestore.

xet7:
In directory where you are running mongorestore command, there is dump subdirectory that has raw database .wt files etc

CaptainPackers:
no reachable server

[...]

Issue #4144, opened 2021-11-13
Attachment names


gusben2020:
Hi,
this is about attachment
The name is not resolved in a good way by chrome when in the name there are spaces or similar chars.
trello execute a repalce method over thouse " chars " by "_" at the name sugested to download.

In firefox works well.

xet7:
Fixed at https://github.com/wekan/wekan/pull/5083 . Thanks to @VidVidex !

Issue #3776, opened 2021-04-30
v5.27 Filename with # can not be downloaded


xet7:
Moved to here from #2759 

From @gpalyu 

I have updated to most recent version, 5.27.0, and I am still having the same issue.

I went to an old card created 4 months ago with a file with # in it, and got error (403) still.

Then I created a new file with a # in it and uploaded and tried to download and got same error (403).

xavier-GitHub76:
Hello,
I have updated to most recent version, 5.27.0, and I am still having the same issue.

A new file with a `#` in it or  a new file with a `?` in it, and got error (403) still.

A new file with a `\` in it, and got error "Page not found".

A new file with a `%` in it, and got error "400 Bad Request"

A new file with a `"` in it or  a new file with a `*` in it, or  a new file with a `:` in it, or  a new file with a `<` in it, or  a new file with a `>` in it, or  a new file with a `|` in it, and file is renamed.

Good luck for fix

xet7:
Fixed at https://github.com/wekan/wekan/pull/5083 . Thanks to @VidVidex !

Issue #3774, opened 2021-04-30
v5.27 Attachment filenames not URL-decoded


EvgenGematogen:
## Issue
Hi! After updating to 5.27 bug with non-unicode attachments was fixed, but not completly. Now I can see filename in cards and preview, but when I download  them their filenames become broken. For example, file named "Тестовый файл.pdf" downloads as "%D0%A2%D0%B5%D1%81%D1%82%D0%BE%D0%B2%D1%8B%D0%B9+%D1%84%D0%B0%D0%B9%D0%BB.pdf".

So it appears to be URL encoded and not decoded back. In older versions filenames were preserved just fine, is there a chance for a fix?

**Server Setup Information**:

* Wekan version: 5.27
* Operating System: Debian 10
* Deployment Method: docker
* MongoDB Version: 4.4
* Browser: Chrome v90

**Problem description**:
Please, check screenshots

![image](https://user-images.githubusercontent.com/34194469/116660227-638ad500-a99b-11eb-9512-4eb73560c67c.png)

![image](https://user-images.githubusercontent.com/34194469/116660370-99c85480-a99b-11eb-8a40-77c7fef66a92.png)

neutrolino:
Confirm
I have the same issue.

exejay:
May I ask what happened to this issue?

xet7:
@exejay 

This is not fixed yet, because I'm working on 1000e+ features and fixes at [Roadmap](https://boards.wekan.team/b/D2SzJKZDS4Z48yeQH/wekan-open-source-kanban-board-with-mit-license) for customers that need them ASAP. Maybe after those I have time to look at this. If someone can help and send pull request, it's very welcome.

xet7:
Fixed at https://github.com/wekan/wekan/pull/5083 . Thanks to @VidVidex !

PR #5083, opened 2023-08-16
Fix downloading attachments with unusual filenames [WIP]

VidVidex:
There are quite a few issues that report problems with attachment downloads not working due to various symbols in the filename.
Issues: #3776, #4227,  #4144,  #3774

Here are my findings for the symbols mentioned in above issues:
- \#     preserved
- ?     replaced with a space
- \     replaced with _
- %     preserved
- "     replaced with a space
- \*     replaced with a space
- :     replaced with a space
- <     can't upload (XSS sanitization problem) --> renamed to sanitized version
- \>     replaced with a space
- |     replaced with a space
- \-     preserved
- _     preserved
- (space) preserved
- non-ascii (Тестовый файл.pdf) preserved

So far the only change I made is to rename files that might have XSS and print a warning so that a user (or an admin) can figure out why the filename was changed. 
Previously uploads for such files would fail silently

Issues mentioned above all complained that file downloads would fail with various errors but in my case, running wekan 7.06, they all work fine.
The only change I would make is to fix inconsistent replacements for \.

Do you think anything else should be done here?

### DIFF
--- a/client/components/cards/attachments.js
+++ b/client/components/cards/attachments.js
@@ -292,13 +292,16 @@ Template.cardAttachmentsPopup.events({
       let uploads = [];
       for (const file of files) {
         const fileId = new ObjectID().toString();
-        // If filename is not same as sanitized filename, has XSS, then cancel upload
-        if (file.name !== DOMPurify.sanitize(file.name)) {
-          return false;
+        const fileName = DOMPurify.sanitize(file.name);
+
+        if (fileName !== file.name) {
+          console.warn('Detected possible XSS in file: ', file.name + '. Renamed to: ', fileName + '.');
         }
+
         const config = {
           file: file,
           fileId: fileId,
+          fileName: fileName,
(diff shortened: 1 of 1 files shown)


xet7:
@VidVidex 

Maybe test that renamed file does not exist, so that it does not overwrite existing file? And then try with some new filename, with some added safe characters?

xet7:
@VidVidex 

Already existing file having same filename is not a problem, when it is stored to MongoDB GridFS. But it is a problem when stored to filesystem.

mfilser:
I think the problems in the issues you wrote were in the old file storage ostrio-files, with current Meteor-Files it's not a problem at all.

Same filename isn't possible because the filename starts with  the database object id

xet7:
Hmm, I think I'll add to filename beginning text about it being sanitized.

xet7:
That is info to admin, and added text in case after sanitization filename is empty.

PR #5088, opened 2023-08-17
Fix broken attachment preview in safari

### LINKED ISSUE #5042: Preview of attachments is broken in Safari
## Issue
The preview of attachments is broken in MacOS safari. When clicking on an attachment to make it bigger, only the size of the card is used and not the full browser screen. Also the image is cut off, so I can't see the full image in preview. 
When using Google Chrome, it works as expected.. 

### Server Setup Information

* Did you test in newest Wekan?: yes
* Did you configure root-url correctly so Wekan cards open correctly (see https://github.com/wekan/wekan/wiki/Settings)?
* Operating System: Debian 11
* Deployment Method (Snap/Docker/Sandstorm/bundle/source): Docker
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): Nginx
* Node.js Version: 14.21.3
* MongoDB Version: 6.08
* What webbrowser version are you using (Wekan should work on all modern browsers that support Javascript)? Safari (16.6)

### Problem description
Preview of screenshot is not shown in full browser width and cut off in the middle: 
Bildschirmfoto 2023-08-01 um 13 32 32

#### Reproduction Steps
Open image in preview in Wekan with Safari, MacOS.

VidVidex:
This PR fixes broken attachment preview in safari.
Closes #5042

### DIFF
--- a/client/components/cards/attachments.js
+++ b/client/components/cards/attachments.js
@@ -231,14 +231,20 @@ Template.attachmentViewer.events({
   'click #viewer-container'(event) {
 
     // Make sure the click was on #viewer-container and not on any of its children
-    if(event.target !== event.currentTarget) return;
+    if(event.target !== event.currentTarget) {
+      event.stopPropagation();
+      return;
+    }
 
     closeAttachmentViewer();
   },
   'click #viewer-content'(event) {
 
     // Make sure the click was on #viewer-content and not on any of its children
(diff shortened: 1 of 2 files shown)


saschafoerster:
Thank you for the fix. But I can't scroll trough images #5089, so the view is still kind of broken.

PR #5095, opened 2023-08-20
Attachments, big images are now fully displayed

mfilser:


saschafoerster:
@xet7 @mfilser It's better then before, but in Safari 16, images are not fully displayed. There is still part of the image at the bottom cut off. 
Safari, 100%: 
Bildschirmfoto 2023-08-21 um 06 48 36

When website is 200%, then you can see the full size image: 
Bildschirmfoto 2023-08-21 um 06 48 46

Issue #5086, opened 2023-08-17
Calendar view broken after updating to v7.XX


ricardogfp:
When creating new cards on the calendar view on version v7.00 and up, the card is created on the list view but it doesn't show on the calendar view. Similarly if a card with due date is created on the list view it is not shown on the calendar view.
Rolling back to v6.99 solved the issue

saschafoerster:
Our calendar is empty as well (Wekan 7.08, docker)

mfilser:
@xet7 
I will have a look at this

DimDz:
After creating a card, I have to hit refresh button in order for the card to appear on the calendar. Is it intended?

mfilser:
I don't think so. I haven't used the calender view yet.
I hope to find some time the next days to look at this again.

mfilser:
I looked at this and tested with Wekan v6.97 and it wasn't there working too. New cards created by calender view aren't visible right after creation with v6.97, there's a reload necessary too.

So it's not a problem of Wekan v7.xx.

mfilser:
@xet7 
If you can confirm this I think you can close this issue because it's not related to v7.xx. It's a new feature request to have cards visible right after creation by calender view.

xet7:
Fixed at https://github.com/wekan/wekan/pull/5105 . Thanks to DimDz !

saschafoerster:
@xet7 Would be lovely to push a new version with it. :) Thank you!

PR #5105, opened 2023-08-25
Fix for CalendarView, Create Card without Refresh

DimDz:
Fix for https://github.com/wekan/wekan/issues/5086
Changing `defaultView` from `'agendaDay'` to `'month'` fixes the problem that we had to refresh in order to show card when it's created with the calendar view.


saschafoerster:
I didn't know I wish for this change and I didn't guess that it was just changing one variable.

Issue #4388, opened 2022-03-03
Feature Request: Select triggering activities for outgoing webhooks


webenefits:
## Description
We have been using Wekan for a year now and are very happy with it so far. We want to stay with Wekan as project management and are therefore willing to donate for a few of our wishes now. :blush: 

This feature request is concerning **webhooks**. We currently have the problem that the generated comments in our Mattermost (via webhook) are just too many and so our channels are a bit spammed. In detail, we have for each project a Wekan board and a matching Mattermost channel in which we communicate in the team. Now both should act in symbiosis with each other, so that comments in Wekan or a new card in the respective board sends a matching post in Mattermost, but without flooding the channel too much with messages.

Our request would be an option similar to "webhooks-attributes" ([Supported settings keys »](https://github.com/wekan/wekan-snap/wiki/Supported-settings-keys)), where you can easily set for which activities in Wekan the webhook is triggered. Specifically, we just want notifications about:
- Newly created cards
- Moved cards
- New comments

Example configuration code:
```
snap set wekan webhooks-activities='cardcreated,cardmoved,commentcreated'
```

How complex would such a customization be?

### Server Setup Information
* Wekan Version: 6.09.0
* Operating System: Ubuntu 18.04.6 LTS
* Deployment Method: Snap
* Node.js Version: 12.22.10
* MongoDB Version: 3.2.22
* Webbrowser: Firefox and Chrome

webenefits:
Feel free to let me know if you want me to create the idea/feature request rather as "Discussions → Ideas" instead of an issue here.

xet7:
Hmm, maybe 500 or 1000 euro? Depends how many needs to be customized, some or all.

https://wekan.team/commercial-support/

webenefits:
Hi @xet7,
well you'll have to tell me how much to adjust, no? :grin: Or are there still unclear requirements or do you still have questions?

xet7:
Moved to https://github.com/wekan/wekan/issues/2026#issuecomment-1058436251

webenefits:
Hmm ok, since the request for "Feature Request: Push Notification, and notifications area with settings #2026" is already open for a very long time and it is also not clear when this will be implemented, we can unfortunately not donate anything for this. In addition, in my opinion, this is also a different topic, even if there are connections or intersections. Sorry :confused:

xet7:
@webenefits 

I can keep this issue open, if that is more clear for you.

[...]

PR #5106, opened 2023-08-25
Added a condition to filter unwanted webhooks

ipyramiddev:
This condition is needed for addressing issue #4388.


ipyramiddev:
@xet7 I added a condition so that if a user set a custom string of webhook description as "-" using the new function from issue# 5065 then that webhook doesn't work.

PR #5206, opened 2023-11-15
Add one click deploy to readme

jbgh:
Hey!

We built this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.  We saw that you have detailed instructions in the wiki but didn't see a way to contribute to it so we made a PR on the readme

After deployment, users receive a link like this, which can then be CNAME-ed to any domain:
https://e8ca23d7f2c96b20792d3c8de7e20fcc26924990.dome.tools

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,11 @@ that by providing one-click installation on various platforms.
 We also welcome sponsors for features and bugfixes.
 By working directly with WeKan ® you get the benefit of active maintenance and new features added by growing WeKan ® developer community.
 
+You can [self-host Wekan in one-click](https://app.trydome.io/signup?package=wekan) on Dome:
+
+[![Deploy to Dome](https://trydome.io/button.svg)](https://app.trydome.io/signup?package=wekan)
+
+
 ## Screenshot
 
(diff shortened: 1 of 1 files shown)


mfilser:
Maybe a Wiki page is best for it? Otherwise there could be every possible deployment at the Readme page and not only one?

jbgh:
Hey @mfilser I couldn't find a way to contribute to the wiki.  If you point me in the right direction I can add it there.  Thanks!

mfilser:
I dont know too. @xet7 maintains the Wiki

jbgh:
@xet7 I had originally wanted to add this line to the wiki under getting started but didn't know how to make that change.

`you can [self-host WeKan in one-click](https://app.trydome.io/signup?package=wekan) with [Dome](https://trydome.io).`

xet7:
I added info about Dome to https://wekan.github.io

jbgh:
Thanks for adding @xet7 !  Let me know if there's anything else we can do to help.

PR #5265, opened 2024-01-15
Added archive option to  of Wekan API

gustavengstrom:
This adds the a new boolean param option `archive` to edit_card of the Wekan API. 

Note: the API docs also need an update which is not included in this commit.

### DIFF
--- a/models/cards.js
+++ b/models/cards.js
@@ -3860,6 +3860,25 @@ JsonRoutes.add('GET', '/api/boards/:boardId/cards_count', function(
           newBoardId,
         );
       }
+      if (req.body.archive) {
+        function isTrue(data) {
+          return String(data).toLowerCase() === 'true';
+        }
+        var archive = isTrue(req.body.archive);
+        Cards.direct.update(
+          {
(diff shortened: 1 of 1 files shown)


xet7:
@gustavengstrom 

That existing API returns card count.

But your changes look like, that you change all archived cards to not archived, an not archived cards to archived. That does not make any sense.

gustavengstrom:
@xet7 I am not sure I understand.  My code changes are under @operation edit_card on line 3449. 

I have tested the edit_card API call using: 

`requests.put(f'http://localhost:4001/api/boards/{board}/lists/{list}/cards/{card}', headers = headers, data={"archive": True}).json()`

which works perfectly both for archiving and unarchiving a single card. The other cards of the board were not affected by this ! 

You seem to be referring to @operation get_list_cards_count which is in line 3409 above. I haven´t made any changes to this function.

xet7:
@gustavengstrom 

Sorry, I'll try to look and test it more carefully in near future. Thanks!

PR #5312, opened 2024-02-22
cards.js Added a control to check error if card is not updated

xator91:


xet7:
@xator91 

Could you explain, what does code in this PR do? 

Maybe, it could be more useful, if it would list what fields at card were updated?

xator91:
> @xator91
> 
> Could you explain, what does code in this PR do?
> 
> Maybe, it could be more useful, if it would list what fields at card were updated?

It would create a new boolean var with "updated" true or false, and if false throw an error. 

True mean = One of a field updated
False mean = One of a field not updated

PR #5346, opened 2024-03-14
Make the contents of the cards grow to the list width

kuba-orlik:
Before this change, increasing the width of the list does not give much of a benefit, because the contents of the list are wrapped to 230px, anyway: 

image

After this change, the full width of the card is taken advantage of:

image


e-gaulue:
Hi,

Just for my curiosity. How do you change the width of the list?

Regards,

xet7:
@e-gaulue 

> How do you change the width of the list?

At newest WeKan 7.x, click List menu: `☰` / `Set List Width`

PR #5377, opened 2024-04-14
Improve Docker files

C0rn3j:
"version" is obsolete, and using a heredoc for long RUN commands both looks and writes much better.

Additionally some compose comments are wrong - https://github.com/wekan/wekan/blob/3f28c0dd227fee70d3a1aae0513cb736178a634b/docker-compose.yml#L12 - but am unsure what should actually be there.


mfilser:
@C0rn3j
I would not suggest to remove version tag from docker-compose.yml, as far as I know it defines the used commands. Version 2 and 3 is much different, different users with different docker-compose version can get different errors

C0rn3j:
`docker-compose` has been EOL for almost a year now, it has been replaced by `docker compose` for a while.

Issue #5000, opened 2023-07-09
Upcoming performance update v7.00


mfilser:
There's coming a big PR with performance updates, especially to board and list loading.

I will add more informations here, especially for new "guidelines" for programming and data structure to make sure new features /  PRs will also use the performance code.

mfilser:
Here are some information regarding the performance commit for future development.

For database selection there's a new class, [ReactiveCache](https://github.com/wekan/wekan/pull/5014/files#diff-4751587058319b3b8229de441fb926843203cf5782373a4dcc9219a980ba5618R857) in [imports/reactiveCache.js](https://github.com/wekan/wekan/pull/5014/files#diff-4751587058319b3b8229de441fb926843203cf5782373a4dcc9219a980ba5618).
Every database selection **must** be done with this class, except this places:
- [server/migrations.js](https://github.com/wekan/wekan/blob/master/server/migrations.js)
- [packages](https://github.com/wekan/wekan/tree/master/packages)

If you need really need the database query and not an array set the parameter "getQuery" to true, e.g. at this commit: a5336056087987415c73576c13e0783d7ee46320.
This in only the case in some special situations, mostly with code working with the Meteor publication in [server/publications/](https://github.com/wekan/wekan/tree/master/server/publications)).

For combining database queries with an index search there's a new class [ReactiveMiniMongoIndex](https://github.com/wekan/wekan/pull/5014/files#diff-4751587058319b3b8229de441fb926843203cf5782373a4dcc9219a980ba5618R1267).
For examples how to use it, have a look at the source code, e.g. here: [ReactiveMiniMongoIndex.getChecklistItemsWithChecklistId](https://github.com/wekan/wekan/pull/5014/files#diff-f5de3c17bbc6bbbd447e1a046ae408076d5133d7f9a8f0db67ceb3a3dd5753eaR90)
Using it at the right places can make many queries a lot faster.

PR #5418, opened 2024-05-25
Devcontainer, reduce image size and other necessary changes for development

mfilser:


xet7:
@mfilser 

But but... Debian 11 is old, very vulnerable version of Linux. Previosly, I have compared Debian and Ubuntu, and at that time Debian had many vulnerabilities, and Ubuntu had none. WeKan uses Ubuntu 24.04 based container https://github.com/wekan/wekan/blob/main/Dockerfile#L1 . In Meteor WeKan, security is more important than container size.

If size really is a problem, I try to get more progress on WeKan Studio, where Node+MongoDB+WeKan is changed to one executeable that currently size is 9.5 MB. https://github.com/wekan/wekanstudio

xet7:
@mfilser 

When previously testing distos, Meteor WeKan had some problems when using Alpine Linux. WeKan has used Ubuntu for a long time, because it has worked bes.

xet7:
@mfilser 

If using Debian, I think current version is Debian 12.

xet7:
If anyone else is reading this, here is more details:

@mfilser has contibuted major features to WeKan, like major speedups at https://github.com/wekan/wekan/issues/5000 , and very many other features and fixes.

WeKan Studio https://github.com/wekan/wekan is currently my small prototype of future WeKan, where is not even login possible yet. It can not replace Meteor WeKan yet.

mfilser:
@xet7
Thank you for your hints.

I saw already in the main Dockerfile that the Ubuntu base image changed often, but not in the devcontainer directory Dockerfile, so I thougt you don't care if the version there.

For development Debian is good, i never had problems with it. I have more often problems with Ubuntu at development because the repositories are changing more often.

xet7:
@mfilser 

Well, you are right, actually I don't care, because I'm not using this devcontainer.

PR #5419, opened 2024-05-26
Dockerfile production and dev same build

mfilser:
Over the years both Dockerfile's changed many times. In the end they nearly have the same goal, but were build differently.
Now, both Dockerfile's are mostly the same build, except the places which needs to be different.


.devcontainer/Dockerfile
```diff
@@ -1,20 +1,24 @@
 FROM debian:11
-LABEL maintainer="sgr"
+LABEL maintainer="wekan"
+LABEL org.opencontainers.image.ref.name="ubuntu"
+LABEL org.opencontainers.image.version="24.04"
```
mueller-ma, line 4: These labels are wrong. The image is based on Debian 11, not Ubuntu 24.04.

PR #5455, opened 2024-06-27
Fix ModernDark Cards not showing correctly. 

walster001:
When ModernDark theme selected on a board, and more columns (lists) created than on screen, inciting a horizontal scrollbar. When scrolling horizontally and then trying to open cards, the cards would open off-screen due to the css parameters set as absolute positioning as opposed to fixed positioning.


walster001:
Apologies for my overuse of commits for this pull. I'm actually quite new to all this github stuff! 
Kind Regards, Matthew.

xet7:
@walster001 

It's OK to use many commits. Thanks!

PR #5480, opened 2024-08-03
changing card color now closes the popup and doesn't refresh the whole boards page

mfilser:


mfilser:
I think there was also an open issue, but not sure ...

PR #5485, opened 2024-08-06
don't set background image in .board-canvas too

mfilser:
the background image must only be set at 1 div, not at many div's

### DIFF
--- a/client/lib/utils.js
+++ b/client/lib/utils.js
@@ -4,7 +4,7 @@ Utils = {
   setBackgroundImage(url) {
     const currentBoard = Utils.getCurrentBoard();
     if (currentBoard.backgroundImageURL !== undefined) {
-      $(".board-wrapper,.board-wrapper .board-canvas").css({"background":"url(" + currentBoard.backgroundImageURL + ")","background-size":"cover"});
+      $(".board-wrapper").css({"background":"url(" + currentBoard.backgroundImageURL + ")","background-size":"cover"});
       $(".swimlane,.swimlane .list,.swimlane .list .list-body,.swimlane .list:first-child .list-body").css({"background-color":"transparent"});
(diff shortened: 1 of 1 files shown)


walster001:
It appears this commit has broken the ModernDark theme backgrounds. I'm not entirely sure why yet @mfilser @xet7 will look into it a bit more but not sure if you both would have any ideas? 

To replicate, create a board with a modern dark theme, and try to add a background URL image to it. It does not show :)

walster001:
Any themes created after "Dark" theme so from "Dark" to "CleanLight" appear to have this issue. Themes previous to Dark - "Belize" to "Midnight" appear normally with a background added.

xet7:
@walster001 

I will revert this change, and release new version of WeKan.

walster001:
I think I might have found the problem before you revert it @xet7 
BrokenWallpapersFix

I believe it's having the board class and having ui-sortable class on the background. Examples with class removed and without below. 

Class removed ui-sortable
BGFix1

Class present ui-sortable. 
BGFix2

If having background on ui-sortable is not essential (not exactly sure what that class controls) then I'm happy to edit all the other themes with it removed and push it as a pull request to main branch for you :)

xet7:
@walster001 

Please make those edits and send pull request. Thanks a lot !

walster001:
Actually upon further inspection it appears different themes set the background colours differently (likely due to different developers of those themes :O) so will have to play around with it a bit and see what I can do. I can see why it used to override board-canvas as moderndark theme sets the background colour on board-canvas. Will have to see how the dark theme sets the colour, set it to that, and remove it on ui-sortable.

walster001:
#5500 should fix broken themes. 

It mentions I have a conflict in the file? Just wondering if that's going to be an issue, trust it should work okay :)

Issue #5502, opened 2024-08-12
ModernDark List Titles not constrained properly. 


walster001:
## Issue

ModernDark List Titles (column titles) do not display correctly when viewed on an iPhone XR (assuming this is a problem on all mobile devices). When list titles are longer than a certain number of characters you get the following effect
![ModernDarklist](https://github.com/user-attachments/assets/08856364-f15c-462b-bcb9-85679286e142)

I believe it is related to this line-height parameter in the css. Was tempted to submit a pull req, but it shows up okay when I emulate it on my computer on chrome inspector mobile. Was wondering if @jghaanstra would have any insight into this as the creator/maintainer of ModernDark. When line-height parameter is removed it fixes the odd gaps in list titles shown on narrow screens >800px as the media suggests. 

Thinking to create pull for this change @xet7 , unset line height and add vertical-align:middle to list header property. From what I can see that will fix the alignment issue. 

![MDFix2](https://github.com/user-attachments/assets/1ac5fcf4-b2c2-4138-a944-218293a2bf22)

#### Reproduction Steps

Create a test board with relatively long titles on the lists. 
Switch theme to ModernDark (all other themes appear to be okay for alignment. Including modern)
View it on a mobile device.

xet7:
@walster001

Please send pull request. Thanks!

In general, anyone can send pull request anytime. No need to wait for me or anyone else to answer for something, because some people are sometimes busy doing other stuff. If pull request needs fixes, then close it, and send new one. Or if pull request is not needed, leave it closed.

walster001:
Cheers @xet7 I'm still a bit new to all this github stuff 😃

xet7:
Fixed at https://github.com/wekan/wekan/pull/5503 . Thanks to @walster001 !

PR #5503, opened 2024-08-12
Fix ModernDark Mobile View List Titles

walster001:
See issue #5502 

Modified List Header/Text values to unset default line height, and center text when viewed on mobile devices.


walster001:
@TheExo just wanted to check that you would be okay with drop shadow being removed on the Exo theme and replaced with a solid background the same colour as the drop shadow (when board image background not applied. 

Examples below. 

Without background applied (reworked list shadow)
ExodarkSolid

With background applied (no shadow or background colour applied to lists)
Exodark-Bgd

TheExo:
do it as you please :)

PR #5576, opened 2024-11-16
Change move-to-archive keyboard shortcut to '-', because of Ctrl + C shortcut

NadavTasher:
Ctrl + C from card text would cause the card to be archived :(

### DIFF
--- a/client/lib/keyboard.js
+++ b/client/lib/keyboard.js
@@ -159,7 +159,7 @@ Mousetrap.bind('space', evt => {
   }
 });
 
-Mousetrap.bind('c', evt => {
+Mousetrap.bind('-', evt => {
   const cardId = getSelectedCardId();
   if (!cardId) {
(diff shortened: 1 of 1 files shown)


mohammadZahedian:
why '-' ?
it is far in keyboard in one handed...
how about 'a' stands for a Archive

NadavTasher:
'a' is already used for 'My issues only'.
Also the issue was that in combination with 'Ctrl' (e.g. when you copy text), the shortcut would trigger, which is why 'z' is not a good candidate either.

mohammadZahedian:
@NadavTasher 

how about ' ` ' or ' q '
' ` ' means key below the ' esc '

NadavTasher:
Yeah that could work

NadavTasher:
Created https://github.com/wekan/wekan/pull/5589.

PR #5577, opened 2024-11-16
Potential bug(fix?) - Add assign-self keyboard shortcut (old shortcut actually adds)

NadavTasher:
The current `[SPACE]` shortcut toggles the current user as a member of the card, but the translation suggests it should assign.

This PR does not fix that - it retains the original behaviour, but updates the translation.
In addition, a new `[m]` shortcut is defined to actually toggle the current user as an assignee.

This is a potential bug, which could also be fixed by the following patch instead:
```
diff --git a/client/lib/keyboard.js b/client/lib/keyboard.js
index 79f90456c..6d315b984 100644
--- a/client/lib/keyboard.js
+++ b/client/lib/keyboard.js
@@ -152,7 +152,7 @@ Mousetrap.bind('space', evt => {
 
   if (ReactiveCache.getCurrentUser().isBoardMember()) {
     const card = ReactiveCache.getCard(cardId);
-    card.toggleMember(currentUserId);
+    card.toggleAssignee(currentUserId);
     // We should prevent scrolling in card when spacebar is clicked
     // This should do it according to Mousetrap docs, but it doesn't
     evt.preventDefault();

```

### DIFF
--- a/client/lib/keyboard.js
+++ b/client/lib/keyboard.js
@@ -139,6 +139,26 @@ Mousetrap.bind(numArray, (evt, key) => {
   }
 });
 
+Mousetrap.bind('m', evt => {
+  const cardId = getSelectedCardId();
+  if (!cardId) {
+    return;
+  }
+
+  const currentUserId = Meteor.userId();
+  if (currentUserId === null) {
+    return;
+  }
+
+  if (ReactiveCache.getCurrentUser().isBoardMember()) {
+    const card = ReactiveCache.getCard(cardId);
+    card.toggleAssignee(currentUserId);
+    // We should prevent scrolling in card when spacebar is clicked
+    // This should do it according to Mousetrap docs, but it doesn't
+    evt.preventDefault();
+  }
+});
+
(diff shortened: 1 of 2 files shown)


NadavTasher:
@xet7 

Is this a bug or a translation mismatch?
Either way, I think the new shortcut is appropriate.

xet7:
Shortcuts etc should be same as in Trello. Check at Trello how it works.

WeKan is original design, and some parts similar like best parts of some other kanban software.

https://github.com/wekan/wekan/wiki/Design-Principles

NadavTasher:
From their documentation, `SPACE` should assign (which now adds), while `M` should open the members window or whatever.
This means that the current behaviour is a bug as it does not assign.
Because many people already use WeKan, changing the behaviour of the `SPACE` shortcut would be annoying, which is why I suggest on keeping it the same while adding the new shortcut.

If you believe that the current behaviour should be changed, let me know and I will switch the two actions in my commited changes.

Issue #5554, opened 2024-10-24
Collapsed list locks all cards after uncollapse


datadelft:
## Collapsed list locks all cards after uncollapse

* Did you test in newest Wekan?:
Using docker instance (wekanteam/wekan:latest) updated 12/10/2024

* Did you configure root-url correctly so Wekan cards open correctly (see https://github.com/wekan/wekan/wiki/Settings)?

Root URL is set correctly trough ENV

* What webbrowser version are you using (Wekan should work on all modern browsers that support Javascript)?

Tested on latest Chrome and Firefox

### Problem description

After collapsing and uncollapsing a list, all cards in that list are locked for the user who collapsed that list. The cards can't be moved anymore

#### Reproduction Steps

Create a few lists in a swimlane. 
Put some cards in the lists. 
Collapse 1 of the list. Uncolapse the list. 
Try to move the cards

[youtube video showcasing the reproduction on second pc](https://www.youtube.com/watch?v=aqGjvJv8S3k)

#### Logs

[docker_Wekan_logs (last 1000).txt](https://github.com/user-attachments/files/17513883/docker_Wekan_logs.last.1000.txt)

xet7:
@datadelft 

What version of WeKan you have? Look at right top username / Admin Panel / Version.

xet7:
@datadelft 

Try to upgrade like this:

```
docker compose stop

docker rm wekan-app

docker compose up -d
```
Your data will stay there, these commands do not affect wekan-db MongoDB container.

datadelft:
Version  7.60.0

Updated to 7.61.0 but issue persist.

xet7:
When locking happends, to unlock, it's possible to reload webpage.

I'll try to find out why this locking happens.

xet7:
At Firefox, there seems to be some errors, like too much recursion.

```
Adding a listener for DOMSubtreeModified is deprecated and will be removed soon. Instead of a MutationEvent, use MutationObserver. https://developer.mozilla.org/docs/Web/API/MutationObserver 78fd2530170044ef6a5170acd77adaa2ace6a733.js:556:1557
Uncaught InternalError: too much recursion
    init https://boards.example.com/78fd2530170044ef6a5170acd77adaa2ace6a733.js?meteor_js_resource=true:446
    _ https://boards.example.com/78fd2530170044ef6a5170acd77adaa2ace6a733.js?meteor_js_resource=true:446
[...]

PR #5579, opened 2024-11-16
Opinion - Disable week-of-year in date displays

NadavTasher:
I think this feature does not improve the overall WeKan experience.

This number does not mean anything for the regular user.
It is difficult to understant what this number means without looking at the code, and it makes the UI too crowded.

I suggest removing this feature.

### DIFF
--- a/client/components/cards/cardCustomFields.jade
+++ b/client/components/cards/cardCustomFields.jade
@@ -79,13 +79,11 @@ template(name="cardCustomField-currency")
 
 template(name="cardCustomField-date")
     if canModifyCard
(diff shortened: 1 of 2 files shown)


xet7:
This feature is ISO week numbering, ISO standard 8601.

https://en.wikipedia.org/wiki/ISO_week_date

https://en.wikipedia.org/wiki/ISO_8601

NadavTasher:
Acknowledged.
Maybe this should be optional (maybe add a toggle in settings)
What do you think?

xet7:
@NadavTasher 

Too many toggles is worse.

For example, Discord settings has too many toggles.

NadavTasher:
Okay.

xet7:
It would be nice, if someone could help to get per-user swimlane and list collapse working, saved to users profile. #5554

PR #5581, opened 2024-11-17
Make lists fill up space evenly, change listWidth to max-width

NadavTasher:
Before:
image

After:
image

With large values in list width:
image

This looks nicer and more even when large widths are set on all lists, but retains current behaviour.

### DIFF
--- a/client/components/lists/list.css
+++ b/client/components/lists/list.css
@@ -7,12 +7,14 @@
   border-left: 1px solid #ccc;
   padding: 0;
   float: left;
+  flex: 1;
 }
 [id^="swimlane-"] .list:first-child {
   min-width: 20px;
 }
 .list:first-child {
   border-left: none;
+  flex: none;
 }
 .card-details + .list {
   border-left: none;
@@ -30,6 +32,9 @@
   border-color: transparent;
   box-shadow: none;
   height: 100px;
+}
+.list.list-collapsed {
+  flex: none;
 }
 .list.list-composer .open-list-composer,
 .list .list-composer .open-list-composer {
--- a/client/components/lists/list.jade
+++ b/client/components/lists/list.jade
@@ -1,6 +1,6 @@
 template(name='list')
   .list.js-list(id="js-list-{{_id}}"
-                style="{{#unless collapsed}}width:{{listWidth}}px;{{/unless}}"
+                style="{{#unless collapsed}}max-width:{{listWidth}}px;{{/unless}}"
                 class="{{#if collapsed}}list-collapsed{{/if}}")
     +listHeader
     +listBody
--- a/client/components/swimlanes/swimlanes.css
+++ b/client/components/swimlanes/swimlanes.css
@@ -1,8 +1,3 @@
-@media screen and (min-width: 801px) {
-  .swimlane.ui-sortable {
-    width: max-content;
-  }
-}
 [class=swimlane] {
   position: sticky;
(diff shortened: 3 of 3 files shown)


mfilser:
@NadavTasher 
commit 465e2b1b6 breaks some things at my side:
- list columns are very small
- Drag Scroll (PR #5477) doesn't work anymore

NadavTasher:
Could you share screenshots?

mfilser:
Screenshot:
grafik


I am using FireFox 132.0

NadavTasher:
I see. Can you try assigning a larger list width for each list? (manually)

mfilser:
I have dozen of boards with many columns.
There are users out their which have hundret of boards and thousand of columns. I doubt that's a good solution to adjust always the column width manually.

NadavTasher:
I'm sure there are edge cases where this feature is more of a hassel then a blessing - what do you suggest we do? (Instead of reverting, as I do think this feature is good)

When you adjusted the column width manually, did it work as expected?

mfilser:
For me not. This PR breaks things which were working before.
Better is a solution which breaks nothing and just add a new feature 🙂

mfilser:
This sounds good.
Please consider both issues (DragScroll too)
Thank you 🙂

NadavTasher:
Of course :)
I'll keep you posted.

NadavTasher:
@mfilser

I have opened https://github.com/wekan/wekan/pull/5584 to fix both regressions while retaining the new feature.

mfilser:
@NadavTasher 
Thank you for the fix, it works well for me :slightly_smiling_face:

NadavTasher:
@mfilser 
Which version did you test?

mfilser:
Commit 12d22f9

NadavTasher:
Great!
Unfortunately the entire feature was reverted in https://github.com/wekan/wekan/commit/096fe130f68e0d8d082d309901c75ed04285b7e2 :(

mfilser:
I still had scrollbar's. Were only some of them hidden. What reason do you have to hide scrollbar's?

NadavTasher:
Just personal design opinion.
I think that scrollbars in the middle of the page are kind of ugly, but I agree that it is debatable.

xet7:
It's very hard to scroll, when scrollbar is not visible. Also related to accessibility.

NadavTasher:
Noted.

xet7:
Bigger scrollbars are better, for those that do not see well.

walster001:
> Bigger scrollbars are better, for those that do not see well.

@xet7 Just curious, for the ModernDark theme I had utilised thin scrollbars (in keeping with the original theme design of thin scrollbars on the list columns) for the overall window scrollbars. I believe this might've been reverted/broken in a recent change?

Question, would you be okay with reverting and keeping the thin scrollbars in the ModernDark theme as I believe there is an override to scrollbar width at a top level rather than a theme level at the moment as my thin scrollbar entries for ModernDark are greyed out (overridden) by something :) 

All good either way just wanted to check.

xet7:
@walster001 

If it now has big scrollbars, that is how it needs to be. I have hard time trying to click thin scrollbars.

I will also later add more accessibility fixes, like Declaration of Accessibility, changes to colors of themes for accessibility, and fixed to keyboard navigation for accessibility. Those changes are required by law, and paid by public sector.

PR #5583, opened 2024-11-19
Change way of disabling scrollbars, disable swimlane scrollbars

NadavTasher:
The old way of disabling scrollbars does not work in newer versions of Chrome.

### DIFF
--- a/client/components/cards/cardDetails.css
+++ b/client/components/cards/cardDetails.css
@@ -89,10 +89,7 @@
   box-shadow: 0 0 7px 0 #b3b3b3;
   transition: flex-basis 0.1s;
   box-sizing: border-box;
-}
-.card-details::-webkit-scrollbar {
-  /* Hide scroll bars from middle of the screen */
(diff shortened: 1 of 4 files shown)


listenerri:
Thanks for the large number of pull requests you've contributed recently, but the latest tag version contains the problem of missing the right list, which occurs when there are a large number of lists, such as 9 lists

When there are too many lists, the rightmost list is squeezed out of the window and no horizontal scroll bar is available

Do you have any plans to deal with this problem?

xet7:
Related #5585

NadavTasher:
You are correct that there is no scroll bar.
This was intentional, but might not be the best idea.
Dragscroll still works, as well as Shift + Scroll, but commit https://github.com/wekan/wekan/pull/5583/commits/6c32c210f85826c18a006433392f0e34055e96bd may be reverted which will bring back the scroll bars.

@listenerri @xet7 
What do you think?

PR #5584, opened 2024-11-19
Improve list auto-width, fix regressions

NadavTasher:
The previous PR (https://github.com/wekan/wekan/pull/5581) introduced two regressions:
1. Lists were shrunk down to their minimal width when no space was left
2. Dragscroll stopped working inside swimlanes

This PR fixes both of these regressions, and improves on dragscroll support IMO:
1. Instead of declaring the maximum size a list can be, we declare the minimum size. This defaults to `270px`, and by that we ensure that lists are only narrow (as configured) when there is no space on screen. When there is enough space on screen (for example, when there are only 4 ~ 5 lists, they will fill the available space.
2. Dragscroll support inside swimlanes is now supported, but now scrolls without the swimlane header (title), which looks better, and allows to scroll each swimlane individually. This is an improvement in my opinion.

Overflow - Before regression (state in v7.63):
image

Overflow - Before (state in https://github.com/wekan/wekan/pull/5581):
image

Overflow - After (state here):
image

You can see that the layout before the regression and after the new fixes looks essentially the same (give or take the scroll bars which are disabled in https://github.com/wekan/wekan/pull/5583)

Underflow - Before regression (state in v7.63):
image

Underflow - After:
image

You can see what the feature is used for - filling up available space when available.

This is an example of how this could be used when adjusting the minimal width of lists:
image

Looks nice IMO :)

These regressions were brought to my attention by @mfilser. Thanks!


NadavTasher:
@xet7 
Notice translations might have to be updated.

xet7:
@NadavTasher 

Yes, I have already made some fixes to translations, pushed translations to Transifex, announced new changes to translators, and translated to Finnish. I have not made new WeKan release yet.

https://app.transifex.com/wekan/wekan

NadavTasher:
Sounds great!

mfilser:
@NadavTasher 
Well done :partying_face:

NadavTasher:
These changes was unfortunately reverted completely by @xet7

Issue #5347, opened 2024-03-14
Large Board, with "Uncaught RangeError: Maximum call stack size exceeded." or other error


Dexus:
### Server Setup Information

* Did you test in newest Wekan?: yes v7.38
* Did you configure root-url correctly so Wekan cards open correctly (see https://github.com/wekan/wekan/wiki/Settings)? yes
* Operating System: windows 11
* Deployment Method (Snap/Docker/Sandstorm/bundle/source): Docker
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first): haproxy
* Node.js Version: 14.21.4
* MongoDB Version:  6.0.9-7
* What webbrowser version are you using (Wekan should work on all modern browsers that support Javascript)? edge

#### Reproduction Steps
create a large board, and try to open it. 

**cached browser request**
Uncaught RangeError: Maximum call stack size exceeded.... 

**if you disable cache your get on the console:**
```
Exception from Tracker recompute function:
5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:1 TypeError: Cannot read properties of undefined (reading '_id')
    at e.originRelativeUrl (5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:831:1047117)
    at 5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:549:34947
    at f.call (5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:609:2119)
    at f.mustacheImpl (5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:609:1305)
    at f.mustache (5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:609:1364)
    at href (5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:831:425510)
    at d._withCurrentView (5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:549:23492)
    at t.visitAttribute (5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:549:22617)
    at t.visitAttributes (5c2e2ca6a015f74da29f01295b5e8092e84051c4.js?meteor_js_resource=true:547:9410)
[...]

Issue #5140, opened 2023-09-22
become slow after a week 


mohammadZahedian:
after using wekan about a week it becomes slow 
i have to restart snap wekan, then it is fast and amazing again
now i am thinking about a cron job to restart wekan once a day😂😎

ubuntu / 6GB ram / 6cpu core / 100G ssd
 
![image](https://github.com/wekan/wekan/assets/25129673/b5f206a9-cfdf-4433-95c2-6413615ec1c9)

mohammadZahedian:
and sometime it wont load until restart wekan
@xet7

xet7:
@mohammadZahedian 

Yes, that happens on my WeKan server too. I'm thinking how to fix this.

xet7:
@mohammadZahedian 

I do not know does it make any sense of trying to fix this yet.

Because with WeKan v7.00 ReactiveCache changes, it did speed up WeKan, and make that "becoming slow" to be more rare, only happening once a week. Some years ago, I had similar problem with Golang based server, it did become unresponsive after one week. It would be easier to debug, if memory leaks, or memory usage growing, would happen more frequently, like after one day, so that it would be possible to see changes in profiler. But one week it much harder to debug.

Also, soon there will be upgrade to Meteor 3.0. That will add many additional fixes and speedups. It could be, that this "becoming slow" bug will just disappear sometime during upgrades. This kind of bug could be anywhere, like in Node.js itself, Meteor code, or WeKan code.

It would also be an option to make WeKan restart automatically daily, with some internal cron. I have not tried to implement it yet.

xet7:
That bug could also be dependencies WeKan has. WeKan dependencies are updated very often. So if someone fixes some bug in some dependency, upgrades also bring that fix to WeKan.

Upgrading to Meteor 3.0 also upgrades to Node.js 18, that has many fixes that came after current Node.js 14. After that, Meteor will upgrade to Node.js 20 etc newest Node.js and other dependencies, being the leading realtime web framework.

mohammadZahedian:
@xet7 
what should i log when it stuck in loading?

xet7:
@mohammadZahedian 

You should have nightly cron that runs these commands in bash script

```
sudo snap stop wekan
sudo snap start wekan
```

mohammadZahedian:
> @xet7 what should i log when it stuck in loading?

i mean,  how should i undrestand  why i stucked?
i there any log file?
which logs should i check?

xet7:
[...]

Issue #3585, opened 2021-02-18
Wekan Crash if more than 2GB of RAM is used


r4nc0r:
## Issue

Other Wekan issues can be added here.

**Server Setup Information**:

* Did you test in newest Wekan?: yes 
* For new Wekan install, did you configure root-url correctly so Wekan cards open correctly https://github.com/wekan/wekan/wiki/Settings ?
* Wekan version: 4.96
* Deployment Method: docker swarm
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first):
* Node Version:
* MongoDB Version: 4.0.11

**Problem description**:
if NodeJs uses more than 2GB of Memory, Wekan crashes with a Exception:

I researched a little and as it turns out NodeJs v12 has a standard limit of 2GB Memory Consumption, but it can be changed with the following Nodejs CLI Command: ` --max-old-space-size xxxx` where xxxx is the memory in MB [(source)](https://github.com/nodejs/node/pull/25576) Can this be implemented to raise the Wekan max Memory Limit?

> FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory,
,
<--- Last few GCs --->,
,
[1:0x38a2e80]   324979 ms: Mark-sweep 2029.2 (2080.6) -> 2018.8 (2083.1) MB, 1608.6 / 0.1 ms  (average mu = 0.120, current mu = 0.022) allocation failure scavenge might not succeed,
[1:0x38a2e80]   327768 ms: Mark-sweep 2031.4 (2083.1) -> 2022.9 (2063.4) MB, 2738.9 / 0.1 ms  (+ 0.1 ms in 30 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 2751 ms) (average mu = 0.061, current mu = 0.018) allocation f,
,
<--- JS stacktrace --->,
,
==== JS stack trace =========================================,
,
    0: ExitFrame [pc: 0x140a859],
Security context: 0x22ae44d408d1 <JSObject>,
[...]

PR #5588, opened 2024-11-23
Created new x86_64 Dockerfile, changed source of two packages in packages.json

NadavTasher:
This PR requires a thorough review, as it changes the build process of WeKan docker containers targeting x86_64.
If desired, this new Dockerfile can be adapted to build multi-platform images using buildx features.

Note - even after this PR, the Dockerfile can still be improved!
More caching can be acheived by building the client and server bundles seperately, a more minimalistic base image can be used, Node dependencies can be reduced (more of a package.json makeover).

I might be targeting these inefficiencies in the future.

## Prerequisites
This new `Dockerfile` uses `buildkit`, which is the default build backend since Docker 23.0 (2023-02-01).
This means that building (not running) the docker image requires using Docker 23.0 or specifically specifing a buildkit build using `docker buildx`.

## Updates done to Dockerfile
1. Use a long-term-support version of Ubuntu (24.04) which will be supported for much longer then 24.10 (means that if you build this image in 10 months from now (just an example), the `ubuntu:24.04` image will also be updated with latest security patches)
2. Install NodeJS and Meteor from official archives
3. Install Meteor from archive with specific version instead of using remote script (which did not take version into account and is dangerous, now actually using meteor 2.16)
4. Use multi-stage build to reduce final image size and unnecessairy bloat (for instance, the original `Dockerfile` created a sources layer using `COPY`, which was included in the final image. It could not have been removed due to how docker layers work)
5. Copy only required files in build to allow docker to cache some of the build state (improves build speeds)
6. Explicitly set UID to `999` (same behaviour as current image), to make sure nothing breaks in the future
7. Allow `STACK_SIZE` to be adjusted through environment variables
8. Use multiple `RUN` commands which improves caching ability in builds and also makes builds more predictable, as old `Dockerfile` did not fail when any build command (in the build part) failed.
9. `package.json` - Changed source URL of GitHub dependencies to HTTPS as the git protocol is not available since 2022 but the `npm` version used still tries that, which takes some time (https://blog.readthedocs.com/github-git-protocol-deprecation/)

#### Full uncached builds
Old Dockerfile:
```
$ time docker build . -t wekan-test-original --no-cache

real    16m43.746s
user    0m4.895s
sys     0m2.935s
```

New Dockerfile:
```bash
$ time docker build . -t wekan-test-new --no-cache

real    7m25.917s
user    0m1.359s
sys     0m0.906s
```

Overall, a ~x2 improvment!

### Partially cached builds (only a small code change was made)
Old Dockerfile:
```
$ time docker build . -t wekan-test-partial-original

real    16m39.338s
user    0m4.767s
sys     0m2.890s
```

New Dockerfile:
```
$ time docker build . -t wekan-test-partial-new

real    5m41.032s
user    0m1.003s
sys     0m11.877s
```

Overall an average improvement of ~3x!

### Final image size
Current latest image (`wekanteam/wekan:v7.67`):
```bash
$ docker save wekanteam/wekan:v7.67 | wc -c

1019538432
```

New image built with new Dockerfile:
```
$ docker save wekan-test-new | wc -c

537956864
```

Overall a ~x2 improvement (from 1.02Gb down to 537Mb).


NadavTasher:
As you can see, `ghcr.io/wekan/wekan:pr-5588` is much smaller then `ghcr.io/wekan/wekan:latest`.
The build https://github.com/wekan/wekan/actions/runs/11990748626/job/33428596619?pr=5588 only took 5m53s, compared to https://github.com/wekan/wekan/actions/runs/11943125448/job/33291559590 which took 15m14s.

xet7:
You changed to use older Node.js 14.21.3 that has more vulnerabilities.

Newest Node.js 14 is 14.21.4 that was released by Meteor: https://github.com/wekan/node-v14-esm/releases/tag/v14.21.4

NadavTasher:
You are correct. I overlooked that when I was looking for the latest v14 Node version.
I have updated the `Dockerfile` to use the official bundle provided by Meteor (can be found [here](https://guide.meteor.com/using-node-v14.21.4#download))

xet7:
The point in using Ubuntu 24.10 base image is, that WeKan has always used newest Ubuntu base image. When next Ubuntu 25.04 is released, WeKan will change to use that newest release. Newest Ubuntu always has newest versions of software for all dependencies. Please change base image back to 24.10.

NadavTasher:
This is not concidered a best practice.
Please see https://hub.docker.com/_/ubuntu:
> The ubuntu:latest tag points to the "latest LTS", since that's the version recommended for general use. The ubuntu:rolling tag points to the latest release (regardless of LTS status).

mfilser:
Using LTS base image sounds for me better, too 🙂

xet7:
@NadavTasher 

Please remove setting stack size. I reverted setting stack size at WeKan v7.63, because that setting made WeKan much slower.

https://github.com/wekan/wekan/releases/tag/v7.63

NadavTasher:
You have not completely reverted this change, as can be seen here: https://github.com/wekan/wekan/blob/main/Dockerfile#L276
I thought you wanted to keep this `ulimit` so I left it this way.
I'll gladly remove it if you want.

xet7:
@NadavTasher 

I removed this from everywhere:
```
--stack-size=65500 --max-old-space-size=8192
```
Ulimit needs to be there.

NadavTasher:
This is exactly what I have done :)
`ulimit --help` will tell you that `-s` is `stack size`, which is why I called the variable `STACK_SIZE`.

Dockerfile
```diff
@@ -158,119 +203,35 @@ ENV \
...
+# Change to the wekan user
 USER wekan
 
+# Set the port to listen on
 ENV PORT=8080
-EXPOSE $PORT
+EXPOSE ${PORT}
+
+# Set the maximum stack size
+ENV STACK_SIZE=65500
```
xet7, line 228: This line is the problem 1.

Dockerfile
```diff
@@ -158,119 +203,35 @@ ENV \
...
-# Add more stack:
-#   bash -c "ulimit -s 65500; exec node --stack-size=65500 main.js"
-#---------------------------------------------------------------------
-#
-# CMD ["node", "/build/main.js"]
-# CMD ["bash", "-c", "ulimit -s 65500; exec node --stack-size=65500 /build/main.js"]
-# CMD ["bash", "-c", "ulimit -s 65500; exec node --stack-size=65500 --max-old-space-size=8192 /build/main.js"]
-CMD ["bash", "-c", "ulimit -s 65500; exec node /build/main.js"]
+# Set startup command
+CMD ["bash", "-c", "ulimit -s ${STACK_SIZE} && exec node /wekan/main.js"]
```
xet7, line 237: This is the problem 2, to be removed:
```
-s ${STACK_SIZE}
```

NadavTasher: https://github.com/wekan/wekan/blob/main/Dockerfile#L276 does `-s 65500`.

Dockerfile
NadavTasher: I can make it non-configurable if this is what you want, but if you wanted to eliminate this completely, that can be done just bear in mind that it is now the same behaviour as https://github.com/wekan/wekan/blob/main/Dockerfile#L276

Dockerfile
xet7: I think it should be about The maximum number of open file descriptors. Not about stack size.
```
ulimit -n 65500
```
https://ss64.com/bash/ulimit.html

NadavTasher: Furthermore, If you want `-s $(STACK_SIZE}` removed, it would mean you want `ulimit && node /wekan/main.js`?
This does not do anything at all, as `ulimit` without parameters just displays the current `ulimit` (see https://ss64.com/bash/ulimit.html):
> If limit is given, it is the new value of the specified resource. Otherwise, the current value of the soft limit for the specified resource is printed, unless the '-H' option is supplied.

Would you like me to change to:
```dockerfile
CMD [ "node", "/wekan/main.js" ]
```
?

NadavTasher: > I think it should be about The maximum number of open file descriptors. Not about stack size.
> 
> ```
> ulimit -n 65500
> ```
> 
> https://ss64.com/bash/ulimit.html

I see, so there is a mistake on `main`.
Should l make the change?
It seems to me like it is not required, since it did not exist in `main`, and we'd just be better off with `CMD [ "node", "/wekan/main.js" ]`.

xet7: Stack size setting removed, and only this:
```
CMD ["bash", "-c", "ulimit -n 65500 && exec node /wekan/main.js"]
```

NadavTasher: Acknowledged. Making the changes now.

xet7: @NadavTasher 

Original reasons are these, and I still am not sure about what would be correct settings:
- https://github.com/wekan/wekan/issues/3585#issuecomment-1021522132
- https://stackoverflow.com/questions/20936486/node-js-maximum-call-stack-size-exceeded
- https://github.com/wekan/wekan/issues/5347

Setting stack too high causes too much RAM usage, that causes server to start swapping.

Also, because of memory leaks https://github.com/wekan/wekan/issues/5140 there is a very bad workaround:

https://github.com/wekan/wekan/wiki/Cron

NadavTasher: @xet7 

These two problems do not seem to have any relation to maximim FDs

xet7: Ok, these are not related to FDs.

NadavTasher: I will change the Dockerfile.

xet7:
@NadavTasher 

With these changes, `wekan-app` container is restarting many times, never stopping restarting:
```
wekan@dev:~/repos/wekan$ docker ps
CONTAINER ID   IMAGE                        COMMAND                  CREATED         STATUS                          PORTS       NAMES
906d2bf404fa   ghcr.io/wekan/wekan:latest   "node /wekan/main.js"    7 minutes ago   Restarting (1) 25 seconds ago               wekan-app
7fcff851c83c   mongo:6                      "docker-entrypoint.s…"   7 minutes ago   Up 7 minutes                    27017/tcp   wekan-db
```
It looks like attachments volume is broken:
```
wekan@dev:~/repos/wekan$ docker logs wekan-app
/wekan/programs/server/node_modules/fibers/future.js:280
						throw(ex);
						^
errorClass [Error]: [FilesCollection.attachments] Path "/data/attachments" is not writable! [401]
    at new FilesCollection (packages/ostrio:files/server.js:356:15)
    at module (models/attachments.js:49:15)
    at fileEvaluate (packages/modules-runtime.js:336:7)
    at Module.require (packages/modules-runtime.js:238:14)
    at Module.moduleLink [as link] (/wekan/programs/server/npm/node_modules/meteor/modules/node_modules/@meteorjs/reify/lib/runtime/index.js:52:22)
    at module (server/publications/attachments.js:1:24)
    at fileEvaluate (packages/modules-runtime.js:336:7)
    at Module.require (packages/modules-runtime.js:238:14)
    at require (packages/modules-runtime.js:258:21)
    at /wekan/programs/server/app/app.js:194586:1
    at /wekan/programs/server/boot.js:385:38
    at Array.forEach (<anonymous>)
    at /wekan/programs/server/boot.js:210:21
    at /wekan/programs/server/boot.js:439:7
    at Function.run (/wekan/programs/server/profile.js:256:14)
    at /wekan/programs/server/boot.js:438:13 {
  isClientSafe: true,
  error: 401,
  reason: '[FilesCollection.attachments] Path "/data/attachments" is not writable!',
  details: Error: EACCES: permission denied, mkdir '/data/attachments'
      at Object.mkdirSync (fs.js:1014:3)
      at new FilesCollection (packages/ostrio:files/server.js:350:10)
      at module (models/attachments.js:49:15)
      at fileEvaluate (packages/modules-runtime.js:336:7)
      at Module.require (packages/modules-runtime.js:238:14)
      at Module.moduleLink [as link] (/wekan/programs/server/npm/node_modules/meteor/modules/node_modules/@meteorjs/reify/lib/runtime/index.js:52:22)
      at module (server/publications/attachments.js:1:24)
      at fileEvaluate (packages/modules-runtime.js:336:7)
      at Module.require (packages/modules-runtime.js:238:14)
      at require (packages/modules-runtime.js:258:21)
      at /wekan/programs/server/app/app.js:194586:1
      at /wekan/programs/server/boot.js:385:38
      at Array.forEach (<anonymous>)
      at /wekan/programs/server/boot.js:210:21
      at /wekan/programs/server/boot.js:439:7
      at Function.run (/wekan/programs/server/profile.js:256:14)
      at /wekan/programs/server/boot.js:438:13 {
    errno: -13,
    syscall: 'mkdir',
    code: 'EACCES',
    path: '/data/attachments'
  },
  errorType: 'Meteor.Error'
}
```
This was built with this command to build Dockerfile and start WeKan:

```
docker compose up -d --build
```
These changes were done to docker-compose.yml so that it builds Dockerfile from source:
```
wekan@dev:~/repos/wekan$ git diff
diff --git a/docker-compose.yml b/docker-compose.yml
index bfc1bce6a..838164ded 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -141,9 +141,9 @@ services:
     #-------------------------------------------------------------------------------------
     # ==== BUILD wekan-app DOCKER CONTAINER FROM SOURCE, if you uncomment these ====
     # ==== and use commands: docker compose up -d --build
-    #build:
-    #  context: .
-    #  dockerfile: Dockerfile
+    build:
+      context: .
+      dockerfile: Dockerfile
     #-------------------------------------------------------------------------------------
     ports:
       # Docker outsideport:insideport. Do not add anything extra here.
```

NadavTasher:
Okay, this is an issue with the permissions on the `/data` directory.
I will fix this today.

xet7:
@NadavTasher 

```
root@dev:/home/wekan/repos/wekan# docker exec -it wekan-app bash
wekan@2d3cf19e0d01:~/app$ ls -lah /data
total 16K
drwxr-xr-x 4 wekan root  4.0K Nov 24 08:15 .
drwxr-xr-x 1 root  root  4.0K Nov 24 08:15 ..
drwxr-xr-x 2 wekan wekan 4.0K Nov 24 08:15 attachments
drwxr-xr-x 2 wekan wekan 4.0K Nov 24 08:15 avatars
```

NadavTasher:
@xet7 

I have pushed a commit that fixes this.
I was caused by a missing command.

Will you be able to verify that `ls -lAR` returns the same output? (Works for me)

PR #5589, opened 2024-11-24
Opinion - Change archive-card shortcut to backtick for better ergonomics

NadavTasher:
This was requested [here](https://github.com/wekan/wekan/pull/5576) by @mohammadZahedian.

This improves the ergonomics by allowing single left-handed handling of shortcuts.

IMO, this shortcut is a bit too close to the `ESC` button, which might cause misclicks.

@xet7 - What do you think?

### DIFF
--- a/client/lib/keyboard.js
+++ b/client/lib/keyboard.js
@@ -190,7 +190,7 @@ Mousetrap.bind('space', evt => {
   }
 });
 
-Mousetrap.bind('-', evt => {
+Mousetrap.bind('`', evt => {
   const cardId = getSelectedCardId();
   if (!cardId) {
     return;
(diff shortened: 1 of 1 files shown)


NadavTasher:
Might be possible, but @mohammadZahedian has specific requirements, and not all a-z can be used, as some of them are special in combination with Ctrl.

xet7:
Ok. I think this is OK.

NadavTasher:
Am not so sure, but we'll see.
If it is too annoying we can always revert it.

mohammadZahedian:
@NadavTasher 
it is much better now.
but it does work in persian keyboard , why?

mohammadZahedian:
> Is there some problem? Does it work or not?

that was miss typing.
it doesn't work in persian

mohammadZahedian:
'÷'

Issue #5593, opened 2024-11-26
list width too long for long titles


joshlupo:
### Server Setup

* Did you test in newest Wekan?: **yes**
* Hosted via pikapods.com
* What webbrowser version are you using?: **Brave**

### Problem description
- Lists with long titles are too wide after PR #5581 in my opinion
- It's not a nice experience to read a paragraph of text in one line across a wide monitor.  It doesn't look much like a "card" then.  The version of Wekan before this PR handled this nicer in my opinion.  Lists had a fixed width that was much smaller, which made them easier to read, look more like cards, and have a more consistent viewing experience across different boards that each have a different number of lists.
![image](https://github.com/user-attachments/assets/ee7da88e-272a-48c5-b906-0444f4bb8fdc)

### Reproduction Steps
- Create any board with a single list and card that has a long title.

### Recommendations
- My choice would be to revert the change. If others don't agree then I suggest adding a new option to set the **Max** Width, in addition to the Min Width option.
![image](https://github.com/user-attachments/assets/afa391c2-21c6-40e8-b843-dfb54ba01e87)

xet7:
@NadavTasher 

Please send PR, where default is previous list width, and optional is fill up space evenly. Thanks!

NadavTasher:
I will get to that today / tomorrow.

NadavTasher:
@joshlupo 
This happens only when you have just one list.
If you have more then one list it handles them great.

@xet7 WDYT?

xet7:
@NadavTasher 

Every list should have that original small width by default. Anything else is optional.

NadavTasher:
Okay. I will work on that ASAP.

NadavTasher:
I will open a PR soon.

computrav:
+1 on this request, however, some thoughts (all IMO of course!):
( @NadavTasher if you are working on this )

1. Both a MIN width and MAX width should be able to be defined for list widths. For MAX, it could be definable in both pixels or a special value of "auto max width" (but a better name) -- meaning it would act like it does now where the list width = total width divided by the number of lists. I still think a default of a fixed pixel width is more likely to be desired by users, however. Bare minimum though, having both MIN and MAX in pixels would be MVP.
2. Each value should have sensible defaults - e.g., 270px for default MIN width, but also a default MAX width, perhaps something around 400px.
3. These settings should be able to be defined both at the application level (via Admin panel), but also per-board (like min-width is now).

[...]

PR #5607, opened 2024-12-07
New feature - Optional board list auto-width, Support for min & max width for lists

NadavTasher:
This feature comes as a solution for issue https://github.com/wekan/wekan/issues/5593.

This PR introduces changes to allow for optional automatic widths for all lists in a board.
It also adds support for minimum and maximum width for each list (in pixels).

A new toggle was added:
image
image

When auto-width is enabled, lists will try to expand to fill the size of the screen, as demonstrated here:
image

When there are too many lists and this feature is enabled, the lists will just add up, each list taking just the `min-width` as defined in the `Set Widths` popup:
image

Here's how the new `Set Widths` popup looks (note that the values in the screenshot are not the defaults):
image

Please thoroughly review the changes in this PR before merging.


mfilser:
@NadavTasher
is it possible to use functions from the Util class, e. g. this one?
https://github.com/wekan/wekan/blob/a3f70735ac174ac6c1c37008680e30312bf7fdbc/client/lib/utils.js#L44

NadavTasher:
> @NadavTasher is it possible to use functions from the Util class, e. g. this one?
> 
> https://github.com/wekan/wekan/blob/a3f70735ac174ac6c1c37008680e30312bf7fdbc/client/lib/utils.js#L44

@mfilser 
`Utils.getCurrentBoardId()` seems like a better fit here.
I have pushed a commit that uses it.
Thanks!

mfilser:
I think in the last commit you changed to Utils: It's good to get the boardId from the list data.

My thought was more to replace the other parts of Session.get with the Utils class.

Sorry, I am on my mobile right now, so I can't write well.

NadavTasher:
Oh I see.
I have reverted that change and updated the code in `boardHeader.js` to use the aforementioned utility function.

I only used `Session.get` because it was used in the line above (I just copied it)

mfilser:
I see. Sometimes that deprecated code comes in.
To use the Utils function is necessary for the mobile version.

Could you change this too?

const boardId = Session.get('currentBoard');

NadavTasher:
I see I forgot to commit this change.
Done. 😄

mfilser:
Great, thank you very much, and also for your contribution.
I didn't have the time to try and test, I hope I can do it soon...

PR #5611, opened 2024-12-09
Cards.js Architecture Change

AdenKoziol:
The post method before the change handled business logic, repository logic, and the post logic all within one method. This would be hard for a developer to understand if something was broken or if new logic needed to be added. The fix now separates the business and repository logic from the post method so it is much easier for a developer to understand and maintain. I made individual methods for business logic for authentication and updating custom fields and repository logic for fetching cards and updating their fields.


mfilser:
The diff is harder to read because of indentation changes. Is this necessary too?

PR #5619, opened 2024-12-12
Update OIDC configuration

walster001:
Fix the fact that it didn't add the later part of the code which actually checks and adds user to board/updates their permissions on the first time run. Code has also been optimised to use .includes instead of .contains


walster001:
Hey @xet7 I figured out how to do the draft pull req thing!

I was going through analysing the code and comparing it to the last modification done in April. And found that AI left off the bottom half of the code, so double checked it piece by piece. Added in the board updates at the bottom and verified everything else. I will add those keys into the other spots also. 

Would appreciate review and thoughts :) 

Probably important to note the code will fail in my other pull req if released so might be good to revert that pull. Apologies!

walster001:
I still haven't been able to build a test server so this code is untested but its more tested than the code I submitted yesterday. Trying to build it on windows initially but then decided I would have to use a linux machine :) 

Sorry everyone!

packages/wekan-oidc/oidc_server.js
```diff
@@ -180,16 +181,16 @@ var getToken = function (query) {
...
-    var serverUserinfoEndpoint = config.userinfoEndpoint.includes("https://") ?
+    var serverUserinfoEndpoint = config.userinfoEndpoint.includes("https://") ? 
         config.userinfoEndpoint : config.serverUrl + config.userinfoEndpoint;
 
     var response;
     try {
         var getOptions = {
             headers: {
-                "User-Agent": "Meteor",
+                "User-Agent": userAgent,
```
xet7, line 193: @walster001 

What is happening here?

xet7:
I think it is not a good idea to read something from file path.

Data should at database or environment variables.

I'll think later more.

packages/wekan-oidc/oidc_server.js
walster001: AI Modified the useragent to be "Meteor" instead of it's original variable "userAgent" according to the last submitted version of oidc_server.js
However, is userAgent even set as a variable or would it benefit from being set to meteor permanently?

walster001:
> I think it is not a good idea to read something from file path.
> 
> Data should at database or environment variables.
> 
> I'll think later more.

No worries :)

Just for reference however, would you say the oidc_server.js file and login flow where we want to insert it is correct? We've been searching through the code trying to find where to implement it even if we just do it locally for us now while testing. I'm hoping this file is correct, hence why I'm trying to get a test server/build running :) 

The other implementation idea we had was to store database access details ie URL/username/pass/etc as environment variables so that an external database could be accessed, read, check a specific column for emails then return the result back to wekan and proceed with the login process if the email was there, if not revert user back to login screen. 

We specifically want to read an external MYSQL database :) query it and then return a string to wekan oidc server.

PR #5633, opened 2024-12-29
Move auto-width toggle to top of list width popup

NadavTasher:
Looks better IMO - I like that the last think in the popup is the Apply button.


xet7:
It is not better.

Apply button affects those options that are above Apply button.

Automatic list width is a once-click toggle button. It is not related in any way to form and Apply button that are above of Automatic list width.

NadavTasher:
Ok.

What if we put it in a different section? Like add a title with a description of what it does?

xet7:
@NadavTasher 

All options at that popup are already about list width. It already works OK.

PR #5635, opened 2024-12-29
Move swimlane creation button to board header when using swimlane view mode

NadavTasher:
When all swimlanes are deleted, no new swimlanes can be created.
This PR moves the button to the board header, but is only visible in "Swimlanes" view mode.

New swimlanes view:
New swimlanes view

Lists view (notice the `+` button is missing):
Lists view

### DIFF
--- a/client/components/boards/boardColors.css
+++ b/client/components/boards/boardColors.css
@@ -2181,9 +2181,6 @@ THEME - Modern Dark
   padding: 6px;
   font-size: 16px;
 }
-.board-color-moderndark .swimlane .swimlane-header-wrap .swimlane-header-plus-icon {
-  font-size: 16px;
-}
 .board-color-moderndark .swimlane {
   background: #2a2a2a;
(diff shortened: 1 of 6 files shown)


xet7:
@NadavTasher 

Please change New Swimlane button back to be where it originally was.

When there is No Swimlanes, please show there New Swimlane button, like this. Thanks!

swimlane2

NadavTasher:
Why?

xet7:
@NadavTasher 

Because that way keeps original WeKan design.

Please do not move buttons from their original location to Board Header. Thanks!

NadavTasher:
Can we just have this button there all the time instead of having a `+` button on each swimlane header?

xet7:
Button to add New Swimlane at each Swimlane Header is for adding new Swimlane below current Swimlane, without need to scroll to top of board. This is by design.

NadavTasher:
Changed to:
image

PR #5643, opened 2024-12-30
Allow vertical scrollbars to be disabled (new preference)

NadavTasher:
Allow users to enable or disable vertical scrollbars:

Default (with scrollbars):
image

Optional (disabled scrollbars):
image

The default behaviour keeps WeKan as it was.
This feature was requested by some colleagues of mine.


NadavTasher:
@xet7 
This feature adds another string that needs translation.

PR #5652, opened 2025-01-01
Feature - Add toggle for week-of-year in date displays (ISO 8601)

NadavTasher:
Default (same as before):
image

New optional "Disabled" state:
image

### DIFF
--- a/client/components/cards/cardCustomFields.jade
+++ b/client/components/cards/cardCustomFields.jade
@@ -79,13 +79,14 @@ template(name="cardCustomField-currency")
 
 template(name="cardCustomField-date")
     if canModifyCard
-      a.js-edit-date(title="{{showTitle}} {{_ 'predicate-week'}} {{showWeek}}" class="{{classes}}")
+      a.js-edit-date(title="{{showTitle}} {{_ 'predicate-week'}} {{#if showWeekOfYear}}{{showWeek}}{{/if}}" class="{{classes}}")
         if value
           div.card-date
             time(datetime="{{showISODate}}")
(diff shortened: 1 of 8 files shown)


xet7:
@NadavTasher 

Hmm, now there is Buttonitis at Right Sidebar.

What do you think, could you change places of card settings buttons this way, so that card settings would be at card etc?

- Right Sidebar / Board Settings / Card Settings => Opened Card / ☰ ⚙   (at Card Titlebar Cog Settings ⚙ icon)
- Right Sidebar / Hide minicard label text =>  Opened Card / ☰ ⚙   (at Card Titlebar Cog Settings ⚙ icon)
- Right Sidebar / Board Settings / Custom Fields => Opened Card / ☰
- Right Sidebar / Board Settings / Subtasks Settings => Opened Card / Subtasks ⚙ (at Subtasks section of card)
- Right Sidebar / Enable vertical scrollbars [_] => List ☰
- Right Sidebar / Show week of year [_] => Opened Card / Due Dates ⚙ (settings at right side of Due Dates etc)

NadavTasher:
These settings affect things that are not just cards (e.g. Vertical scrollbars also affects the board body and sidebar)

Personally, I think that this small toggle is not that bad since it is not visible all of the time, only when the sidebar is open.

Do you think we can merge this and work on movong these settings another time?

xet7:
Yes. I merged this.

Moving settings is only an idea. We will see later is there time for that, is it useful enough.

NadavTasher:
I will move it when I have some extra time.

PR #5710, opened 2025-03-08
added feature of confirm pop up before archiving or cloning a board

Rayene123:
added feature of confirm pop up before archiving or cloning a board

### DIFF
--- a/client/components/boards/boardsList.js
+++ b/client/components/boards/boardsList.js
@@ -239,34 +239,41 @@ BlazeComponent.extendComponent({
           evt.preventDefault();
         },
         'click .js-clone-board'(evt) {
-          let title = getSlug(ReactiveCache.getBoard(this.currentData()._id).title) || 'cloned-board';
-          Meteor.call(
-            'copyBoard',
-            this.currentData()._id,
-            {
-              sort: ReactiveCache.getBoards({ archived: false }).length,
-              type: 'board',
-              title: ReactiveCache.getBoard(this.currentData()._id).title,
-            },
-            (err, res) => {
-              if (err) {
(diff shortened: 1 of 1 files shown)


mfilser:
@Rayene123
"Update boardsList.js" isn't a helpful commit subject too.
Can you add same details to commit subject in git than this PR has as subject?

xet7:
This does not seem to use translations at all, it's only English.

This also does not use `Popup.afterConfirm(` for WeKan styled confirm popups.

mfilser:
@xet7
Is there a API endpoint to clone boards?
If yes, is it still working with the API ?

xet7:
@mfilser 

Yes, there is API endpoint:

https://github.com/wekan/wekan/blob/main/models/boards.js#L2285

You can use it with api.py that is at https://github.com/wekan/wekan :

https://github.com/wekan/wekan/blob/main/api.py#L47

It works:

```
$ ./api.py copyboard RwhZbyrNDDatkiX9x "CopiedBoard"
http://localhost:4000/api/boards/RwhZbyrNDDatkiX9x/copy
=== COPY BOARD ===

"Lv64GpLqMujofBYR7"
Succesfull!
```

Rayene123:
> This does not seem to use translations at all, it's only English.
> 
> This also does not use `Popup.afterConfirm(` for WeKan styled confirm popups.

you didn't mention that i need to use wekan styped confirm popups

xet7:
@Rayene123 

> you didn't mention that i need to use wekan styled confirm popups

Sorry, docs needs improvements. I was able to mention it above only after testing this PR. Thanks for your contributions!

Issue #5684, opened 2025-02-12
Exception while invoking method 'copyCard' TypeError: Cannot read property 'name' of undefined


xet7:
```
Exception while invoking method 'copyCard' TypeError: Cannot read property 'name' of undefined
    at models/cards.js:537:21
    at Array.map (<anonymous>)
    at Document.mapCustomFieldsToBoard (models/cards.js:533:30)
    at Document.copy (models/cards.js:579:32)
    at MethodInvocation.copyCard (models/cards.js:3089:24)
    at packages/check/match.js:118:15
    at packages/meteor.js:365:18
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1389:31)
    at Object._failIfArgumentsAreNotAllChecked (packages/check/match.js:116:43)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1935:18)
    at getCurrentMethodInvocationResult (packages/ddp-server/livedata_server.js:808:38)
    at packages/meteor.js:365:18
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1389:31)
    at packages/ddp-server/livedata_server.js:827:46
    at new Promise (<anonymous>)
    at Session.method (packages/ddp-server/livedata_server.js:775:23)
    at packages/ddp-server/livedata_server.js:639:43
```

Firas-Git:
The error message you're encountering, TypeError: Cannot read property 'name' of undefined, indicates that the code is trying to access the name property of an object, but the object is undefined. This occurs in the file models/cards.js, specifically at line 537.

- The error originates from the method copyCard, where it's trying to map over a set of custom fields, and somewhere along the line, it encounters an undefined object that doesn't have the name property.

- The function mapCustomFieldsToBoard is trying to iterate through a collection (likely of fields or cards) but fails because one of the objects in the collection is undefined or has an unexpected structure.
Possible Fix for the function mapCustomFieldsToBoard:

mapCustomFieldsToBoard(boardId) {
    // Map custom fields to new board
    return this.customFields.map(cf => {
        const oldCf = ReactiveCache.getCustomField(cf._id);

        // Check if oldCf is undefined or null
        if (!oldCf) {
            console.error(`Custom field with ID ${cf._id} not found.`);
            return cf;  // Skip this field if oldCf is not found
        }

        const newCf = ReactiveCache.getCustomField({
            boardIds: boardId,
            name: oldCf.name,
            type: oldCf.type,
[...]

PR #5711, opened 2025-03-08
fix the "copyCard" error

Firas-Git:


xet7:
Fixes #5684 . Thanks to @Firas-Git !

Issue #5434, opened 2024-06-05
Adding attachments to cards semi-fails.


inDane:
### Server Setup Information

* Did you test in newest Wekan?:
** Newest release in sandstorm is: 6.15.0~2022-04-05, yes.
* Did you configure root-url correctly so Wekan cards open correctly (see https://github.com/wekan/wekan/wiki/Settings)?
** They open correctly
* Operating System:
** ubuntu 24.04
* Deployment Method (Snap/Docker/Sandstorm/bundle/source): 
** Sandstorm (v308)
* Http frontend if any (Caddy, Nginx, Apache, see config examples from Wekan GitHub wiki first):
** nginx
* Node.js Version: n/a
* MongoDB Version: n/a
* What webbrowser version are you using (Wekan should work on all modern browsers that support Javascript)?
** firefox 126.0.1

### Problem description
I've upgraded my WeKan Grain from 5.xx  to the latest 6.15.0. Unfortunately the attachments are not working anymore. It is generating a thumbnail and attaches this to the card, but the actual attachment is not available. I cannot open nor download it. If i delete it i get an error log (see below in Logs section).
`/opt/sandstorm/var/sandstorm/grains/McnQ4XKT6vZFsDBQsXctBE/sandbox/wekan-uploads/uploads/attachments`
is writable for sandstorm user.

![image](https://github.com/wekan/wekan/assets/5693407/9b3febf3-71a5-43bd-825d-54065c779c40)

Please let me know if I should provide more info.

#### Reproduction Steps
Create Wekan grain, add attachment.

#### Logs

When deleting "broken" attachment, I can see this in /opt/sandstorm/var/sandstorm/grains/McnQ4XKT6vZFsDBQsXctBE/log

```
Exception in callback of async function: TypeError: Cannot read property 'name' of undefined
    at Object.<anonymous> (models/activities.js:251:38)
    at packages/matb33_collection-hooks.js:422:18
    at Array.forEach (<anonymous>)
    at Function._.each._.forEach (packages/underscore.js:139:11)
    at after (packages/matb33_collection-hooks.js:421:9)
    at Object.<anonymous> (packages/matb33_collection-hooks.js:436:12)
    at Object.collection.<computed> [as insert] (packages/matb33_collection-hooks.js:150:21)
    at ns.Collection.insert (packages/mongo/collection.js:525:39)
    at ns.Collection.Mongo.Collection.<computed> [as insert] (packages/aldeed_collection2-core.js:232:19)
[...]

PR #5713, opened 2025-03-09
bug fixing number #5434

Firas-Git:
In the section of code that handles activities related to labels (activity.labelId), I added a null check to ensure that the label object exists before trying to access its properties (name or color).
**Explanation:**
Before attempting to access properties like name or color of the label object, I made sure that label exists by adding the condition if (label) {}.
This prevents trying to access properties on an undefined label, which was causing the error.


xet7:
@Firas-Git 

But but... this looks like it removes BIGEVENTS code?

PR #5733, opened 2025-04-10
Fixing adding attachments to cards semi fails 

Firas-Git:
This PR fixes a bug in the Activities.after.insert hook where the .subtasks() helper method was being incorrectly used to retrieve subtask information. The method returns a single card (not a list of subtasks), which led to runtime errors when trying to access properties like title on an undefined or unexpected structure.

Changes Made:
Replaced .subtasks() with .card() in the subtaskId handling block.

Updated variable naming for clarity (subtask instead of subtasks) to reflect that a single card is retrieved.

Verified that the correct card title and ID are now being set in params.

### DIFF
--- a/models/activities.js
+++ b/models/activities.js
@@ -155,7 +155,10 @@ if (Meteor.isServer) {
     }
     if (activity.memberId) {
       participants = _.union(participants, [activity.memberId]);
-      params.member = activity.member().getName();
+      const member = activity.member();
+      if (member) {
+        params.member = member.getName();
+      }
     }
     if (activity.listId) {
       const list = activity.list();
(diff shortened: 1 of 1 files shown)


mfilser:
Thank you for your contribution.
For me, as far as i see on my mobile there was some code removed, like
integrations.forEach

and
export default Activities

Is this and other changed needed? Did you test this commit and other things around in Wekan?
I have a gut feeling that something could be broken  by this commit...

Firas-Git:
Hi! Thank you for your thoughtful review.

You're right — after checking again, I noticed that integrations.forEach and export default Activities were unintentionally removed during my edits. These were not meant to be deleted, and I appreciate you catching that.

The main goal of my commit was to fix the error in Activities.after.insert, specifically to prevent issues when params was undefined. I tested that part thoroughly, but I’ll make sure to restore the removed code and retest everything around it to avoid any regressions.

I'll push an updated commit shortly with those fixes in place.

Thanks again for your help!

mfilser:
Your welcome, thank you too 🙂

xet7:
Hmm, I tried to merge some other PR, and it merged also this. Strange. I'll try to revert this.

xet7:
Hmm, it looks like there were no changes to `models/activities.js` anyway, so it seems this was somehow already reverted.

PR #5765, opened 2025-05-12
Fix #5732: incorrect list insertion order

nourguidara:
Previously, lists added using the "Add After" functionality were incorrectly inserted at the end of the board. This PR uses `Utils.calculateIndexData` to correctly position the new list immediately after the selected list.

Fixes #5732

PR #5766, opened 2025-05-12
Fix #5732 list insert position

### LINKED ISSUE #5732: New List Misplaced at End Instead of After Specified List
Hello,

When attempting to add a new list using the "Add After" functionality (e.g., adding "List 5" after "List 1"), the new list is incorrectly positioned at the end of all existing lists, rather than immediately following the specified list.

here in the example: list 5 is added at the end after list 4, instead of being added after list 1

Image
Image

nourguidara:
This PR replaces and improves on #5765  and still addresses #5732 

It fixes two issues:

-Previously, lists added using "Add After" were inserted at the end instead of right after the selected list.
-A runtime error occurred when the .list-position-input

 I have tested this manually:

-Adding a list "after" another works and respects list order
-The browser console shows no JavaScript errors
test1
yes

### DIFF
--- a/client/components/swimlanes/swimlanes.js
+++ b/client/components/swimlanes/swimlanes.js
@@ -250,17 +250,18 @@ BlazeComponent.extendComponent({
     return [
       {
         submit(evt) {
-          evt.preventDefault();
-          const titleInput = this.find('.list-name-input');
-          const title = titleInput.value.trim();
-          const positionInput = this.find('.list-position-input');
+        evt.preventDefault();
+        const titleInput = this.find('.list-name-input');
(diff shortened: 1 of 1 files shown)


mfilser:
@nourguidara
normally, selections should use ReactiveCache, here, ReactiveCache.getList was replaced with another selection

valhalla-creator:
I believe I’m experiencing the same issue, and I’ll be creating a PR to fix these problems. Instead of replacing the use of ReactiveCache, I’ll ensure that all selections properly leverage it for consistency and reactivity.

xet7:
Moved to https://github.com/wekan/wekan/pull/5767

Issue #5782, opened 2025-05-21
After bugfix/fix due date #5774, the start date, due date, and end date are displayed wrong


bbyszio:
After bugfix/fix due date #5774, the start date, due date, and end date are displayed in American format on the minimap. However, the language is set to German. The receipt date is displayed correctly on the mini card. The dates are displayed correctly on the open card.

![Image](https://github.com/user-attachments/assets/b114670c-8c7f-47de-80ff-52cef36ed1f4)
![Image](https://github.com/user-attachments/assets/d2175c94-5b63-4c6b-a816-9c8d3d13c32c)
![Image](https://github.com/user-attachments/assets/2ccbefe9-105a-4717-aa6f-53b2ab11f802)

PR #5786, opened 2025-05-23
Bugfix/date format minimap 5774

valhalla-creator:
title: Fix Due Date Problem in Non-English Numbers https://github.com/wekan/wekan/issues/5782

Added proper number normalization for Persian/Arabic digits using regex patterns
Added a new formatDate utility method in Utils that handles both date-only and date-time formats
Used moment.js locale-aware formatting with 'L' for dates and 'LT' for times
Ensured consistent date formatting across the entire application
Added proper imports for all dependencies including moment.js
This solution ensures dates are displayed correctly in all locales while maintaining consistent formatting between minicards and full cards. The digit normalization handles Persian and Arabic numbers seamlessly before any date operations occur.

### DIFF
--- a/client/components/cards/cardDate.js
+++ b/client/components/cards/cardDate.js
@@ -3,13 +3,6 @@ import moment from 'moment/min/moment-with-locales';
 import { TAPi18n } from '/imports/i18n';
 import { DatePicker } from '/client/lib/datepicker';
 
-// Add this helper function at the top level
-const formatDate = (date, format) => {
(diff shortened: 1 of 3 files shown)


xet7:
@valhalla-creator 

This shows empty screen and popup error. Login does not work, because you removed flowrouter import code.

flowrouter-error

xet7:
Moved to https://github.com/wekan/wekan/pull/5787

PR #5800, opened 2025-06-11
fix: update OpenAPI according to wiki

Atry:
The fields are updated according to https://github.com/wekan/wekan/wiki/REST-API-User

I also added an endpoint to create a board according to https://github.com/wekan/wekan/wiki/REST-API-Boards

### DIFF
--- a/public/api/wekan.yml
+++ b/public/api/wekan.yml
@@ -31,50 +31,55 @@ paths:
       operationId: login
       summary: Login with REST API
       consumes:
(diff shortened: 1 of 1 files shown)


xet7:
@Atry 

You can not edit wekan.yml directly, because API docs wekan.yml is generated with this script, and changes would always be overwritten:

https://github.com/wekan/wekan/blob/main/releases/rebuild-docs.sh#L43

PR #5812, opened 2025-06-19
Add assigneeId to webhooks

Atry:


xet7:
It should not include emails.

PR #5815, opened 2025-06-23
feat: enhance API documentation and add board-related endpoints

Atry:
Previously, boards.js contains a `?.` operator, which is not supported by esprima, therefore boards.js is skipped.

This PR fixes the API generation for borads.js by removing the `?.` operator.

See https://github.com/jquery/esprima/issues/2112


mfilser:
@Atry 
Maybe I don't see it, but I think Wekan isn't using esprima at all. As i see esprima is a third party framework evaluating a text expecting Javascript inside. Your PR is something different...

Atry:
https://github.com/wekan/wekan/blob/f399b5119f2a48d19546d657e8d2ade4331505ef/openapi/README.md
OpenAPI is generated by parsing JSDoc using esprima-python

Atry:
@xet7 Do you think if this PR is ready to merge?

PR #5871, opened 2025-08-20
docs(windows): WRITABLE_PATH requirement and BIND_IP/ROOT_URL tip for…

sowwos:
… LAN dev


xet7:
That info is already at step 8:

https://github.com/wekan/wekan/blob/main/docs/Platforms/Propietary/Windows/Offline.md

Issue #5887, opened 2025-09-25
Collapse button not there when using clearlight board theme


ghost:
I've noticed a bug where the Collapse button does not appear at the top when using the Cleanlight board theme.

I haven't tested on all themes so that it could be more widespread than just the Cleanlight theme.

<img width="1454" height="91" alt="Image" src="https://github.com/user-attachments/assets/25e8e6af-b189-4780-abf9-60303c3f4d19" />

<img width="386" height="965" alt="Image" src="https://github.com/user-attachments/assets/d0a8fd73-35b4-4050-b283-bb0efd18d117" />

This is what I expect to see:
<img width="357" height="75" alt="Image" src="https://github.com/user-attachments/assets/672b9ae6-90da-412a-bb2c-751b4f3b183c" />

PR #5900, opened 2025-10-07
fixed issue 5887

seve12:


xet7:
Fixes #5887 . Thanks to seve12 !

Issue #6049, opened 2025-12-28
Announcements cannot be closed


Chostakovitch:
Since https://github.com/wekan/wekan/commit/2947238a021b6952b56e828d49a8c0094520d89a, the class allowing to close announcement popup is gone, such as only admins disabling the announcement globally can dismiss them.

Now, the red cross is only textual as seen there: https://github.com/wekan/wekan/blob/main/client/components/main/header.jade#L109-L114

I'll try to come up with a (tested) PR (but I am unfamiliar with web dev in general and the templating language in particular).

xet7:
Fixed at https://github.com/wekan/wekan/pull/6050 . Thanks to @Chostakovitch

PR #6050, opened 2025-12-28
Re-add JS closing class to unicode close announcement symbol

Chostakovitch:
Addresses #6049.


Chostakovitch:
Tested as follows: 
wekan

Issue #6002, opened 2025-11-06
Drag & Drop of Cards in a Board is no longer functional in 8.17.0


daltonmcallan:
I updated to the latest candidate snap. 

WeKan ® Version	8.17.0
Meteor version	2.14
Node version	14.21.4
MongoDB version	7.0.25
MongoDB storage engine	wiredTiger

Drag and Drop within a board is gone for cards, any idea why @xet7 ? Thank you so much for your work.

xet7:
@daltonmcallan 

Have you run migrations yet? At right sidebar / Board Settings / Migrations?

gre3x:
I am getting the same issue after migrating my boards. I can't drag cards between columns anymore

daltonmcallan:
@xet7 yes sir! Do I need to run the "Restore all archived migrations"  that's the one I don't want to run so that I don't unarchive. otherwise everything else was executed.

BabyFnord:
@xet7 Same here:
All migrations except for option `archived` done. 
No drag'n'drop of lists possible. 
Drag'n'drop to change the order of boards also does not work on the board overview page.
Yes, I ran migrations on my boards as per right sidebar/board settings/migrations. 

Here's a few things:
When I browse to some of my boards which do not have many assets, drag'n'drop only becomes available when I reload them. On boards with more assets, reloading does not work. 

When I cycle though the board view options (swimlanes/lists/calendar), drag'n'drop becomes available on all boards. This is neither persistent across browsers nor sessions. For drag'n'drop to become available, I have to cycle the options every time I access WeKan in a new session.

EDIT:
![Image](https://github.com/user-attachments/assets/6d6f4a21-7a07-41be-84f1-cc9caa4f5bc7)

Rearranging lists is not persistent: After arranging list of one boards, browsing to another board and then come back to the previously used board, reverts the order of lists to the order of creating them.

EDIT:
![Image](https://github.com/user-attachments/assets/2455c0f4-b5fb-48c1-89d8-5334cf525f3a)

As there are no options for cycling view options in the board overview page, drag'n'drop isn't available there. Also, I miss the previously available option to see all boards in one view, regardless of starred or not.

WeKan version: v8.17.0
Did you configure root-url correctly so Wekan cards open correctly? Yes.
Operating System: Linux 3.10.0-1160.119.1.el7.tuxcare.els21.x86_64
Deployment Method: bundle
Http frontend if any: Nginx
Meteor-Version: 2.14
[...]

PR #6052, opened 2025-12-29
Merge list component methods with same name

### LINKED ISSUE #6048: [8.17] Cannot re-arrange lists within swimlanes
Hi there!

#6002 may have been closed a little early, because the issue described in the comments hasn't been fully addressed: https://github.com/wekan/wekan/issues/6002#issuecomment-3510535659

I still get the same behavior, *i.e.* lists behave as if the "drag and drop" symbol was only a picture.
Version information:

```
WeKan ® Version 	8.17.0 [note: actually, commit 98d05ce
Version de Meteor 	2.14
Version de Node 	14.21.4
Version de MongoDB 	8.0.17
Moteur de stockage MongoDB 	wiredTiger
MongoDB Oplog activé 	
Type d'OS 	Linux [note: standard Docker image]
OS Plate-forme 	linux
OS Architecture 	x64
OS Version 	6.12.57+deb13-amd64
```

However, something puzzles me: on a brand new instance, with a new user and a new board, two new lists, and after running sidebar migrations, it starts to work.

On the "old" instance, with exactly the same configuration, doing the same procedure change nothing.

Can I do something to debug further this?
Thanks :)

Chostakovitch:
Probably results in inconsistant behavior (which one overwrites the other?)

Hopes to fix #6048 (more details there).

### DIFF
--- a/client/components/swimlanes/swimlanes.js
+++ b/client/components/swimlanes/swimlanes.js
@@ -2,7 +2,147 @@ import { ReactiveCache } from '/imports/reactiveCache';
 import dragscroll from '@wekanteam/dragscroll';
 const { calculateIndex } = Utils;
 
+function saveSorting(ui) {
+  // To attribute the new index number, we need to get the DOM element
+  // of the previous and the following list -- if any.
+  const prevListDom = ui.item.prev('.js-list').get(0);
+  const nextListDom = ui.item.next('.js-list').get(0);
+  const sortIndex = calculateIndex(prevListDom, nextListDom, 1);
 
+  const listDomElement = ui.item.get(0);
+  if (!listDomElement) {
+    return;
+  }
+
+  let list;
+  try {
+    list = Blaze.getData(listDomElement);
+  } catch (error) {
+    return;
+  }
+
+  if (!list) {
+    return;
+  }
+
+  // Detect if the list was dropped in a different swimlane
+  const targetSwimlaneDom = ui.item.closest('.js-swimlane');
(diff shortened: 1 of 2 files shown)


Chostakovitch:
⚠️ While it works visually, the sorting order is not persistent; https://github.com/wekan/wekan/issues/6002#issuecomment-3510535659 still seems valid in the sense that, even when the race-condition gives enough luck to go through the "right" function,

> After arranging list of one boards, browsing to another board and then come back to the previously used board, reverts the order of lists to the order of creating them.

Maybe another issue should be opened, if I understood correctly so far.

Chostakovitch:
I think that this second commit mitigates the issue of sorting saving for the same reason: the `stop()` method associated with sortable components was defined twice, with the same timeout, but one of the method did not have the save logic. 

I tried not to touch the code too much (and sorry about my IDE and spacing/tabs, haven't checked this time), but I have a few remarks/things about I'm not sure:
- `initSortable` is never called yet contains recent code (for the future?) **but** I used a function to deduplicate the saving code;
- I have no idea what `listGroups` are, so I let the whole code at the bottom of the file even if some parts felt duplicated to me;
- I had to add `updatedAt` (and `modifiedAt` for a reason I couldn't find in client code) in authorized fields of the list's model to avoid console warnings.

Still on my example old Wekan instance, running with those patches, persistent sorting of lists now works and I hadn't seen other impacts yet, but, as I said, I am kind of in the dark.

PR #6086, opened 2026-01-21
Replace mquandalle:collection-mutations with collection helpers

harryadel:
[mquandalle:collection-mutations](https://github.com/mquandalle/meteor-collection-mutations) is just a wrapper around https://github.com/dburles/meteor-collection-helpers. So for now they're moved to helpers but during the 3.0 bump we're going to need to also rework them due to my [PR](https://github.com/meteor/meteor/pull/13830) :grin: 

I couldn't test all functionality so please help me with that part before merging @xet7

### DIFF
--- a/client/components/boards/boardArchive.js
+++ b/client/components/boards/boardArchive.js
@@ -37,7 +37,7 @@ BlazeComponent.extendComponent({
           await board.restore();
           Utils.goBoardId(board._id);
         },
-        'click .js-delete-board': Popup.afterConfirm('boardDelete', function() {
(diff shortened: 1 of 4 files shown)


xet7:
@harryadel 

I did not get this working yet.

First I added back REST API docs (like `/***** Some API ****/ info` ) and code comments you removed.

Then I tried to get working restore board from archive, that showed empty grey screen after remove board, and after some fixes there was still swimlanes etc not visible.

Then I did run out of time to get this working. I'll try to do something else.

PR #6123, opened 2026-02-04
Add brief Kanban explanation to README

AymenHassini19:
This pull request adds a short explanatory paragraph to the README.md describing what Kanban is and how it helps visualize and manage workflows. The goal is to make the project more accessible to new users who may be unfamiliar with the Kanban methodology.


xet7:
That is too limited explanation. I have current comparison of terms at https://wekan.fi/docs/#spreadsheet-vs-kanban , later that will be improved. Docs itself belong to docs directory to specific subdirectories of each feature, not readme.

AymenHassini19:
got it! thank you for the clarification.

PR #6127, opened 2026-02-06
Fix #6118: List widths minimum is too small, all are stuck together

### LINKED ISSUE #6118: Bug at WeKan v8.25: List widths minimum is too small, all are stuck together now
### Bug Description

List widths minimum is too small, all are stuck together now.  If you adjust the width and refresh the page, they revert back to being squished again.

Reported at https://github.com/wekan/wekan/issues/6108#issuecomment-3825591378

### Platform / Installation Method

Snap Candidate

### CPU

amd64

### Steps to Reproduce

In newest WeKan v8.25 lists have too small minimum width

### Relevant Logs

```shell

```

### Additional Context

_No response_

KhaoulaMaleh:
**What this fixes**

This PR fixes issue #6118 where lists could be resized to be extremely narrow, causing them to bunch together and become hard to use. That small width was saved, so refreshing the page kept the layout broken.

**What’s changed**

Lists now have a minimum width of 270px to keep them usable
Previously saved tiny widths are automatically corrected
CSS also enforces the minimum width as a fallback
Collapsed lists still shrink to 30px as expected

**Result**
Lists don’t get squished together anymore
Resizing feels more predictable
Reloading the page won’t bring back broken layouts
Collapsed lists continue to work as expected

**How to test**

Open a board with several lists
Try resizing a list — it should stop shrinking at 270px
Refresh the page — list widths should stay usable
Collapse a list — it should still reduce to 30px

### DIFF
--- a/client/components/lists/list.js
+++ b/client/components/lists/list.js
@@ -290,7 +290,7 @@ BlazeComponent.extendComponent({
     let isResizing = false;
     let startX = 0;
     let startWidth = 0;
-    let minWidth = 100; // Minimum width as defined in the existing code
+    let minWidth = 270; // Minimum usable list width to prevent lists from being squished together
     let listConstraint = this.listConstraint(); // Store constraint value for use in event handlers
     const component = this; // Store reference to component for use in event handlers
 
--- a/client/components/lists/listHeader.jade
+++ b/client/components/lists/listHeader.jade
@@ -221,8 +221,8 @@ template(name="setListWidthPopup")
   #js-list-width-edit
     label {{_ 'set-list-width-value'}}
       p
-        input.list-width-value(type="number" value="{{ listWidthValue }}" min="100")
-        input.list-constraint-value(type="number" value="{{ listConstraintValue }}" min="100")
+        input.list-width-value(type="number" value="{{ listWidthValue }}" min="270")
+        input.list-constraint-value(type="number" value="{{ listConstraintValue }}" min="270")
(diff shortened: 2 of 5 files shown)


xet7:
@KhaoulaMaleh 

Your current code hardcodes new values, it is not a correct way to fix it.

Instead, please check existing feature where at `☰` menus of list and swimlane user can set min and max width and height. It is saved to `models/users.js` profile settings to database.

You can search for related commits at ChangeLog:
https://github.com/wekan/wekan/blob/main/CHANGELOG.md

And also after cloning wekan repo search this way:
```
cd wekan
./find.sh max
```

KhaoulaMaleh:
Thanks for pointing that out @xet7 you’re right. Hardcoding new values isn’t ideal here. 
I’ll review the related logic, check past changes referenced in the ChangeLog, and update the fix to rely on the existing user-configurable settings so behavior stays consistent and flexible.

KhaoulaMaleh:
@xet7  
instead of overriding stored values with hardcoded CSS/JS floors, the fix raises the validation boundaries in the existing configurable system (resize handler, ☰ popup, and storage validators) from 100 to 270 — the same default the system already uses. Users can still configure their own widths via the ☰ menu, but can't go below 270px.

xet7:
@KhaoulaMaleh 

This PR has old git version tags and adds old bugs back, it is not possible to merge this. Please make new wekan fork and PR with only minimal changed lines.

PR #6141, opened 2026-02-12
Use standard AM/PM time format in calendar view (remove localized hour notation)

### LINKED ISSUE #6054: The due date hour notation in the Calendar board view is not intuitive in Traditional Chinese(Taiwan) locale
The following is the hour notation (and their literal translation) in the calendar view of the recent Wekan version(8.19):

Image

The current notation is confusing and does not map to what we consider to be _N_ AM/PM, the _N_ a/p expression is slightly better but is still uncommon to what we accustomed to.

Consider:

* Avoid applying localization to this notation
* Use am/pm(or their uppercase counterparts) instead of a/p

KhaoulaMaleh:
**Description**

In the calendar view (v8.19), the hour notation is localized (e.g., 上/下/中 or similar variants), which can be confusing and does not align with the commonly expected AM/PM format.

This PR avoids applying localization to the hour notation and switches to a standard non-localized AM/PM format for better clarity and consistency.

**Related issues**
closes #6054


xet7:
@KhaoulaMaleh 

Please only do minimal changes. Do not format existing code like changing whitespace, dividing code to multiple lines, etc. It is too hard to see at this PR, what is the actual real changes.

KhaoulaMaleh:
@xet7 
Thanks for the feedback! I’ll update the PR to include only the minimal required changes without any formatting modifications.

PR #6178, opened 2026-03-08
Remove peerlibrary:blaze-components

harryadel:
`peerlibrary:blaze-components` is a nice package but unfortunately it has become unmaintained. I've considered doing a local fork here under wekan's name but decided not to because:
-  This package has so many [underlying packages](https://github.com/peerlibrary/meteor-blaze-components/blob/45643cf90b69c1ae2a3f31fb16b1041744fe4080/package.js#L60)
	I was unsure whether I'd lump all of those packages into one or keep them as separate
- It utilizes Coffescript
- Overrides Blaze internals 
- It's an Anti-Pattern in a sense. If you'd like to use components then use a UI library that has components from the ground up like React.

All of these points made maintaining or keeping it a burden rather than a blessing. Also it's important for Wekan and any Blaze user in general to remain close to the API especially considering the [latest developments](https://forums.meteor.com/t/blaze-next-vision-ideas-tasks-etc/64460). 

## Changes:

- Remove the `peerlibrary:blaze-components` Meteor package and all 6 peer dependencies
- Migrate all components from `BlazeComponent.extendComponent()` to native Meteor `Template.onCreated/helpers/events` pattern
- Replace the `Mixins` global and BlazeComponent-based `InfiniteScrolling` mixin with standalone utility modules
- Clean up ESLint globals (`BlazeComponent`, `Mixins`, `InlinedForm`, `ReactiveTabs`)

I think we this out of the way, we may consider dropping coffeescript package next. Also I need to understand how jade/pug templates would work in the 3.0 version.

@xet7 Please give it a thorough review and a test run. I tried my best and solved a few bugs post the migration with cards actions and PDF export.

### DIFF
--- a/models/exportPDF.js
+++ b/models/exportPDF.js
@@ -32,8 +32,8 @@ runOnServer(function() {
    */
   Picker.route('/api/boards/:boardId/lists/:listId/cards/:cardId/exportPDF', async function (params, req, res) {
     const boardId = params.boardId;
-    const paramListId = req.params.listId;
-    const paramCardId = req.params.cardId;
+    const paramListId = params.listId;
+    const paramCardId = params.cardId;
     let user = null;
     let impersonateDone = false;
     let adminId = null;
@@ -48,8 +48,12 @@ runOnServer(function() {
     // If board is public, skip expensive authentication operations
     if (board.isPublic()) {
       // Public boards don't require authentication - skip hash operations
-      const exporterCardPDF = new ExporterCardPDF(boardId);
-      exporterCardPDF.build(res);
+      const exporterCardPDF = new ExporterCardPDF(
+        boardId,
+        paramListId,
+        paramCardId,
+      );
+      await exporterCardPDF.build(res);
       return;
     }
 
@@ -69,6 +73,11 @@ runOnServer(function() {
       user = await ReactiveCache.getUser({
         'services.resume.loginTokens.hashedToken': hashToken,
       });
+      if (!user) {
+        res.writeHead(401, { 'Content-Type': 'text/plain; charset=utf-8' });
+        res.end('Invalid token');
+        return;
+      }
       adminId = user._id.toString();
       impersonateDone = await ReactiveCache.getImpersonatedUser({ adminId: adminId });
     } else if (!Meteor.settings.public.sandstorm) {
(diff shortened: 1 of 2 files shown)


xet7:
@harryadel 

When I add new board, it automatically scrolls down showing only "Add List" popup automatically opened to bottom of page. To see remaining of board I need to scroll up.

xet7:
@harryadel 

Blaze works with Meteor 3, and Blaze works with realtime updates to page, like everyone sees immediately when card has been moved. There is no Jade/Pug for Meteor 3 that has realtime updates like there is Blaze. Pug can render static pages, but not realtime like Blaze.

At that future of Blaze talk was mentioned that Blaze supports some legacy browsers. I'm not sure is that really true, and what browsers he is defining as legacy? Internet Explorer? Legacy Edge?

I have prototype where login etc works at all browsers without using any browserside Javascript and cookies, so there is no possibility for any XSS attacks at all:

https://github.com/wekan/omi

But point of Meteor WeKan is the realtime drag-drop with immediate feedback. It is only visible to developers that read code, is it using Blaze templates or something else.

xet7:
Also, all those prototypes are very minimal. They are not ready to replace full-featured Meteor WeKan.

For those minimal prototypes, I first need to implement remaining with minimal code, so everything works without Javascript and cookies, and have Javascript for browsers that supports it, abstracting it to no-code IDE that supports all browsers, and use itself to build remaining features, just because nobody else has yet made web framework that supports all browsers. And then add round-robin etc that Meteor does not support. But that does not exist fully yet.

Meanwhile, Meteor WeKan is what everybody uses, and it needs to be maintained when there is not any full-featured drop-in replacement yet. Internal code changes to Meteor WeKan can be anything, users only see web interface, only developers reading code see internal structure. Meteor WeKan keeps being useful anyway as an example how to do full-featured kanban with Meteor.

harryadel:
Why are your mentioning prototypes and Jade?? I don't get it!! I understand that this PR might have a few issues here and there but overall they're fixable. And this is exactly why I asked you to test.

>When I add new board, it automatically scrolls down showing only "Add List" popup automatically opened to bottom of page. To see remaining of board I need to scroll up.

Are there any other bugs you'd like to report?

xet7:
@harryadel 

> Why are your mentioning prototypes and Jade?? I don't get it!! 

Sorry, at your description you mentioned templates like Blaze and Jade, I tried to answer something, unsuccessfully.

harryadel:
@xet7 Yeah, I did mention how I'm considering the Jade situation in regards to 3.0 migration. Fret not, this too will be fixed ;)

xet7:
@harryadel 

I would test for more bugs, but this very big PR has merge conflicts, I don't know how to fix those merge conflicts.

harryadel:
Leave the conflict fixes for me. I'd greatly appreciate if you test the PR E2E style not actual code and tell me where things are off.

PR #6187, opened 2026-03-15
Fix #6165: Improve global search relevance and add exact-field operators

### LINKED ISSUE #6165: Bug: Card search returns irrelevant results and non-related cards
### Bug Description

When searching for cards, the results display a heading like 'XXX Cards Found', but many results are not relevant to the search terms. The search often returns cards where the keyword was found in unrelated fields (e.g., comments, attachments, custom fields), not in the card title or main description.

Example:
- Searching for a card titled 'Task A' may return cards where 'Task A' is only mentioned in a checklist, comment, or attachment, even if the card is about something else. The relevance is low, leading to user confusion.

Expected: Cards most relevant to the search term should appear first, ideally prioritizing title and description matches. There should be options to filter/search more precisely (e.g., title only, description only) or add relevance ranking.

### Platform / Installation Method

Kubernetes

### CPU

amd64

### Steps to Reproduce

1. Go to the card search feature
2. Enter a search term that should match a specific card title
3. Review the results – unrelated cards may appear because the term exists in attachments, comments, or custom fields
4. The results heading shows 'XXX Cards Found', but many are not actually related

### Additional Context

From the code, search is performed on title, description, comments, custom fields, checklists, and attachments (see server/publications/cards.js:642-669). Suggest adding relevance ranking and field prioritization, as well as user filters for more precise searching.

KhaoulaMaleh:
### Description
This PR addresses issue #6165 where the global card search was returning confusing or irrelevant results because it matched keywords in obscure fields (like comments and attachments) and treated them with the same weight as exact title matches.

To fix this, two major enhancements have been added:

Relevance Ranking on Global Search: Text searches now assign a _score to matching cards and sort them by relevance before serving them to the client. Matches in the Card Title are scored highest, followed by the Description, and then Custom Fields. This ensures the most visually relevant cards always appear first.

Precise Search Operators: Users can now search specific fields explicitly to narrow down their search and bypass global searching.      

### How to Test
Navigate to the global search bar.
Search for a generic keyword that exists in deeply nested comments/attachments as well as the title of a card.
Verify: The card with the keyword in the title is ordered visibly earlier/first in the results.
Try combining literal operators to narrow down matches:
title:"Exact Title Text"
description:keyword
attachment:filename
customfield:value

### Fixes
Closes #6165


xet7:
@KhaoulaMaleh 

What is point here with changes to translations?

KhaoulaMaleh:
@xet7 
The changes to the translation file  are actually mechanically required for the new search operators to work 
In Wekan's search architecture, the query parser inside [config/query-classes.js] maps what the user types in the search bar back to internal constants by relying on `TAPi18n`. 
It essentially says: "Check if the word the user typed matches the translated value of `operator-title`, `operator-description`, etc."
If we don't add those keys to the [en.i18n.json]  file:
1. `TAPi18n` will fail to find a translation for `operator-title`.
2. The parser won't recognize that typing `title:something` is an operator.
3. The search will treat `title:something` as just a normal text string instead of filtering by the title.

PR #6275, opened 2026-04-12
Fix: include historical comments and activities in board exports

### LINKED ISSUE #6274: Export includes only comments from current year
### Bug Description

when using the export function in the GUI, only the comments from 2026 are included. Comments from 2025 and earlier are not included.

Here the Excel as example:

Image

Here as JSON example:

Image

### Steps to Reproduce

1. Board Settings
2. Export Board

### Relevant Logs

```shell

```

### Additional Context

_No response_

KhaoulaMaleh:
### Reason: 
Older records in the card_comments and activities collections often lack the boardId field. Since the exporters were filtering strictly by boardId, these records were being skipped.

### Changes:

Exporter Logic: Modified both JSON and Excel exporters to retrieve comments and activities based on the cardId of all cards on the board, ensuring historical data is captured.
Bug Fix: Fixed an undefined variable in ExporterExcel.js that was preventing Custom Fields from being exported.
Robustness: Added username fallbacks in the Excel export for deleted users.

Related Issue: Fixes #6274

### DIFF
--- a/models/exporter.js
+++ b/models/exporter.js
@@ -124,8 +124,17 @@ export class Exporter {
       { boardIds: this._boardId },
       { fields: { boardIds: 0 } },
     );
-    result.comments = await ReactiveCache.getCardComments(byBoard, noBoardId);
-    result.activities = await ReactiveCache.getActivities(byBoard, noBoardId);
+    const cardIds = result.cards.map(card => card._id);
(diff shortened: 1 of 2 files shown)


Tuphal:
thank you very much !!!

One small issue:
The export is now missing the "lists" part:

grafik

I could copy&paste from the previous export and have now a full backup :)

PR #6278, opened 2026-04-13
Repair broken swimlane list rendering and add board integrity repair flow

### LINKED ISSUE #6266: Cards disappear from lists that has been minimized
### Bug Description

Lists that has been at some point minimized (using the arrow over the name of the list) sometimes stops working. I have not been able to fully pinpoint the needed actions for this to happen.

I have a board where the behavior is 100% consistent, all three lists that has been minimized faild, all other lists works as expected. After a brief state of panic, I have confirmed via the API that the cards are still there, but I have no way to access them via the Wekan UI.

An 'unfunctional list' does not show any cards (even though it contain several), it is not possible to drag-and-drop cards to the list and it is not possible to create new cards. Cards can be moved to the list via the drop-down in the card, but they will then disappear. All cards (both original cards and cards assigned via drop-down) are available through the API.

On a new test board (on the same instance of Wekan) the same thing happened but the list re-activated upon browser reload - on the first board where the problem appeared, I have not found a way to get the cards back - neither a change of browsers or restart of backend helps.

Image

### Steps to Reproduce

As mentioned above, I am not able to reproduce this fully on a new board. If needed, I am willing to share a dump of the board via private email, but I do not want to upload it in public.

### Relevant Logs

```shell
No messages in log.

Wekan version: 8.74.0
```

### Additional Context

_No response_

harryadel:
## Summary

This PR stops swimlane view from silently rendering structurally invalid list data and adds a board-level repair flow for admins when integrity issues are detected.

## Root Cause

Some boards still contain pre-migration or inconsistent per-swimlane data, including shared lists, lists linked to deleted swimlanes, and cards whose list and swimlane references no longer agree.

Recent fallback rendering made those broken structures appear as normal swimlane data instead of surfacing them as board integrity problems. That could leave a list visible in swimlane view while its cards were filtered into a different swimlane context, which made the list look empty and caused moved cards to appear to disappear.

## What Changed

- made swimlane rendering strict again so a swimlane only shows lists and cards that actually belong to it
- removed the client-side fallback that displayed shared and orphaned swimlane data as if it were valid swimlane content
- added a board integrity banner in the board UI that checks for migration issues and lets board admins run a repair directly from the board
- wired the repair action to the existing comprehensive board migration flow
- expanded comprehensive migration detection and repair so it also handles invalid swimlane references, invalid list references, shared lists, and mismatched list/swimlane relationships

## User Impact

- broken boards are surfaced as integrity issues instead of showing ghost lists in swimlane view
- board admins get a direct repair path from the affected board
- after repair, the board returns to a consistent per-swimlane structure

Fixes #6266

### DIFF
--- a/client/components/boards/boardBody.css
+++ b/client/components/boards/boardBody.css
@@ -10,6 +10,103 @@
   min-width: 100%;
 }
 
+.board-wrapper.has-integrity-banner .board-canvas {
+  top: 76px;
+}
+
+.board-integrity-banner {
+  position: absolute;
+  top: 12px;
+  left: 12px;
+  right: 12px;
+  z-index: 7;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 12px;
+  padding: 10px 12px;
+  border: 1px solid rgba(0, 0, 0, 0.18);
+  border-radius: 6px;
+  background: rgba(255, 245, 204, 0.98);
+  color: #2c2411;
+  box-shadow: 0 2px 8px rgba(0, 0, 0, 0.08);
+}
+
+.board-integrity-banner.is-repairing {
+  background: rgba(217, 238, 255, 0.98);
+  color: #14304f;
+}
+
+.board-integrity-banner.is-error {
+  background: rgba(255, 224, 224, 0.98);
+  color: #4a1919;
+}
+
+.board-integrity-copy {
+  min-width: 0;
+}
+
+.board-integrity-title {
+  font-weight: 700;
+  margin-bottom: 2px;
+}
+
+.board-integrity-text {
+  font-size: 13px;
+  line-height: 1.35;
+  white-space: nowrap;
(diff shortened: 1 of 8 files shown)


xet7:
Related https://github.com/wekan/wekan/blob/main/docs/ImportExport/History-of-Schema-Migrations-Systems.md

robertjensen:
Hi - Is there any chance for this to be merged, the issue is quite a problem for the use of Wekan, and with a solution this close, it would be amazing if it could find its way into production :)

xet7:
@robertjensen 

This PR is still at draft mode, so it is still in progress, not ready to be tested or merged.

If someone has time to test these changes, do they work, comments welcome to this PR.

Related, about building WeKan: https://github.com/wekan/wekan/blob/main/docs2/DeveloperDocs/Build-and-Create-Pull-Request.md

There is also other issues like #6307 , any help very welcome.

Issue #6007, opened 2025-11-08
list width


mimZD:
change list width  size min and max not apply ❌
auto size not work ❌
dragging width works but after refresh it revert to last size ❌
also dragging causes disappearing that list scrollbar disappearing and scrolling ❌

_______________

i updated from 7.91 to 8.17

i did migration
i also test new board

mimZD:
now my lists are small and cant change them

<img width="1762" height="284" alt="Image" src="https://github.com/user-attachments/assets/6f87dcc4-2ca2-4edc-ab5d-8e307ed72e48" />

mimZD:
> now i my list are small and cant change them
> 
> <img alt="Image" width="1762" height="284" src="https://private-user-images.githubusercontent.com/25129673/511682185-6f87dcc4-2ca2-4edc-ab5d-8e307ed72e48.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjI2MTI1OTgsIm5iZiI6MTc2MjYxMjI5OCwicGF0aCI6Ii8yNTEyOTY3My81MTE2ODIxODUtNmY4N2RjYzQtMmNhMi00ZWRjLWFiNWQtOGUzMDdlZDcyZTQ4LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTExMDglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUxMTA4VDE0MzEzOFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWMwN2I4ZmI1MThmMDkxMWYzYmYxZGVmMjY4MmYxZjY5NDQ0OTAyZjc5ODdiNTM3OWY2MmFmMmFiZDYzOGI1Y2UmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.Swc9iFTuRJCjLPnFsLE1V-BNR0ZXTz0GOcfv_byS7R4">

[...]

PR #6303, opened 2026-04-18
 Fix Issue #6007: Add permission checks to list width method

Raawen-dev:
## Fixes Issue #6007
List widths not persisting after page refresh and 403 errors when resizing

## Root Cause
The `applyListWidth` method was missing permission checks and used deprecated `update` instead of `updateAsync`

## Solution
- Added user authentication check
- Added board membership verification
- Changed `Lists.update` to `Lists.updateAsync`
- List widths now persist after page refresh

## Testing
- Resizing lists works
- Widths save to database
- Widths persist after page refresh
- No more 403 errors

### DIFF
--- a/server/models/users.js
+++ b/server/models/users.js
@@ -319,14 +319,21 @@ Meteor.methods({
     (await ReactiveCache.getCurrentUser()).setDateFormat(dateFormat);
   },
 
-  async applyListWidth(boardId, listId, width, constraint) {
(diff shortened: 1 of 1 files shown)


xet7:
Thanks!

Next time, you can use this kind of text, so that it automatically closes issue, when I merge PR:
```
Fixes #6007
```
I had to manually close that issue, because this PR has text `Issue` in the middle of text:
```
Fixes Issue #6007
```

PR #6313, opened 2026-04-25
fix(myCards): fix subscription crash and show all board cards

aminsaidane:
Fix myCards subscription crash and empty view
Bug 1 — Server crash
findCards() was returning a plain object { _sessionDataToFetch: storedSessionData } mixed into the cursors array. Meteor's publish system only accepts MongoDB cursors, causing:
Error: Publish function returned an array of non-Cursors
Fixed by refactoring findCards() to return { cursors, sessionData } and using this.added() in the publisher for the session data.
Bug 2 — Empty cards view
myCards was filtering by members/assignees on each card, but most cards have empty arrays. Changed to show all cards on the user's boards by removing the OPERATOR_USER predicate and deleting query.selector.$and.
Files changed: server/publications/cards.js
Note: globalSearch, nextPage, previousPage, and brokenCards call the same findCards() and need the same { cursors, sessionData } destructuring fix — left as a follow-up.

### DIFF
--- a/server/publications/cards.js
+++ b/server/publications/cards.js
@@ -79,20 +79,25 @@ import { CARD_TYPES } from '../../config/const';
 import Org from "../../models/org";
 import Team from "../../models/team";
 
+
 Meteor.publish('card', async function(cardId) {
   check(cardId, String);
 
+
   const userId = this.userId;
   const card = await ReactiveCache.getCard({ _id: cardId });
+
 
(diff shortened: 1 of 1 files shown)


xet7:
My Cards only shows cards where user is assigned or member, so I think there is some misunderstanding here.

Also, I don't know why you insert so many empty lines. Point is to change only required lines.

PR #6337, opened 2026-05-10
Fix attachments API 500 error and upload crash

### LINKED ISSUE #6326: Accessing attachments api endpoint still fails
### Bug Description

The inability to access the attachments endpoint (https://wekan.fi/api/v7.93/#get_board_attachments) was reported in #6319 for Wekan `8.74.0` and was apparantly fixed in `9.0.3`.

However, I am now at `9.07.0` but the endpoint still fails. All other endpoints works as expected.
Is the bug not really resolved or is there something that should be done to the local installation? 

The attachments are functional from the normal UI.

For others with the same problem, I have realized that as a (slightly desperate) workaround the same information is available from `http://<hostname>/attachments`  - however this is a very messy solution  and the api endpoint would be much preferred.

### Steps to Reproduce

Access the attachment endpoint, http://<wekan_url>/api/boards/<board_id>/attachments

Internal Server Error is returned.

### Relevant Logs

```shell

```

### Additional Context

Wekan version `9.07.0`

In case this could be related to local configuration, I provide the configuration for attachments (I have never touched these):

### Attachment Storage Configuration
**Writable Path**
/data

**Attachments Path**
/data/attachments

KhaoulaMaleh:
**Description:**

### What was wrong

1. **API crash**: The `/api/boards/:boardId/attachments` endpoint called `attachment.link()`, but the objects returned from the database are plain objects that don't have a `.link()` method. This threw a TypeError → 500 error.

2. **Upload crash**: The `onAfterUpload` hook in `attachments.server.js` had an unguarded `await Attachments.updateAsync()` call. If it failed, the unhandled rejection killed the Node process, causing the backend to restart and the UI to flicker.

3. **File watcher loop** (dev only): Without `WRITABLE_PATH`, uploaded files were saved inside the project directory, triggering Meteor's file watcher to restart the server on every upload.

### What this PR does

- Replace `attachment.link()` with manual URL construction (`/cdn/storage/attachments/${id}`)
- Wrap `updateAsync` in try/catch to prevent unhandled rejections
- Move default storage path to `.meteor/local/data/` so uploads don't trigger file watcher restarts
Fixes #6326


xet7:
Did not work, reverted this here: https://github.com/wekan/wekan/commit/8b6c4594537fd1da9110b733ccb71f07e2cb9854

KhaoulaMaleh:
Thanks @xet7 for checking it. I'll look at what broke in the revert commit and come back with a more targeted fix. Could you share any error logs from when you tested it? That would help me narrow down which part caused the issue.

xet7:
@KhaoulaMaleh 

Look at git log history, at what commit ID (replace 12345, copy to clipboard for pasting) I merged PR, checkout at that commit, build WeKan, you see the errors:

```
git clone ...

cd wekan

git log

q

git checkout 12345

./rebuild-wekan.sh

1

./rebuild-wekan.sh

2
```

xet7:
@KhaoulaMaleh 

After looking at it, change git history position back to newest:

```
git checkout main
```

KhaoulaMaleh:
@xet7 since the related issue is closed to which issue do I relate the new PR ?

PR #6499, opened 2026-07-21
AI: Fix card attachment filename helper bootstrap (#6493)

### LINKED ISSUE #6493: 10.04: Cards not opening
In 10.04, cards don't open any more. If I click on a card, it moves a bit to the right (why?), but nothing else happens.

Problems reported are only the ferretdb at 170% CPU
`FerretDB is using 168% CPU (100% = one core; this host has 8 cores) while host-wide CPU is only 65% — FerretDB alone is monopolising the machine, which the host-wide percentage hides. 65371 commands processed total; busiest operations: aggregate=64075, find=944, distinct=168, listIndexes=94, getMore=33. FerretDB self-regulates at 0ms/op.`

brlin-tw:
**Content warning:** This pull request is _mostly_ AI generated.  While I did test the code I am not capable of fully validate it.  
_Feel free to drop it if it's not salvageable._

Load safeFilename.js from the explicit client feature entry point so Blaze registers cleanFilename and downloadFilename before attachment galleries or the admin files report render. Without this import, opening any card with an attachment aborts with "No such function: cleanFilename".

Extend the filename display wiring test to verify that the helper module remains reachable from the client bootstrap, rather than only checking that its source and template references exist.

Fixes #6493.

Assisted-by: Codex:gpt-5.6  
Tested-by: 林博仁(Buo-ren Lin) <buo.ren.lin@gmail.com>  
Signed-off-by: 林博仁(Buo-ren Lin) <buo.ren.lin@gmail.com>

### DIFF
--- a/client/features/main.js
+++ b/client/features/main.js
@@ -38,6 +38,7 @@ import '/client/components/main/layouts.js';
 import '/client/components/main/myCards.js';
 import '/client/components/main/popup.js';
 import '/client/components/main/recoveryMaintenance.js';
+import '/client/components/main/safeFilename.js';
 import '/client/components/main/spinner.js';
 import '/client/components/main/support.js';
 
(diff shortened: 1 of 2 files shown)


xet7:
Already fixed.

PR #6593, opened 2026-08-13
Use neutral cards on All Boards

khuongsatou:
## Summary

- keep All Boards card surfaces neutral instead of inheriting board colors
- preserve board-color accents on dedicated color indicators
- add a source guard covering the neutral-card behavior

## Validation

- `node tests/allBoardsNeutralCards.test.cjs`
- `git diff --check upstream/main...HEAD`


xet7:
There will be similar feature, `All Boards / Member Settings / Change color / [_] All boards` that is per-user override of board icons colors.

PR #6621, opened 2026-08-22
Await REST user creation before returning its ID.

AhmedLukman:
## Summary

- await `Accounts.createUser` in `POST /api/users/`
- return the resolved string user ID instead of serializing the Promise as `{}`
- add a plain-Node regression guard for the route

## Live verification

Tested with a clean MongoDB replica set and the official
`wekanteam/wekan:v11.07` image.

Before the fix, the endpoint returned HTTP 200 with:

```json
{"_id":{}}
```

The user appeared shortly afterward with a string ID, confirming that account
creation continued after the response had serialized the unresolved Promise.

After applying the source-equivalent await change to the same image, the
endpoint returned HTTP 200 with a non-empty string ID. The user was immediately
queryable and the response ID exactly matched the stored user ID.

## Tests

- `node tests/run-node-suites.cjs restCreateUserAwait` — passed
- full plain-Node run — all 500 suites executed; 456 passed and 44 unrelated
  suites failed in the fresh Windows checkout because of missing npm
  dependencies, POSIX path/executable assumptions, and existing translation
  file checks

### DIFF
--- a/server/models/users.js
+++ b/server/models/users.js
@@ -2246,7 +2246,7 @@ WebApp.handlers.post('/api/boards/:boardId/members/:userId/remove', async functi
 WebApp.handlers.post('/api/users/', async function(req, res) {
   try {
     await Authentication.checkUserId(req.userId);
-    const id = Accounts.createUser({
+    const id = await Accounts.createUser({
       username: req.body.username,
       email: req.body.email,
       password: req.body.password,
--- a/tests/restCreateUserAwait.test.cjs
+++ b/tests/restCreateUserAwait.test.cjs
@@ -1,0 +1,47 @@
+'use strict';
+
+// Plain-Node regression guard for POST /api/users/.
+// Run: node tests/restCreateUserAwait.test.cjs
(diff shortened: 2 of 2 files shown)


xet7:
Some parts of this are added to next WeKan release.

AhmedLukman:
Hi @xet7, thanks for incorporating the fix. I’d also like to understand the contribution decision here so I can submit PRs in the form you prefer. This PR’s substantive change, `await Accounts.createUser(...)`, was included after the PR was opened, but the PR itself was closed rather than merged. Was there a technical or structural reason the PR could not be merged or cherry-picked? Merging or cherry-picking would preserve the contribution history automatically.

xet7:
@AhmedLukman 

There were many overlapping REST API fixes from many different people. Some of those were security issues, so they were privately reported at GitHub. Those were combined to fix the general issue related to all of them.

It is not possible to guarantee that every PR would be merged directly. Often there is something to improve, combine, etc.

PR #6626, opened 2026-08-23
Opened by mistake - work belongs to the Jalor fork.

Clement-Plancon:


xet7:
This is too many changes. You should not change translations and so many other files.

xet7:
@Clement-Plancon 

There is already WeKan feature to change font and font size. You should not add any new font files.

xet7:
@Clement-Plancon 

You should not add any foundation-specific names, logos, etc. It is already possible at `Admin Panel / Settings / Visibility` to change WeKan to different Product name, add different logos etc.

PR #6642, opened 2026-08-27
Confirm REST user deletion results

AhmedLukman:
## Summary

- make user deletion report what actually happened in the database
- return `200` only when exactly one account was removed
- return `404` with `{"error":"User not found"}` when no account matched
- return a safe internal error for an unexpected removal result or database failure
- include the success and not-found responses in the generated OpenAPI documentation
- add fast route tests and a live REST-to-MongoDB test

## Why

API clients need to know whether a deletion really happened. Previously, the endpoint simply repeated the user ID from the URL. This made a real deletion and a request for a user that did not exist look the same.

Meteor returns the number of documents removed by MongoDB. Because the route selects a unique user ID, the normal result is `1` when the user was deleted or `0` when no user matched. This change uses that result instead of assuming the deletion worked.

Calling `DELETE /api/users/:userId` with a missing user ID returned `200`, which means success, even though nothing was deleted. Scripts, integrations, and administrators could therefore believe an account was gone when it was not.

Only a zero removal count means the user was not found. An unexpected count or a database error is treated as an internal error instead of being mislabeled as a missing user.

## Testing

- `node tests/restUserDelete.test.cjs` — 5 passed
- `node tests/openapiResponses.test.cjs` — 4 passed
- focused Playwright REST test against a fresh Docker build of this checkout — 1 passed
- changelog format, release-placement, archive, and commit-link suites — 37 passed
- `node tests/run-node-suites.cjs` — both new suites pass under automatic discovery; the complete Windows run has 55 unrelated existing/environment failures involving Unix path semantics, optional modules, and translation symlink handling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5179,9 +5179,10 @@ browser build to verify).
 **In short:** **HostnameBleed**, found by GitHub CodeQL, makes tenant-hostname
 translation coverage compare example domains as exact text instead of permissive
 regular expressions. A repository-wide negative guard prevents the same test
-fault from returning. The table below is carried over from the release under
-this one, and is refilled from each build's provenance.tsv when this release is
-made.
+fault from returning. The **REST API** now confirms user deletion from the
+database result, reports missing users as 404 and publishes both response
+contracts. The table below is carried over from the release under this one, and
+is refilled from each build's provenance.tsv when this release is made.
 
 | Platform | Binary | From | Version | SHA256 |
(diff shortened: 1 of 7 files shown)


xet7:
Merged at https://github.com/wekan/wekan/commit/6acfb396e4cd01d5cdec5dfd513e50163db1198c and https://github.com/wekan/wekan/commit/29f75595ad70d6a66bef3b7cfe286008ec8da3f1 . Thanks to @AhmedLukman !

